PR #12, opened 2025-02-21
Fixed bug ignoring sampling args in Simple Env, Added </answer> as stop sequence for Math Env

darinkishore:
I think the Math Env thing should be a default, made training run take a lot longer and model consistently outputted gibberish (like some REAL weird shit sometimes) after the </answer> . 

Also, we can't override sampling args in simple env because it was just overwriting them before. so this should fix that .

### DIFF
--- a/verifiers/envs/math_env.py
+++ b/verifiers/envs/math_env.py
@@ -15,7 +15,18 @@ class MathEnv(SimpleEnv):
                  few_shot: List[Dict[str, str]] = MATH_FEW_SHOT[0],
                  fields: List[str | Tuple[str, ...]] = ["reasoning", "answer"],
                  **kwargs):
-        super().__init__(system_prompt=system_prompt, few_shot=few_shot, **kwargs)
+        sampling_args = {
+            "skip_special_tokens": False,
+            "spaces_between_special_tokens": False,
+            "stop": ["</answer>"],
+            "include_stop_str_in_output": True,
+        }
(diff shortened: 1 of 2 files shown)


willccbb:
thanks! need to set up CLA before taking PRs, will get that sorted out ASAP + do some more testing but seems sensible

amitlevy:
this change is suprisingly important for performance. I haven't ran it for long enough to verify it fixes the gibberish outputs, why do you think it helps? I don't understand why the model doesn't learn by itself to end the output after </answer>

willccbb:
Latest push addresses the SamplingArgs bug, README is changed to a tool use env which is working well in my tests. While the single step envs are faster, they aren't the primary intended uses--multi-step is the main advantage of using this over vanilla TRL so far. Don't really want to hardcode "</answer>" as a default stop token, formats will differ + we want the model to learn this directly.

OK if I close this?

darinkishore:
Yes, ofc. 

IMO advantage over TRL is your great design, agree that we want model to learn this directly, wondering if you got it to do that at all?

willccbb:
Thanks!

Have had some successful multi turn runs on 7B, this is for search with DuckDuckGo image

There's an issue with some reward computations which I'm working on now, recent refactor makes a lot of things easier to tweak without depending on the TRL fork (e.g masking env responses), should be more stable soon. still "pre-release", there will be rough edges.

PR #17, opened 2025-03-01
Makes installable from pip using git url

ronald-d-rogers:
Makes directly installable from pip without having to clone first

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,3 +19,5 @@ dependencies = [
     "liger-kernel>=0.5.2",
     "vllm==0.7.3",
 ]
+[tool.setuptools]
+py-modules = []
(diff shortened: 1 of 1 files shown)


willccbb:
pypi soon, getting close to the level of stability where i'm ready to call it a "v0.1 release" :)

ronald-d-rogers:
@willccbb: No worries. This will make it installable from a git url in the mean time I think.

willccbb:
I'm not really intending to support install methods other than the one described in the README for now, sorry. Adds more overhead + testing, and I think people will generally want to clone it anyway at this stage. It's intended to be used for research, and the easiest way to do that is to play with the example scripts.

PR #20, opened 2025-03-02
modified examples to work with commodity hw. Also added zero3 config …

cielonet:
…with cpu offloading


cielonet:
I trained for about 6 hours using this library. Modified reward funcs but am not adding the changes to the reward funcs in this pr or until I figure out why I had issues (i.e. correctness_reward_func). Thank you for starting this project. ;-)

image

PR #21, opened 2025-03-02
add reward_weights support to rubrics

stochasticchasm:
Hello, wanted to set different weights for each of the reward functions and didn't see a nice and clean way to do it. This PR has a minimally invasive implementation where you can attach it to the rubric and then use the rubric to pass it into the `get_default_grpo_config` function.

Aimed to not break existing stuff, but I feel like a cleaner option later down the line might be to just have the envs have `get_reward_funcs` and `get_reward_weights` functions, or maybe a single `get_rubric` that returns a dict with `"funcs"` and `"weights"` keys?


willccbb:
Seems sensible, let me get the CLA set up and then can merge.

PR #28, opened 2025-03-06
GSM8K should use CALCULATOR_FEW_SHOT

Some-random:
Otherwise it will keep calling the wrong tool and keep getting parsing errors...

Also, I think we should merge the trajectories into a single list in few_shots.py, as this aligns with how vLLM processes few-shot examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,14 +25,14 @@ Tested with Python 3.11 and this [image](https://hub.docker.com/layers/pytorch/p
 # script.py
 import verifiers as vf
 from verifiers.tools import calculator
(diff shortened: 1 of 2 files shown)


willccbb:
oops good catch! do you mind reverting the merge of the nested arrays so that the traces stay separate? then, `CALCULATOR_FEW_SHOT[0]` is picking out just one complete trajectory (which should be plenty for learning). people can use more few-shot examples + flatten if they'd like (maybe this should be supported automatically...), but this is the pattern being used throughout the codebase/examples.

PR #30, opened 2025-03-08
added support for data-dependent system prompts

stochasticchasm:
Hello, was thinking of some experiments and thought it would be helpful to have an option to change the system prompt depending on a particular column of the input. Implemented as an optional callable function + an optional dataset column name passed to preprocess_dataset so the system prompt can be arbitrary transforms of a field of the input dataset. The changes don't break existing uses.

Example uses could be multi-task RL for combined datasets or system-prompt-conditioned reasoning effort/length training.

### DIFF
--- a/verifiers/utils/data_utils.py
+++ b/verifiers/utils/data_utils.py
@@ -1,6 +1,6 @@
 import random
 import json
-from typing import List, Dict
+from typing import List, Dict, Callable
 
 from datasets import Dataset, load_dataset # type: ignore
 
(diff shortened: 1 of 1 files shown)


willccbb:
for now i'm gonna close this + leave it up to the user to mix in sys prompts if they want

system prompt field is optional, you can just have it be pre-included in your "prompt" column during dataset prep if you want

better abstractions for multi-env training will come later :) `preprocess_dataset` is a fairly weak abstraction that's mostly here for convenience, not intended to be load-bearing

PR #51, opened 2025-03-28
use deepcopy instead of copy

### LINKED ISSUE #45: Using Copy rather than Deep Copy
MultiStepEnv uses copy to copy states.  https://github.com/willccbb/verifiers/blob/66f9391aaa6d1e9d9cd6f4d95b857d55ad0c43a7/verifiers/envs/multistep_env.py#L69

This appears to pass some things by reference, resulting in prompts that are passed in to have model output appended to them. Suggested fix is to use copy.deepcopy.

aniemerg:
Fixes #45


willccbb:
resolved in later updates :)

PR #54, opened 2025-04-03
Fix: Data utils and Training CLI

hahuyhoang411:
When run training with `mmlu` it throws this error due to double quotes
```
    question = f"Question: {x["question"]}\n"
                               ^^^^^^^^
SyntaxError: f-string: unmatched '['
```


hahuyhoang411:
Update the example training CLI by adding `--num_processes` to handle the error:

```
[rank4]: RuntimeError: CUDA error: invalid device ordinal
[rank4]: CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
[rank4]: For debugging consider passing CUDA_LAUNCH_BLOCKING=1
[rank4]: Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.
```

willccbb:
ah sorry, missed this, but has been fixed in recent pushes

PR #63, opened 2025-05-26
vLLM native OpenAI compatible server with weight syncing

BjarniHaukur:
This PR adds a new `vllm_serve_async.py` script to verifiers. It adds:

* a fully featured OpenAI compatible endpoint
* that mirrors the weight syncing logic from `vllm_serve.py`
* while delegating endpoint complexity to `vllm.entrypoints.openai.api_server`

image

### DIFF
--- a/temp_test.py
+++ b/temp_test.py
@@ -1,0 +1,229 @@
+import os
+import signal
+import subprocess
+import unittest
+
+import torch
+import psutil
+import pytest
+from transformers import AutoModelForCausalLM
+
+from verifiers.inference.vllm_serve_async import VLLMClient
+
+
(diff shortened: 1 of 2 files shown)


willccbb:
i ended up implementing my own dynamic batching server which builds around LLM() but exposes an async endpoint, i never found AsyncLLM to be reliable for weight-syncing in training runs and it seems like the general guidance is that isn't fully supported without a custom build config/container (e.g. how it's handled in veRL and related libraries)

at some point we'll migrate to AsyncLLM but the current solution works well for now (perf tests are fairly close in my measurements) and offers a bit more control for error handling.

PR #66, opened 2025-06-02
Fix README quickstart instructions

pawalt:
The repo needs to be installed as editable or vllm won't be able to start up, and `math_train.py` uses Qwen 1.5B.


pawalt:
@willccbb lmk if there are other things I can clean up while I'm in the area or if these aren't desired changes.

willccbb:
many fixes + docs are coming with changes from dev branch soon; gonna hold off on any PRs for now

PR #77, opened 2025-06-07
Add BLEURT-based semantic similarity rubric for natural language generation tasks

rkbadhan:
A new reward function that uses Google's BLEURT metric to evaluate semantic similarity between generated text and reference answers via an auxiliary API server. This introduces the EmbedRubric architectural pattern for model-based metrics while addressing process isolation requirements for multi-GPU training.

BLEURT (https://github.com/google-research/bleurt) is a neural evaluation metric developed by Google Research that measures how well generated text conveys the meaning of reference text. Unlike traditional string-based metrics, BLEURT was trained on human judgments asking "How adequately does the system output express the meaning of the reference?" - making it inherently aligned with human preferences.

The implementation introduces EmbedRubric as a new base class for embedding-based metrics, using auxiliary hosting with
async endpoints to completely isolate model evaluation from the training process.

Key architectural innovations:
-EmbedRubric base class: Foundation for all future embedding-based and model-based rubrics
-Auxiliary hosting: BLEURT server runs as separate FastAPI process, completely isolated from trainer
-Async throughout: Native async/await with aiohttp client and FastAPI server for maximum performance
-Batch optimization: Override _score_all() to make single API call per batch instead of individual requests
-Process isolation: Zero GPU memory conflicts - BLEURT never loads in training process

Semantic advantages:
-Semantic understanding: Recognizes that "The cat is sleeping" and "A feline is resting" have similar meaning
-Human-aligned: Trained on actual human adequacy judgments from translation tasks
-Continuous scores: Provides 0-1 similarity scores rather than binary correct/incorrect

Complete example showing summarization use case with auxiliary hosting setup.Originally designed for machine translation evaluation, BLEURT has proven effective for various NLG tasks where semantic fidelity matters more than exact word matching - making it an ideal reward signal for training language models on summarization, paraphrasing, and open-ended generation tasks.

 **Before (limited options):**
  Only exact match or string similarity, models loaded in training process
  rubric = vf.Rubric(funcs=[exact_match_reward], weights=[1.0])

 **After (semantic understanding with auxiliary hosting):**
  Start auxiliary server: python -m verifiers.servers.bleurt_server
  Use semantic similarity with complete process isolation
  rubric = vf.BleurtEmbedRubric() # Async batch API calls automatically!

  **Or combine with other rewards using EmbedRubric pattern**
  rubric = vf.BleurtEmbedRubric(parser=parser)
  rubric.add_reward_func(exact_match_reward, weight=0.5) # exact correctness
  rubric.add_reward_func(format_reward, weight=0.1) # structure compliance
  **BLEURT gets remaining weight, all via async batch calls**

  Perfect for these tasks
-Text Summarization: "The study shows X" vs "Research demonstrates X" (same meaning, different words)
-Paraphrasing: Preserving meaning while changing style or complexity
-Open-ended QA: "Rome fell due to economic issues" vs "Roman Empire collapsed from financial problems"
-Explanation Tasks: Multiple valid ways to explain the same concept

  Architecture highlights:
 - EmbedRubric base class: Establishes pattern for future embedding models (ready for planned EmbedRubric expansion)
 - Auxiliary hosting: FastAPI server with health checks, error handling, and graceful model loading
 - Async optimization: Single batch API call replaces N individual calls (20x+ performance improvement)
 - Standard interface: Drop-in replacement maintaining exact same verifiers usage patterns
 -Future foundation: Ready template for sentence transformers, other embedding models, and neural metrics

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,8 @@ all = [
     "reasoning-gym",
     "smolagents>=1.15.0",
     "textarena",
-    "nltk"
+    "nltk",
+    "bleurt-pytorch"
 ]
 
 docs = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,10 @@ dependencies = [
     "datasets>=3.6.0",
     "transformers",
     "nest-asyncio>=1.6.0",
+    "aiohttp>=3.9.0",
+    "fastapi>=0.104.0",
+    "uvicorn>=0.24.0",
+    "pydantic>=2.5.0",
 ]
 
 [project.optional-dependencies]
@@ -48,7 +52,13 @@ all = [
     "smolagents>=1.15.0",
     "textarena",
     "nltk",
-    "bleurt-pytorch"
+    "bleurt-pytorch",
+    "tensorflow>=2.13.0"
+]
+
+bleurt = [
+    "bleurt-pytorch",
+    "tensorflow>=2.13.0"
 ]
 
 docs = [
--- a/verifiers/examples/bleurt_example.py
+++ b/verifiers/examples/bleurt_example.py
@@ -1,0 +1,167 @@
+"""
+Example usage of BleurtRubric for semantic similarity evaluation.
+
+This example demonstrates how to use BLEURT for text summarization tasks,
+where semantic adequacy is more important than exact string matching.
+"""
+
+from datasets import load_dataset
+import verifiers as vf
+
+# Check if BLEURT is available
+try:
+    from verifiers.rubrics import BleurtRubric
+    BLEURT_AVAILABLE = True
+except ImportError:
+    BLEURT_AVAILABLE = False
+    print("BLEURT not available. Install with: pip install bleurt tensorflow")
+
+
+def main():
+    if not BLEURT_AVAILABLE:
+        return
+    
+    print("Setting up BLEURT summarization example...")
+    
+    # Load a small dataset for demonstration
+    # For real use, consider using cnn_dailymail or similar
+    dataset = load_dataset('wikipedia', '20220301.simple', split='train[:50]')
+    
+    # Transform dataset for summarization task
+    def transform_for_summarization(examples):
+        return {
+            'question': [f"Summarize this article in 2-3 sentences:\n\n{text[:500]}..." 
+                        for text in examples['text']],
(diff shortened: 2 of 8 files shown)


willccbb:
Thanks for the suggestion! Do you know of any works comparing BLEURT to more lightweight metrics like ROUGE or difflib-based metrics? I am generally hesitant to add metrics which involve loading additional models directly into memory in the same process, as this can interfere with the expectations of the trainer in multi-GPU settings. 

Would you be interested in reworking the PR to support an API inference client + auxiliary hosting of an async endpoint? I am also planning on soon adding an EmbedRubric class, which may be a good way to support BLEURT.

rkbadhan:
Hi Will. I was going through some of work like BLUERT paper (Sellam et al., ACL 2020 https://arxiv.org/pdf/2004.04696), BERT Score paper (Zhang et al., ICLR 2020 https://arxiv.org/pdf/1904.09675). 

BLEURT paper has extensive benchmarks against BLEU, ROUGE, and surface-level metrics:

WMT Shared Task Results:

-BLEURT vs BLEU: 49-92% better correlation with human judgments
-BLEURT vs chrF++ (character-level): 58% improvement
-BLEURT vs METEOR (word-level): 37% improvement

The BERTScore paper also - shows similar patterns with 30-100% improvements over lightweight metrics, particularly for segment-level evaluation. 
I did not find head to head comparison of difflib and ROUGE. This study with respect to verifier would be another interesting work!

I think going by this, we can use these type of metrics for accuracy gain.

Also, I'd be very interested in reworking PR as an **API inference client + auxiliary hosting of an async endpoint**. Just to clarify the architecture - you're asking for both the auxiliary BLEURT hosting service, an AND a generic ApiRubric foundation that could later support EmbedRubric, correct? I want to make sure I'm building the right abstraction layer

willccbb:
I am probably going to hold off on merging anything like this for now, sorry. Generally I'm aiming to keep the repo examples very streamlined, and will expand them once I'm confident that they represent the "best" way to do something with the repo. You're of course welcome to use this for your own purposes; the repo is designed to be highly extensible, but not all extensions need to end up in the main branch. I don't want to add tensorflow as a dependency in "all", and would prefer if the first self-contained self-hosted embedding example to be something more standard/well-established as a similarity metric.

PR #93, opened 2025-06-19
Some fixes

tcapelle:
Hey Will, I am playing with the verifiers lib and it's a nice piece of code.
- Fixed some circular imports
- Added `_rewards` to the expected rewards naming
- Added minimal weave support to see the rollouts (better than the tables)
weave


tcapelle:
if you don't call weave.init, the weave.op are no-calls, but yeah, you need it as deps.

tcapelle:
I had some issues with the rewards being casted as long, that's why I casted them to float.

PR #134, opened 2025-07-11
custom types, rework single+multiturn env shared logic, silence nltk …

willccbb:
…downloads in TA, additional minor changes


verifiers/__init__.py
```diff
@@ -1,7 +1,18 @@
...
+
+# typing aliases
+MessageType = Literal['chat', 'completion']
+ModelResponse = Union[Completion, ChatCompletion, None]
+ChatMessageField = Literal['role', 'content']
+ChatMessage = Dict[ChatMessageField, str]
+Info = Dict[str, Any]
+State = Dict[str, Any]
+SamplingArgs = Dict[str, Any]
 RewardFunc = Callable[..., float]
```
mikasenghaas: definitely a step in the right direction, long term my vote would still be with typing the environments io at the very least. like it doesn't help me that much to know that info and state are both `Dict[str, Any]`. what i really want to be able to do as a new user of verifiers is to `outputs.` and then get auto-completions for fields with type guarantees for

mikasenghaas: also i think the right place for these to live longterm is in a `types.py` somewhere. could be global or local, depending on what you prefer

verifiers/envs/environment.py
```diff
@@ -176,26 +183,26 @@ async def get_model_response(self,
                 messages=prompt, # type: ignore
                 **sampling_args
             )
-            return response.choices[0].message.content, response
```
mikasenghaas, line 179: yea nice

verifiers/envs/singleturn_env.py
```diff
@@ -1,42 +1,30 @@
...
+                     messages: List[ChatMessage],
+                     state: State,
+                     **kwargs: Any) -> bool:
+        if len(state['responses']) > 0:
+            return True
+        return False
+
+    def env_response(self, messages: List[ChatMessage], state: State, **kwargs: Any) -> Tuple[ChatMessage, State]:
+        # never called in MultiTurnEnv.rollout
+        return {'role': 'user', 'content': ""}, state
```
mikasenghaas: yep i like this! single turn as instance of multi-turn makes more sense than the other way around

verifiers/envs/textarena_env.py
```diff
@@ -1,15 +1,23 @@
...
 from typing import Tuple, List, Dict, Any
 
 from datasets import Dataset
 import nltk 
-nltk.download('words', quiet=True)
-nltk.download('averaged_perceptron_tagger_eng', quiet=True)
+
+# monkey-patch nltk.download to always be quiet before importing textarena
+_original_nltk_download = nltk.download
+nltk.download = lambda *args, **kwargs: _original_nltk_download(*args, **{**kwargs, 'quiet': True})
```
mikasenghaas, line 10: we love a little monkeypatch

mikasenghaas (approved):
yep, looks good. as mentioned, i'd advocate for validated types for the environments interface at least, but not part of this pr

PR #135, opened 2025-07-13
Resolve multi-GPU scaling in GRPOTrainer with DistributedRepeatSampler

### LINKED ISSUE #112: Question regarding the effective batch size
Hi,

I have a question regarding how effective batch size and the number of global training steps per epoch are computed in trl.GPPOTrainer.

My setup:
- `6000` prompts for training
- `num_generations=8`
- `gradient_accumulation_steps=8`
- `per_device_train_batch_size=8`
- `2` GPUs for training

My current understanding is that the total number of global steps in one epoch can be calculated as:
```
total_global_steps = (#prompts) * num_generations / effective_batch_size
```
where:
```
effective_batch_size = per_device_train_batch_size * num_processes * gradient_accumulation_steps
```
Plugging in my settings:
```
total_global_steps = 6000 * 8 / (8 * 2 * 8) = 375
```
However, I noticed that the logs (e.g., on wandb) show that one epoch actually corresponds to `750 global steps` in my case, which is double what I expected.

Could you clarify how the **effective batch size** and **total steps per epoch** are computed? Am I misunderstanding how batches are constructed in `GPPOTrainer`?

Thanks for your help!

MagellaX:
This pull request addresses the incorrect computation of effective batch size and total training steps per epoch in multi-GPU environments, as reported in Issue #112. The root cause is that the original RepeatSampler processes the full dataset on each GPU without sharding, leading to duplicated prompt handling and step counts that do not scale with the number of processes.
To resolve this, a new DistributedRepeatSampler is introduced, which extends RepeatSampler by sharding the dataset indices across processes before applying chunking and repetition logic. Key adjustments include:
Sharding indices using the process rank and number of replicas to ensure each GPU handles a unique subset of prompts.
Scaling the batch_size parameter to maintain consistency with GRPOConfig's generation_batch_size calculations.
Ceiling division for num_samples to balance workloads across processes.
This ensures correct scaling (e.g., halving step counts when doubling GPUs) while preserving identical behavior for single-GPU setups (num_replicas=1). The change is isolated to the sampler in grpo_trainer.py and integrates with existing async data loading and DeepSpeed/ZeRO configurations without introducing regressions.
Closes #112.


willccbb:
Ah, I just pushed some updates to main which should resolve this -- are you able to test? My runs are working cleanly across epochs with multiple GPUs for training. 

Also, not sure the approach in this PR is correct -- RepeatSampler is already multi-device aware via repeat_count, which depends on generation_batch_size, which is computed using num_processes in GRPOConfig

MagellaX:
> Ah, I just pushed some updates to main which should resolve this -- are you able to test? My runs are working cleanly across epochs with multiple GPUs for training.
> 
> Also, not sure the approach in this PR is correct -- RepeatSampler is already multi-device aware via repeat_count, which depends on generation_batch_size, which is computed using num_processes in GRPOConfig

My understanding is that while repeat_count incorporates num_processes for scaling repetitions, the original RepeatSampler still processes the full dataset on each process without sharding, which leads to duplicated prompts and non-scaling step counts (as observed in the issue). The DistributedRepeatSampler adds explicit sharding of indices before repetition to ensure unique prompts per process. If testing confirms the issue is resolved in your updates, I'll make sure to close this PR accordingly...
But pls just add me as a contributor to the repo

willccbb:
yes, the preprocessing occurs on each node so that each process has a shared view of the dataset (note that only the main process handles submission of requests to the environment), but batches are sliced per-process for actual training

RepeatSampler is inherited from the TRL GRPO trainer, which works identically (for single-turn batch-only completions)

MagellaX:
> yes, the preprocessing occurs on each node so that each process has a shared view of the dataset (note that only the main process handles submission of requests to the environment), but batches are sliced per-process for actual training
> 
> RepeatSampler is inherited from the TRL GRPO trainer, which works identically (for single-turn batch-only completions)

Hmm... thanks for clarifying the processing flow, it's clear that the main-process submission and slicing are designed to streamline things, and I see how that works well for single-turn cases.

i think to address this minimally, integrate sharding into RepeatSampler by adding rank/num_replicas params and sharding indices in iter, ensuring unique prompts per process without a subclass... like this

```python
class RepeatSampler(Sampler):
    def __init__(
        self,
        data_source: Sized,
        mini_repeat_count: int,
        batch_size: int = 1,
        repeat_count: int = 1,
        shuffle: bool = True,
        seed: Optional[int] = None,
        rank: int = 0,
        num_replicas: int = 1,
    ):
        self.data_source = data_source
        self.mini_repeat_count = mini_repeat_count
        self.batch_size = batch_size
        self.repeat_count = repeat_count
        # Ceiling division to balance samples across replicas
        self.num_samples = (len(data_source) + num_replicas - 1) // num_replicas
        self.shuffle = shuffle
        self.seed = seed
        self.rank = rank
        self.num_replicas = num_replicas

        if shuffle:
            self.generator = torch.Generator()
            if seed is not None:
                self.generator.manual_seed(seed)

    def __iter__(self):
        if self.shuffle:
            indexes = torch.randperm(len(self.data_source), generator=self.generator).tolist()
        else:
            indexes = list(range(len(self.data_source)))

        # Shard the indexes across replicas
        sharded_indexes = indexes[self.rank : len(indexes) : self.num_replicas]

        # Split into fixed-size batches
        chunks = [
            sharded_indexes[i : i + self.batch_size]
            for i in range(0, len(sharded_indexes), self.batch_size)
        ]

        # Keep only full batches
        chunks = [chunk for chunk in chunks if len(chunk) == self.batch_size]

        for chunk in chunks:
            for _ in range(self.repeat_count):
                for index in chunk:
                    for _ in range(self.mini_repeat_count):
                        yield index

    def __len__(self) -> int:
        return self.num_samples * self.mini_repeat_count * self.repeat_count
```

And in get_train_sampler:
```python
        return RepeatSampler(
            data_source=self.train_dataset,  # type: ignore
            mini_repeat_count=self.num_generations,
            batch_size=self.generation_batch_size // self.num_generations // self.accelerator.num_processes,
            repeat_count=self.num_iterations * self.gradient_accumulation_steps,
            shuffle=self.shuffle_dataset,
            seed=self.args.seed,
            rank=self.accelerator.process_index,
            num_replicas=self.accelerator.num_processes,
        )
```
This scales batch_size appropriately while sharding indices upfront. I can update the PR with this for review if it aligns, any thoughts???

MagellaX:
> I'm still not convinced that there is a bug with the current approach, which is borrowed directly from TRL. Can you explain more about the issue you're seeing?
> 
> RepeatSampler is used globally, not on a per-process basis, and data is broadcast to separate processes further downstream. The entire rollout generation flow is single-process.

Hmm... I get the TRL reference you're saying. However, in the current distributed setup, each process instantiates its own DataLoader and RepeatSampler, so the sampler is not actually global-it's local to each process...
 This means every process iterates over the full dataset independently, resulting in duplicated prompt sampling and a total number of global steps that does not scale with the number of processes. This is observable in practice: when running with 1 vs. 2 GPUs, the step count remains constant, rather than halving as expected if the workload were properly sharded. While the rollout generation logic is single-process for submission, the DataLoader feeding it is still per-process, so without explicit sharding in the sampler, we get redundant data loading and non-scaling behavior. This is a subtle but critical distinction from the intended TRL pattern, and is the root cause of the scaling issue described in issue-112.

PR #148, opened 2025-07-18
Fix: Prevent execution of tool tags nested inside other XML tags

PastaPastaPasta:
## Description
This PR prevents the execution of `<tool>` tags that are nested inside other XML tags like `<think>` or `<answer>`. Currently, the XML parser will find and execute any tool tag in the content, even if it's nested inside reasoning or thinking sections. This change ensures only top-level tool tags are executed.

This isn't normally an issue, as most hosted/closed models won't provide the thinking tokens. However, when thinking tokens are provided, the thinking may contain tool calls the model has no intent of actually calling. From our understanding of how these models generally work, they assume none of the thinking tokens are being acted upon or viewed. We check that tool calls we are running are at the top level.

## Type of Change
- [ ] Documentation update
- [x] Test improvement

## Testing
- [x] All existing tests pass

### Test Coverage
Added 20 new tests covering:
- Nested tool tags in various contexts (think, answer, deeply nested)
- Multiple top-level tool tags (only first executes)
- Edge cases (empty tags, malformed XML, whitespace handling)
- Unicode content support
- Backward compatibility preserved

## Checklist

### Implementation Details
- Added `top_level_only` parameter to `XMLParser.parse()` (default `False` for backward compatibility)
- `ToolEnv.env_response()` now uses `top_level_only=True` when parsing for tool tags
- The implementation checks if a tag is nested by counting open/close tags before each match

### Example Behavior
Before this change:
```xml
<think>
Let me use this tool: <tool>{"name": "add", "args": {"a": 1, "b": 2}}</tool>
</think>
```
Would execute the tool even though it's inside a thinking block.

After this change:
- The nested tool is ignored
- Only top-level `<tool>` tags are executed
- Error message returned if no top-level tool found

### DIFF
--- a/tests/test_tool_env.py
+++ b/tests/test_tool_env.py
@@ -1,0 +1,261 @@
+"""Tests for the ToolEnv class."""
+
+import pytest
+from unittest.mock import MagicMock
+from verifiers import XMLParser
+from verifiers.envs.tool_env import ToolEnv
+
+
+class TestToolEnv:
+    """Test cases for the ToolEnv class."""
+
+    @pytest.fixture
+    def mock_tools(self):
+        """Create mock tools for testing."""
+        def add_tool(a: int, b: int) -> int:
+            """Add two numbers together.
+            
+            Args:
+                a: First number
+                b: Second number
+                
+            Returns:
+                int: Sum of a and b
+            """
+            return a + b
+
+        return [add_tool]
+
+    @pytest.fixture
+    def tool_env(self, mock_tools, sample_dataset):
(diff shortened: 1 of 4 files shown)


willccbb:
Did a major overhaul of ToolEnv to use native tool-call parsing by default, relying on XMLParser for this was always a bit of a hack, and this also lets us do parallel tool calling much more easily. 

Would you be interested in doing a separate PR with these fixes for XMLParser more broadly, to address the nested tag issue? Otherwise will close; thanks!

PR #156, opened 2025-07-23
math_python eval fix

ntoshev:


willccbb:
the standalone eval scripts are now deprecated in favor of a single one which takes the env id as an arg (to avoid having to maintain all these kinds of little things :))

PR #184, opened 2025-08-03
Removing mutable default arguments to follow pep guidelines

### LINKED ISSUE #183: Use of mutable default arguments
## Description
Across several parts of the codebase, mutable objects (e.g., lists, dictionaries, class instances) are used as default argument values in function or class method definitions. Using mutable default arguments is considered an anti-pattern in Python and can lead to unintended behavior, especially state sharing between multiple instances or calls.

Mutable default arguments are evaluated only once at function-definition time, not at each function call. Thus, all calls to the function share the same mutable object, potentially leading to bugs and unexpected side effects.

An example of that is the method `Rubric.__init__`
```python
def __init__(
        self,
        funcs: List[RewardFunc] = [],   # mutable object, all instances share the same
        weights: List[float] = [],      # mutable object, all instances share the same
        parser: Parser = Parser(),      # mutable object, all instances share the same
        parallelize_scoring: bool = True,
        **kwargs,
    ):
```

## Steps to reproduce unexpected behavior
here is an example that illustrates the issue 
```python
import verifiers as vf

class FirstRubric(vf.Rubric): ...
class SecondRubric(vf.Rubric): ...

first_rubric = FirstRubric()
another_first_rubric = FirstRubric()
second_rubric = SecondRubric()

print(first_rubric.get_reward_funcs())              # Output: []
print(another_first_rubric.get_reward_funcs())      # Output: []
print(second_rubric.get_reward_funcs())             # Output: []

def dummy_func(completion, answer, **kwargs) -> float:
    return 1.0 if completion == answer else 0.0
first_rubric.add_reward_func(dummy_func, weight=0.5)   # adding the reward func to one instance

print(first_rubric.get_reward_funcs())              # Output: [<function dummy_func at 0x7bff15cb8ae0>]
print(another_first_rubric.get_reward_funcs())      # Output: [<function dummy_func at 0x7bff15cb8ae0>]
print(second_rubric.get_reward_funcs())             # Output: [<function dummy_func at 0x7bff15cb8ae0>]

```

## Expected behavior
Each call or instance should get its own independent default object, preventing shared state across calls or instances. Therefore, the above code snippet should've produced this output
```python
print(first_rubric.get_reward_funcs())              # Output: [<function dummy_func at 0x7bff15cb8ae0>]
print(another_first_rubric.get_reward_funcs())      # Output: []
print(second_rubric.get_reward_funcs())             # Output: []
```

## Recommended solution
Replace mutable default arguments with None, then initialize the mutable object inside the function/method:

Here is an example of how the above code of the `Rubric.__init__` can be updated to avoid this issue.

```python
def __init__(
        self,
        funcs: List[RewardFunc] | None = None,
        weights: List[float] | None = None,
        parser: Parser | None = None,
        parallelize_scoring: bool = True,
        **kwargs,
    ):
        self.reward_funcs = funcs or []
        self.reward_weights = weights or []
        self.parser = parser or Parser()
        
```

This is a common best practice recommended by the official Python documentation, Python community standards (PEP 8), and many style guides. Fixing this consistently throughout the codebase will reduce the risk of future bugs and improve code maintainability.

## Happy to Help
If you agree with this change, I’d be glad to prepare a pull request to update the relevant parts of the codebase. Just let me know if you’re open to this change, or if you have any preferences on how you’d like it done!

I'm a big fan of this project. I came across your talk in the AI Engineer World's Fair while I was working on training an LLM to play Contexto. Thanks again for making this project available!

MarwanMashra:
## Type of Change

- [ ] Documentation update
- [ ] Test improvement

## Testing

- [x] All existing tests pass

### Test Coverage

- Current coverage: 34%
- Coverage after changes: 34%

## Checklist

### Main changes
1. got rid of all mutable default arguments.
2. updated type hints to reflect pep standards (replacing Dict, List, Tuple with dict, list, and tuple, and removing Optional and Union)
3. removed redundant attribute assignment in subclass __init__
4. made `tasks` and `infos` required in some methods like `run_rollouts` (more details below).

### Future concerns
I noticed a common pattern in the code that consists of passing around separate lists and zipping them to reconstruct an object instead of grouping related elements in a single object. This can cause a lot of future bugs, one of which I fixed in this PR by making `tasks` and `infos` required in some methods like `run_rollouts`. Using `zip` doesn't allow for any list to be optional, since you iterate over all lists until hitting the shortest. So having `tasks` and `infos` default to an empty list wouldn't have allowed us to enter the loop.

I discuss this further in this [Discussion](https://github.com/willccbb/verifiers/discussions/185)

closes #183


willccbb:
thanks! LGTM

PR #197, opened 2025-08-08
Fix IndexError in eval script when using multiple rollouts per example

JannikSt:
seeing this error with my custom environment. Can also reproduce when running eval on reverse text env:  

**Bug:** IndexError when running evaluations with multiple rollouts per example due to incorrect array indexing formula in eval.py.

**Root Cause:** The eval script uses `(i * r) + j` for indexing results, but should use `(i * n) + j` where:
- `i` = rollout index (0, 1, 2...)  
- `n` = num_examples
- `r` = rollouts_per_example
- `j` = example index (0 to n-1)

**Example failure case:**
- Command: `vf-eval env -n 2 -r 3` (2 examples, 3 rollouts each = 6 results)
- When i=1, j=0: tries to access `results.reward[1*3+0] = results.reward[3]` 
- But should access `results.reward[1*2+0] = results.reward[2]`

**Fix:** Corrected indexing formula and added bounds checking for robustness.

### DIFF
--- a/verifiers/scripts/eval.py
+++ b/verifiers/scripts/eval.py
@@ -95,7 +95,7 @@ Please specify the model name (-m), API host base URL (-b), and API key variable
     r = rollouts_per_example
     for i in range(r):
         # rounded to 3 decimal places
-        trials = [round(results.reward[(i * r) + j], 3) for j in range(n)]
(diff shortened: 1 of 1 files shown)


willccbb:
thanks! good catch

PR #214, opened 2025-08-18
[Draft]: Add policy entropy metric

SamComber:
Would be good to have an option to understand potential policy entropy collapse. Exhaustion is typically indicative of growing determinism in token decoding and diminishes exploration behaviour (see https://arxiv.org/pdf/2505.22617 paper)

image

### WIP

Need to optimise so we don't create `B*L*V` intermediate tensors inside the entropy calc (i.e. basically 3x the memory required right now)... will likely batch it and put a warning flag when calculate entropy is enabled ??

### DIFF
--- a/verifiers/trainers/grpo_trainer.py
+++ b/verifiers/trainers/grpo_trainer.py
@@ -22,7 +22,7 @@ from transformers.trainer_callback import TrainerCallback
 from transformers.trainer_utils import seed_worker
 from trl.models import create_reference_model, prepare_deepspeed
 from trl.trainer.callbacks import SyncRefModelCallback
-from trl.trainer.utils import disable_dropout_in_model, pad, selective_log_softmax
+from trl.trainer.utils import disable_dropout_in_model, pad, selective_log_softmax, entropy_from_logits
 
 from verifiers import Environment
 from verifiers.trainers.async_batch_generator import AsyncBatchGenerator, BatchRequest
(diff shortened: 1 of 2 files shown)


verifiers/trainers/grpo_trainer.py
```diff
@@ -731,18 +760,31 @@ def _get_per_token_logps(
...
             input_ids_batch = input_ids_batch[:, -logits_to_keep:]
             # For transformers<=4.48, logits_to_keep argument isn't supported, so here we drop logits ourselves.
             # See https://github.com/huggingface/trl/issues/2770
             logits = logits[:, -logits_to_keep:]
             # Divide logits by sampling temperature.
             # See https://huggingface.co/blog/the_n_implementation_details_of_rlhf_with_ppo#policy-training-implementation-details
             logits = logits / self.temperature
+
+            if compute_entropy:
+                entropy = entropy_from_logits_memory_efficient(logits, chunk_size=32)
```
qgallouedec: hey! if you want something aligned with trl, we have `entropy_from_logits` that you can directly import

https://github.com/cyyever/trl/blob/2ddf8010881e3bdf215be28ed6fd6f3a8ae2bcf5/trl/trainer/utils.py#L1469

it does the same as here

SamComber: Oh this is perfect! Thanks for the suggestion

SamComber: Still getting OOMs with chunk size 1 here in places where we definitely should not be hmm @qgallouedec , does `entropy_from_logits` internally need to `torch.no_grad()` perhaps?

Issue #211, opened 2025-08-16
Cannot specify `None` as `max_tokens` in eval script


mikasenghaas:
Would be nice to be able to specify `None` as `max_tokens` when evaling against a local server to always go until max context. Running into issues rn where my inference server is hosted with 2k context, but if I specify 2k max tokens I get errors bc prompt + 2k is more than server ctx..

mikasenghaas:
from my side: go for it. @willccbb is this smth you would be happy to merge?

DarkStarStrix:
You can check it out. spent a few hours there; there was a quirk I was not able to quite figure out

willccbb:
Closing as resolved by #241

PR #219, opened 2025-08-19
feat(vllm): Add dynamic max_tokens calculation for VLLMClient completion

DarkStarStrix:
This non-breaking change enables users to specify max_tokens=None in VLLMClient completion calls. The client will automatically compute the maximum allowable tokens based on the model’s context length and the prompt size, streamlining local vLLM evaluations without manual token math.

It addresses #211.
@willccbb
@mikasenghaas

The new /get_model_config endpoint in vllm_server.py exposes max_model_len. VLLMClient now accepts a tokeniser, fetches model config on init, and overrides completions. Create to calculate max_tokens when None Dedicated tests in tests/test_vllm_client_server.py cover all edge cases with 100% coverage; existing tests remain green. Updated tests/README.md with vLLM testing guidelines Project style guidelines were followed; code was reviewed and commented on. All dependent changes merged

Note: A benign RuntimeWarning about an unclosed coroutine may appear during tests; it does not affect functionality.

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [x] Test improvement

## Testing

- [x] All existing tests pass

### Test Coverage

- Current coverage: 100%
- Coverage after changes: 100%

## Checklist

## Additional Notes
None

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -109,16 +109,14 @@ async def test_multiturn_conversation(mock_multiturn_env):
 
 ## vLLM Client/Server Testing
 
-The current test suite does **not** directly test the vLLM client/server integration or the new logic for max_model_len handling. To ensure these features are covered:
+The test suite includes unit tests for the `VLLMClient` in `tests/test_vllm_client_server.py`. These tests ensure that the client's custom logic for handling `max_tokens` and interacting with a vLLM server is correct.
 
-- Add a dedicated test file (e.g., `test_vllm_client_server.py`) that:
-  - Mocks vLLM server endpoints such as `/get_model_config`.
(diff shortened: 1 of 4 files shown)


willccbb:
The PR seems pretty intrusive, and requires a few breaking and/or intricate changes that I'd like to avoid. If we want to allow `max_tokens` to not be set, we should do this in a way which doesn't require any dynamic computation of model sequence length. 

The generic OpenAI clients treat max_tokens as Optional; what's the problem with simply allowing this to be set as None, or potentially leveraging the openai NotGiven type which is used by the client as a fallback?

PR #236, opened 2025-08-25
Fix: Stop tool execution on failure in ToolEnv

Mirza-Samad-Ahmed-Baig:
This PR fixes a critical bug in the ToolEnv class where tool execution would continue even after a tool call failed. This could lead to unexpected and incorrect behavior of the agent. The env_response function has been modified to stop the execution osubsequent tool calls if a tool call fails. The call_tool function has also been modified to return a boolean value indicating whether the tool call was successful or not.

Type of Change

Testing

   - [x] All existing tests pass
   - [x] Tests have been run locally with python -m pytest tests/

Checklist

### DIFF
--- a/verifiers/envs/tool_env.py
+++ b/verifiers/envs/tool_env.py
@@ -31,7 +31,7 @@ class ToolEnv(MultiTurnEnv):
 
     def call_tool(
         self, tool_name: str, tool_args: dict, tool_call_id: str, **kwargs
-    ) -> Message:
+    ) -> tuple[Message, bool]:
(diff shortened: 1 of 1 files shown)


willccbb:
I think it's still too opinionated for ToolEnv, which is intended to be a very simple ReAct pattern that doesn't try to be too clever about behavior shaping. Users are welcome to extend ToolEnv with different is_completed conditions if they'd like (which is where this logic should go, not call_tool), but doesn't feel like a core implementation detail to me, and it's one more thing that users have to reason about when looking at the code.

PR #248, opened 2025-08-26
 Refactor: Improve error handling for missing endpoints file

SyedaAnshrahGillani:
This pull request improves the error handling in the eval.py script. The script now raises a FileNotFoundError if the endpoints.py file is not found at the specified path. This provides a clearer error message to the user and prevents the script from continuing with an invalid state.

### DIFF
--- a/verifiers/scripts/eval.py
+++ b/verifiers/scripts/eval.py
@@ -41,18 +41,18 @@ def eval_environment(
         else:
             endpoints_file = endpoints_path_obj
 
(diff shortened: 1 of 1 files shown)


willccbb:
Hmm not having an `endpoints.py` file isn't "incorrect" really, we don't want to hard-fail if so. It just means that we fall back to OpenAI as default. I like the reorg, mind updating to catch the FileNotFoundError so we just log the warning and continue?

PR #273, opened 2025-08-29
fix(tui): escape user content to prevent markup injection issues

srthkdev:
- Added safe_escape function to sanitize user input in rich markup contexts
- Applied escaping to roles and content in format_prompt_or_completion
- Escaped function names and arguments in tool call displays
- Escaped environment, model, run IDs, and other metadata in various screens
- Ensured all dynamic text rendered in labels and options is safely escaped
- Prevented potential rendering errors or security issues from unescaped markup

Here's the PR description for your rich markup escaping implementation:

## Description

Implement `rich.markup.escape` for vf-tui to prevent display issues when user-generated content contains square brackets that Rich might interpret as markup commands (fixes #267).

**Key changes:**
- Added `rich.markup.escape` import and `safe_escape()` helper function
- Applied escaping to all user-generated content in the TUI
- Preserved intentional UI markup for proper formatting
- Ensures robust display of prompts, completions, environment names, and metadata

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Test improvement

## Testing

- [x] All existing tests pass
- [ ] New tests have been added to cover the changes
- [x] Tests have been run locally with `python -m pytest tests/`
- [ ] Manual testing of TUI with problematic markup content
- [ ] Verified UI markup preservation and user content escaping

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

## Additional Notes

**What this fixes:**
- Prevents TUI display corruption when user content contains `[bold]`, `[red]`, or other Rich markup patterns
- Ensures consistent display regardless of user input content
- Maintains proper TUI formatting and styling

**Implementation details:**
- `safe_escape()` function safely wraps `rich.markup.escape`
- Applied to all user content: prompts, completions, environment names, model names, run IDs, metadata, answers, tasks, and info
- UI markup (like `[b]Environment:[/b]`) is preserved for proper styling
- No performance impact - escaping only applied where needed

**Files modified:**
- `verifiers/scripts/tui.py` - Main implementation

**Ready for production use** - comprehensive testing confirms the implementation works correctly and maintains backward compatibility.

PR #284, opened 2025-08-31
fix markuperror in completion

jalexine:
The TUI currently crashes with a MarkupError when displaying completions. This happens because raw text containing square-bracket indices (e.g. latvec[0]) is passed to a Static widget with markup enabled, and Rich misinterprets these as markup tags.

I ran into this issue when trying to view the results of my environment with vf-tui, since assistant.tools outputs often contain brackets and structured args, while the assistant/user messages are assumed to be plain text.

This PR proposes a fix: render completions/prompts using rich.text.Text instead of raw markup strings, so user/model content is always treated as literal text (no markup parsing). Only the role labels (assistant:, tool result:, etc.) are styled. This avoids markup crashes while preserving the intended formatting.

# Testing
I tried locally with different runs
I ran local tests confirming no crashes with bracketed prompts/completions, correct tool_calls parsing (JSON-string & dict), and properly escaped metadata with bold labels preserved.

## Before 
Screenshot 2025-08-31 at 2 02 39 PM

## After
Screenshot 2025-08-31 at 2 03 30 PM

### DIFF
--- a/verifiers/scripts/tui.py
+++ b/verifiers/scripts/tui.py
@@ -140,7 +140,7 @@ def format_prompt_or_completion(prompt_or_completion) -> Text:
             content = str(msg.get("content", ""))
             # Style by role
             if role == "assistant":
-                out.append(f"{role}: ", style="bold")
+                out.append("assistant: ", style="bold")
                 out.append(content)
             elif role == "tool":
                 out.append("tool result: ", style="bold dim")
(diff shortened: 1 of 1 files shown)


willccbb:
Should be fixed by #273 -- or is this still an issue after the update?

willccbb (approved):

willccbb:
Gonna approve for now, but @ob1-s could you share a link/gist with an eval result that fails for you so I can try and reproduce?

ob1-s:
> Gonna approve for now, but @ob1-s could you share a link/gist with an eval result that fails for you so I can try and reproduce?

https://gist.github.com/ob1-s/1adc19d606e6859a039f80646254a4f3

willccbb:
Added more robust escaping, your acebench-normal outputs render just fine for me now.

PR #325, opened 2025-09-13
ArCHer‑style Hierarchical RL: Utterance‑Level Critic, Replay Buffer, and Advantage Shaping for GRPO

MagellaX:
Verifiers already provide efficient GRPO over token-level trajectories and rich multi-turn environments. This PR adds a hierarchical layer to better optimize long-horizon objectives and delayed rewards by learning a value function at the utterance (turn) level.

## What’s Included
- **Utterance-level critic:** Off-policy value estimation over conversation prefixes using a lightweight MLP head on frozen LM input embeddings.  
- **Replay buffer:** Stores per-utterance transitions and supports off-policy updates.  
- **Advantage shaping:** Optional coupling modes to combine critic baselines with GRPO’s group-relative baseline.  
- **Turn-level rewards:** Hooks to consume env-provided `turn_rewards` for proper TD/MC targets when available; falls back to discounted terminal reward otherwise.  
- **Save/Load:** Critic head and optimizer state persisted under `hierarchical/` in checkpoints; optional load via config.  

## How It Works
- During async generation, build utterance-level prefixes for each assistant turn and compute returns (from `turn_rewards` or discounted terminal reward). Store in a replay buffer and train the critic off-policy in the background.  
- For actor-critic coupling, predict baselines `V(prompt)` and shape GRPO advantages:  
  - **critic:** `adv = reward – V(prompt)`  
  - **blend:** `adv = (1 – α)(reward – group_mean) + α(reward – V(prompt))` with optional std scaling.  
- Training remains distributed and async; critic updates happen on the main process without touching the policy weights.  

## Backward Compatibility
- Disabled by default. Standard GRPO remains unchanged unless `enable_hierarchical=True`.  

## Key Configs (new)
- **Enable & params:** `enable_hierarchical`, `hier_replay_size`, `hier_minibatch_size`, `hier_train_steps_per_batch`, `hier_lr`, `hier_weight_decay`, `hier_gamma`, `hier_max_seq_length`, `hier_value_hidden_size`.  
- **Coupling:** `hier_advantage_mode = none | critic | blend`, `hier_advantage_alpha`.  
- **Persistence:** `hier_load_path`.  
- **Label:** `low_level_algo` (`grpo` or `ppo`, for future branching).


MagellaX:
@willccbb any takes?

willccbb:
probably not something we'd consider merging without plenty of experiments showing that this is 
1) helpful + implemented correctly
2) non-breaking

diff is also very large, would prefer an initial PR to be super minimal + sketch a roadmap for which important features are added in which stages, along with why the design choices were made in a particular way. might make more sense as a fully standalone trainer that doesn't need to live in the verifiers repo

willccbb:
ideal first PR for this would:
- be fully non-breaking / backward compatible
- be a few hundred lines tops
- include experiments of GRPO both before/after to confirm non-breaking, and of the new method to show that it is useful in a meaningful way

it's the kind of thing i plan to add eventually, but haven't had time for yet, and don't want to merge something here unless i'm very confident that it's done adequately

MagellaX:
> ideal first PR for this would:
> - be fully non-breaking / backward compatible
> - be a few hundred lines tops
> - include experiments of GRPO both before/after to confirm non-breaking, and of the new method to show that it is useful in a meaningful way
> 
> it's the kind of thing i plan to add eventually, but haven't had time for yet, and don't want to merge something here unless i'm very confident that it's done adequately

Don't worry brother, i'll do it

PR #360, opened 2025-09-22
finish_reason=length if env caused truncation

cat-state:
## Description
Currently, when the policy's output is truncated by length, the finish_reason is set to length by vLLM, and this pipes through to prime-rl where it can decide how to handle truncated completions. However, if the env response is too long, this causes verifiers to send a too long request to vllm, which returns a 400 error. I argue that in this case, the correct thing to do is to treat it as a truncated completion, rather than to end the training run. This is because the env response size is controlled by the policy, and it exceeding its ctx length via generating too many tokens or asking the env for too much should be handled similarly. This shouldn't be a breaking change as it only affects cases which would have crashed. 

Unfortunately, openai didn't deign to design more informative error codes into their API, which vllm reuses, so we have no way of seeing exactly what kind of 400 (BadRequestError) it is, aside from string matching on the text :/

## Type of Change

- [ ] Documentation update
- [ ] Test improvement

## Testing

## Checklist


verifiers/envs/multiturn_env.py
```diff
@@ -82,15 +82,32 @@ async def rollout(
...
+                )
+                state["responses"].append(response)
+            # In case of requesting a too long completion, e.g from a too-long
+            # environment response, we rewrite the error to a truncated completion
+            # prime-rl will then decide how to mask or penalize this
+            except BadRequestError as e:
+                if len(state["responses"]) == 0:
+                    raise e
+                elif e.message.startswith("This model's maximum context length is"):
+                    print(f"rewriting {e} to truncation")
```
mikasenghaas: remove debug print

verifiers/envs/multiturn_env.py
```diff
@@ -82,15 +82,32 @@ async def rollout(
...
+                    message_type=self.message_type,
+                )
+                state["responses"].append(response)
+            # In case of requesting a too long completion, e.g from a too-long
+            # environment response, we rewrite the error to a truncated completion
+            # prime-rl will then decide how to mask or penalize this
+            except BadRequestError as e:
+                if len(state["responses"]) == 0:
+                    raise e
+                elif e.message.startswith("This model's maximum context length is"):
```
mikasenghaas: maybe combine these two into: one: if not first turn AND err  msg starts with ..., artificially truncate. else, reraise. this way we have one less condition and we more explicitly catch the exception we wanna catch

mikasenghaas (commented):
nice! two small comments. did you do any small experiments against a multi-turn env to validate the impl?

cat-state:
> nice! two small comments. did you do any small experiments against a multi-turn env to validate the impl?

I tested it out against my Search-R1 environment which caused some truncations when running with a lower max_model_len but should perhaps do more tests to validate it

mikasenghaas (approved):
lgtm!

willccbb:
thanks! updated to incorporate a new pattern we just added, which should be more stable going forward -- basically, we want to force termination logic to flow through `is_completed` so that envs can reliably use that function to also encapsulate any teardown logic or precomputation of reward results without worrying about early exits (e.g. due to max_turns being hit, or now an overlong prompt.

PR #365, opened 2025-09-23
Add average_reward column to make_dataset

faresobeid:
Tracking the average_reward for each prompt (over rollouts_per_example) is a useful metric both for difficulty filtering in RL and SFT data


samsja (approved):
lgtm, but will let @willccbb check it out :)

willccbb:
Merging for now but will prob want to revisit how we do this later.

PR #376, opened 2025-09-25
Deserialize function tool call argument before applying chat template

mikasenghaas:
## Description

No need to merge as is, but this makes my experiments on `math-python` with Qwen3-Coder chat template work so would be nice to figure out how to properly upstream.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Test improvement

## Testing

- [ ] All existing tests pass when running `uv run pytest` locally.
- [ ] New tests have been added to cover the changes

## Checklist
- [ ] My code follows the style guidelines of this project as outlined in [AGENTS.md](/AGENTS.md)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published

PR #383, opened 2025-09-27
Dynamically adjust token budget during multi-turn rollout

mikasenghaas:
## Description

> Based on #376 and used for RL training experiments in `math-python` environment. Does not need to be merged as is, although I do think the feature is needed in general. 

As far as I could tell the only way to constrain the rollout length in multi-turn is with `max_tokens` which limits the number of tokens *per turn*. However, in most RL training scenarios I want to limit the number of tokens *per rollout*. Doing this on the inference server via `max_model_len` does not work because the training crashes if the multi-turn prompt exceeds the max context. This PR just makes it work for my specific usecase and might not be the right general API but we definitely need a way to constrain the number of tokens per rollout, and ideally also per turn.

## Type of Change

- [ ] Documentation update
- [ ] Test improvement

## Testing

## Checklist
- [ ] My code follows the style guidelines of this project as outlined in [AGENTS.md](/AGENTS.md)

### DIFF
--- a/verifiers/envs/environment.py
+++ b/verifiers/envs/environment.py
@@ -215,13 +215,10 @@ class Environment(ABC):
         if message_type is None:
             message_type = self.message_type
         # Normalize sampling args:
-        # - If max_tokens is provided for chat, rename to max_completion_tokens
         # - Drop any None-valued entries to avoid sending them to the client
         if "max_tokens" in sampling_args:
             if sampling_args["max_tokens"] is None:
                 sampling_args.pop("max_tokens")
-            elif message_type == "chat":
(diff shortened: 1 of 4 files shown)


cat-state:
This PR might be complementary: https://github.com/PrimeIntellect-ai/verifiers/pull/360 in which it lets you limit it with `max_model_len` . Alternatively this approach doesn't require parsing errors, so perhaps this is the better way

mikasenghaas:
@cat-state ah nice, yea we actually discussed internally that we'd go for this approach to avoid the tokenizer dependency. i will review your pr. high prio to get merged quick

Issue #400, opened 2025-10-01
Hard limit number of sandboxes


mikasenghaas:
Specifying `max_concurrent` does not hard limit the number of concurrent sandboxes, e.g. when running with `max_concurrent=128` I typically have > 128 concurrent sandboxes. We probably want for `SandboxEnv` to have a `max_concurrent_sandboxes` or similar which is a hard limit, i.e. we guarantee that there cannot be more concurrent sandboxes.

<!-- Uploading "Screenshot 2025-10-01 at 11.39.54 AM.png"...

Issue #399, opened 2025-10-01
Automaticaly cleanup `SandboxEnv`


mikasenghaas:
It would be nice if we could ensure that all sandboxes used during a run with `SandboxEnv` are always cleaned up, eg. on crash or keyboard interrupt. Right now, I have to log manually `prime sandbox delete -all` to cleanup an interrupted run.

PR #405, opened 2025-10-01
QoL improvements for `SandboxEnv`

mikasenghaas:
## Description

Addresses issues #400 and #399:
- Adds `asncio.Semaphore` to hard control the maximum number of sandboxes in use concurrently. Acquires before creating the sandbox, releases after deleting the sandbox. This was because I observed that just setting the `max_concurrent` during `a_generate` would not keep the actual number of sandboxes active on the platform below this value
- Registers handlers on SIGINT, SIGTERM (using `signal`) and general exception (using `atexit`) to clean up all remaining sandboxes. This should cover most unexpected program exits and automatically cleanup any dangling sandboxes.
- Also makes all sandbox create arguments configurable by the environment (e.g. memory/ disk usage)

Note: There is known issue around the cleanup handler not cleaning up all the sandboxes as there might be some creation requests in-flight that don't get added to the `active_sandboxes` set which is used for cleanup. Made a [separate issue](https://github.com/PrimeIntellect-ai/verifiers/issues/411) for this that we can tackle in another PR. 

## Type of Change

- [ ] Documentation update
- [ ] Test improvement

## Testing

## Checklist

### DIFF
--- a/verifiers/envs/sandbox_env.py
+++ b/verifiers/envs/sandbox_env.py
@@ -4,6 +4,7 @@ import verifiers as vf
 
 try:
     from prime_cli.api.sandbox import (  # type: ignore[import-untyped]
+        AdvancedConfigs,
         AsyncSandboxClient,
         CreateSandboxRequest,
     )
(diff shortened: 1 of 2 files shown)


verifiers/envs/sandbox_env.py
```diff
@@ -1,9 +1,17 @@
+import atexit
+import signal
+import time
+from asyncio import Semaphore
 from typing import Any
 
+from prime_cli.api.client import APIClient
```
willccbb: can we guard this with TYPE_CHECKING?

willccbb:
Exit handling looks good. The total number of sandboxes *should* be controlled by the `max_concurrent` semaphore, any idea why this isn't the case? Related to cleanup? Could we just wait for sandbox to fully terminate inside is_completed? Adding another one seems redundant.

verifiers/envs/sandbox_env.py
willccbb: also why can't we have all the prime_cli imports in the same spot?

willccbb (changes requested):
Let's group the imports, and then also see if we can remove the semaphore as it should be redundant.

verifiers/envs/sandbox_env.py
mikasenghaas: ah yea fair, this an auto import from my lsp, my bad. fixed now

mikasenghaas:
grouped the imports and removed semaphore after confirming that it indeed hard-limits the sandboxes on `vf-eval`. im 100% certain that when training from `prime-rl` (eg in `math-python`) that the total sandbox count in the ui is above `max_concurrent` but seems like this is indeed not an issue in the sandbox env logic but maybe `prime` CLI?

willccbb (approved):

willccbb:
LGTM

PR #465, opened 2025-10-13
Implement environment for Mastermind game

xlr8harder:
## Description
This implements the guessing game Mastermind as a verifier environment.

## Type of Change

- [X] New feature (non-breaking change which adds functionality)
- [ ] Documentation update
- [ ] Test improvement

## Testing

- [X] All existing tests pass when running `uv run pytest` locally.

## Checklist
- [X] My code follows the style guidelines of this project as outlined in [AGENTS.md](/AGENTS.md)
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] Any dependent changes have been merged and published

## Additional Notes
My first version had a bug because I tried to maintain game state in env_response, but env_response is never called if the model reaches the turn budget, so the final game state is not calculated and I was returning the wrong rewards.

I think the wordle env may have this bug as well.  It looks to me like the partial_credit_reward_func function makes the assumption that a final env_response has been run, but it may not be if the model solves on the final turn:
https://github.com/PrimeIntellect-ai/verifiers/blob/main/environments/wordle/wordle.py#L39

I worked around it by updating the game state in is_complete, but it's a little wasteful because it's called twice per turn.

I think it would be cleaner if there was a separate process_response method that is called for every model response, including on the final turn.  This would be beneficial to MultiTurnEnv's with complex internal state and allow clearer separation of concerns between the reward functions and calculating environment state, since they can reliably just calculate from stored environment state instead of reparsing messages.

### DIFF
--- a/environments/mastermind/README.md
+++ b/environments/mastermind/README.md
@@ -1,38 +1,58 @@
-Mastermind (Multi-Turn)
+# mastermind
 
 Mastermind is a classic deductive reasoning game, first analyzed algorithmically by Donald Knuth, who showed that the standard 4×6 version can always be solved in at most five guesses using a minimax search strategy. For slightly larger boards, exact worst-case bounds are known only in a few cases, and the general problem is NP-hard to solve optimally.
 
 The model plays the codebreaker and receives feedback after each guess until it either solves the code or runs out of attempts. The game difficulty is configurable by increasing the code length and symbol set size.
 
-- Code length: default 4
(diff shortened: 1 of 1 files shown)


xlr8harder:
I realized this should have just been published via the hub, apologies!  Closing PR.

PR #471, opened 2025-10-14
Updated MultiTurnEnv to use the state during rollout

reachv:
## Description
Update the MultiTurnEnv to use the state during rollout. This should allow people to modify the prompt as needed during the rollout. 

Specifically for my use case, I need to update the prompt during setup_state.

@willccbb 
Does it make sense to move rollout, completions, and anything else I'm missing into state? 

## Type of Change

- [ ] Documentation update
- [ ] Test improvement

## Testing

## Checklist

### DIFF
--- a/tests/test_environment.py
+++ b/tests/test_environment.py
@@ -33,12 +33,12 @@ class SimpleEnvironment(Environment):
             completion = [
                 {"role": "assistant", "content": response.choices[0].message.content}
             ]
-            state = {"responses": [response]}
+            state = {"completion": completion, "responses": [response]}
         else:
             assert isinstance(response, Completion)
             completion = response.choices[0].text
-            state = {}
-        return completion, state
+            state = {"completion": completion}
+        return state
 
 
 class TestEnvironmentBase:
--- a/tests/test_environment_extra.py
+++ b/tests/test_environment_extra.py
@@ -46,15 +46,17 @@ class DummyEnvironment(Environment):
                 {"role": "assistant", "content": response.choices[0].message.content}
             ]
             state = {
+                "completion": completion,
                 "responses": [response],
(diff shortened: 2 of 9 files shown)


willccbb:
This should actually already be supported! The `state['prompt']` object is a pointer to the `prompt` list, you can update in-place directly + it'll be reflected in the `prompt` object. 

See tau2-bench for an example where we update `state['prompt']` to set the prompt in `setup_state` instead of the dataset:

https://app.primeintellect.ai/dashboard/environments/will/tau2-bench

Outside of `setup_state` it's not recommended to do this in the middle of the rollout currently (unless you're fine with your environment being "eval-only"). We enforce that `MultiTurnEnv` is only allowed to extend the context via returning new messages, not rewrite old ones, as some trainers expect rollouts as continuous token sequences rather than a list of (prompt, completion) pairs across multiple turns. I'm in the middle of a trainer refactor which will allow a lot more flexibility in supporting prompt rewriting during rollouts, coming soon :) 

In general I do agree that moving more things into `state` is probably ideal for clarity; gonna leave this open and might group it with a few other changes in the same spirit.

reachv:
I combined completion, state to return a state variable instead. I also wrote some local variables out in-favor of state.

willccbb:
"prompt" is only allowed to be a str if you're doing a completion (non-chat) env, in which case you can append to it via env_response

we have a big refactor in progress which addresses a lot of interrelated issues re: data flow inside rollouts, and will make sure we have good support for modifying prompts in setup_state

for MultiTurnEnv we're going to stick to the tuple approach -- changing would be quite breaking for many envs that have been built around the current API, and this is important for how we enforce concat-only behavior. 

we can be more flexible for API changes that don't touch user-facing methods directly, but env_response + is_completed signatures need to stay as-is, sorry.

PR #549, opened 2025-11-09
Support branching rollouts via trajectories; refactor state handling

willccbb:
## Description

Fairly major refactor to support branching rollouts by storing trajectory-level information. Largely deprecates token-slicing and retokenization during sequence extension in favor of horizontal rollout microbatches where ids/logprobs from vLLM responses are the "ground truth".

Also:
- standardizes run_group + score_group as workhorse methods during generation
- support for step-level rewards + advantages
- state as source of truth for all info during rollouts
- native support for group-level reward functions
- lots of other things simplified and cleaned up along the way

See notes/TRAJECTORIES.md for more discussion.

## Type of Change

- [X] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] Documentation update
- [ ] Test improvement

## Testing

## Checklist


verifiers/rl/trainer/config.py
```diff
@@ -311,7 +311,7 @@ def __post_init__(self):
             "n": 1,
             "presence_penalty": self.presence_penalty,
             "frequency_penalty": self.frequency_penalty,
-            "logprobs": True,
+            "logprobs": 1,
```
mikasenghaas: it is a bool in oai spec i think?

mikasenghaas: i think vllm docs are wrong in making this a int type. pretty sure i tested passing different integer values (0,1,2,3,4) but they would just get evaluated as bool(logprobs), so logprob 0 is not returning any logprobs and any value >0 gives the top token's logprob

verifiers/parsers/parser.py
```diff
@@ -24,19 +24,19 @@ def get_assistant_messages(
         self, completion: list[ChatMessage]
     ) -> list[ChatMessage]:
         """Helper function to extract assistant messages from a completion."""
-        return [msg for msg in completion if msg["role"] == "assistant"]
+        return [msg for msg in completion if msg.get("role") == "assistant"]
```
mikasenghaas: why being more defensive here now? isn't this field guaranteed to be apart of an oai message?

verifiers/rl/trainer/orchestrator.py
```diff
@@ -218,72 +218,66 @@ async def generate_batch(self, batch_id: int) -> Batch:
...
-                for k in range(self.rollouts_per_example)
-                if (prompt_idx + k * prompts_in_batch) < len(rewards)
-            ]
-            if not group_indices:
-                continue
-            group = [rewards[i] for i in group_indices]
-            gmean = sum(group) / float(len(group))
-            for idx, r in zip(group_indices, group):
-                advantages[idx] = r - gmean
+        # process trajectories horizontally - each step becomes a separate training example
```
mikasenghaas, line 231: might be nice to expose this as a function similar to `process_env_results_vllm` so that prime-rl can also use it

verifiers/types.py
```diff
@@ -23,22 +23,101 @@
...
+    completion_logprobs: list[float]
+
+
+class TrajectoryStep(TypedDict):
+    prompt: Messages
+    completion: Messages
+    response: ModelResponse
+    tokens: TrajectoryStepTokens | None
+    reward: float | None
+    advantage: float | None
```
mikasenghaas, line 67: nicee, loving that advantage is in here as well. this means prime-rl should just be able to use this type as is. if we can also add a is_truncated field here then it's exactly like prime-rl's [rollout type](https://github.com/PrimeIntellect-ai/prime-rl/blob/52c5392b3529498ead6d1fc1036af727a62630fc/src/prime_rl/utils/vf.py#L135)

mikasenghaas: can rubrics also compute advantages now btw or is this just to be filled later by the trainer?

verifiers/types.py
```diff
@@ -47,15 +126,10 @@
 class GenerateInputs(BaseModel):
```
mikasenghaas: should this also be a typed dict instead?

verifiers/types.py
```diff
@@ -73,8 +147,8 @@ class GenerateMetadata(BaseModel):
     path_to_save: Path
 
 
-class GenerateOutputs(BaseModel):
-    """Pydantic model for generation outputs."""
+class GenerateOutputs(TypedDict):
```
mikasenghaas, line 155: defo breaking change, but i acc prefer it to avoid any future pydantic<>oai madness

mikasenghaas: if we have a `RolloutInput` type, would be natural for that to return `RolloutOutput` (is this the trajectory type?), and then any group or batch call returns `list[RolloutOutput]` -- i think you are alr doing this for the group, just not for regular `generate`. is this purely for backwards compatibility of `generate` method?

mikasenghaas (commented):
this is great, code is smelling great. will have to play around with it a little bit myself and look at code outside of github diff viewer to have more feedback, but initial vibes are great

verifiers/parsers/parser.py
willccbb: was getting a redline at one point, but not an issue anymore, reverted

verifiers/types.py
willccbb: score_group does by default if not set at step-level; users can do custom advantages within Rubric by setting them in-place at step-level in a group reward function

verifiers/types.py
willccbb: probably yeah

willccbb: actually -- we don't use it, removing (just List[RolloutInput] is cleaner)

verifiers/types.py
willccbb: i think the cleanest way is to just have inputs as lists of rows and outputs as lists of states

the GenerateOutputs type is really just for saving + evals, not needed otherwise, but we want to support intermediate offline saving with metadata

willccbb: do we want prime-rl to deal with GenerateOutputs at all? i was imagining that it's still best for it to directly call run_group and handle submission of groups / tqdm / semaphores / clients, unless you think we want that natively within verifiers as default

willccbb:
added overlong_prompt + is_truncated truncation at step-level for tokens only (if env has max_seq_len set); they're not being applied to rewards yet, still not sure the best spot on the chain by default to do it (e.g. before advantages prob makes sense?)

going to sleep

willccbb:
> Looks good. It will probably get the job done for many use cases, such as truncating think tokens and more generally allow for branching rollouts, so if those are the priority, this should work fine.
> 
> I still think there are a few things that are too tightly coupled and this refactor would increase it even further, as the `Environment` class now knows about groups, which are an orchestration layer concept that greatly constrains which algorithms / advantage calculation schemes we can implement.
> 
> Also, from my understanding, the user would have to define the branching behavior himself and it can't be reused, which seems desirable, given that many people would want to use something like `FullHistory` or `TruncateThinking` or plug in a memory module or let the LLM summarize the conversation every `n` steps and write that as new state, without having to rewrite this for each env.
> 
> I think the core problem in all of this is that the `Environment` class does too much. It handles the environment state transitions, policy I/O and rollout orchestration, but perhaps it would be better to decouple those into
> 
> 1. `EnvKernel` or `Environment` proper, which deals with state transitions only
> 2. `AgentLoop`, which defines how the agent interacts with the environment.
>    
>    * This would also be the place to hook up something like the OpenAI Agent SDK or other external agent systems, where we don't control the rollout.
> 3. `ObservationBuilder`, which is where we define how we present the observations that the environment gives us to the agent (truncate thinking, ...). This would be passed to the `AgentLoop`.
> 4. `Orchestrator`, which is the method that the trainer calls to get a batch. If we need something like _groups_, this would be the place to define what that is, though we could have dependence within and between batches just be a strategy that is coupled to the algorithm. GRPO-style algorithms would then use a generator strategy that forces batches to be split into groups with the same seed.
> 
> Let me know what you guys think. I can definitely work on this, but wouldn't want to start a major refactor if you don't like the general direction :)

Totally hear you -- def wanna be cognizant of the future patterns people will want to use, and ensure that we don't box out anything important prematurely without an easy way to recover it later.

Broadly speaking, the way I think about the design goals of the library is that we want to minimize the average complexity of doing a given thing with LLM RL, roughly weighted by current or expected near-term demand for doing that thing (across the set of people who are / may soon be doing LLM RL). As such, it's pretty important to have first-class abstractions for the most common patterns, where the user doesn't write anything more than the bare minimum. For agentic tool use with LLM judge rewards, users currently have to write 0 new classes, and we don't want to increase that if we don't have to.

This set of common patterns now includes pairwise/groupwise rewards, hence a proper group notion (and sandboxes, hence separate cleanup/teardown); trainers can ignore it in favor of singleton rollouts if desired, but this necessarily will limit some applications. But I also totally agree that the abstractions in the library are not "done", and we will certainly have to change things at the Environment level (e.g. for multi-agent RL, or pluggable agent logic).

My philosophy here is that the higher up the "stack" in the library a user is building, the more they can expect stable behavior over time, whereas lower-level components can change more freely. For example, despite the complexity of this refactor, anyone extending SingleTurnEnv or ToolEnv (incl. StatefulToolEnv, SandboxEnv etc.) or JudgeRubric should basically not notice any difference or have to change their code. 

This is also why MultiTurnEnv is a class on top of Environment, even though basically all envs people have made use MultiTurnEnv. Environment is "allowed" to be messy, and may change drastically as needed. It kind of has to be, as it's a bridge between quickly-moving targets both for agent/environment-level goals and trainer-level features/needs (as well as PI platform things for Environments Hub). If users want to override non-abstractmethods, or use the library for things we haven't mapped out as typical patterns yet, they're more than welcome to, but we can't promise they won't break in future releases. If we need to decouple more things into precise abstractions, we'll do it when we have a clearer sense of exactly what those abstractions should be (which we don't yet). 

The biggest "breaking change to recommended usage" here is the removal of `is_completed` as an abstractmethod in MultiTurnEnv in favor of decorators, but we'd been seeing a lot of people run into issues around the old pattern (it was becoming too load-bearing for both rollout termination and resource cleanup), and the migration is quite straightforward so I don't feel too bad about it.

In addition to GymEnv, I'd also imagine we'll want some things like "AgentEnv" and "MultiAgentEnv" in the near future, and it's not obvious what the abstractions there will be. 

Though I do think that AgentEnv (WorkflowEnv? AgenticWorkflowEnv?) becomes fairly simple after this refactor with a contextmanager approach:
```python
class AgentEnv(vf.Environment):

    async def agent_rollout(self, client, model, agent_input) -> obj:
        agent = await agent_library.from_openai(client, model=model)
        return await agent.do_stuff(agent_input)

    async def rollout(self, input, client, ...):
        state = await self.init_state(input)
        state = await self.setup_state(state)  # customizable
        agent_input = state["agent_input"]
        async with agent_context(client, state):       # traces any oai requests into trajectory
            result = await agent_rollout(client, model, agent_input)
            await self.add_result(state, result)
```

As for more direct control over the context, you could do this currently by overriding MultiTurnEnv.get_prompt_messages:
```python
    @override 
    async def get_prompt_messages(self, state: State) -> Messages:
        if len(state["trajectory"]) == 0:
            return state["prompt"]
        else:
            prev_turn_prompt = state["trajectory"][-1]["prompt"]
            prev_turn_completion = state["trajectory"][-1]["completion"]
            messages = concat_messages([prev_turn_prompt, prev_turn_completion])
            env_response = await self.env_response(messages, state)
            context = self.summarize(concat_messages([messages, env_response]))
            return context
```
This is in the danger zone of things that we aren't promising to not break later, but only because it's not yet clear that this is the "best" way for most users. As the I/O becomes clearer, we'll elevate these sorts of things to first-class abstractions for better reusability (the Tinker cookbook has a notion of a "Renderer" which is used for going from token-ids to printable messages, used for e.g. chat templating; I kinda like the idea of repurposing that kind of abstraction but at a message/context level).

For adding new patterns, it's tremendously helpful to have a "hello world" example which highlights the necessity of a use case, how you'd implement it in a vacuum, and where our current approach falls short in supporting it. Would love to see any examples of the sorts of things you'd want to be able to do, and how you'd imagine they'd look!

TLDR: mostly agree with you, but it's too early to nail it right away, we'll get there soon enough :)

verifiers/envs/environment.py
```diff
@@ -323,281 +391,248 @@ async def get_model_response(
...
+        start_time = state["timing"]["start_time"]
+        end_time = time.time()
+        state["timing"]["generation_ms"] = (end_time - start_time) * 1000
+        state["timing"]["total_ms"] = (end_time - start_time) * 1000
+
+    async def _render_completion(self, state: State):
+        last_prompt = state["trajectory"][-1]["prompt"]
+        last_completion = state["trajectory"][-1]["completion"]
+        full_conversation = concat_messages([last_prompt, last_completion])
+        state["completion"] = full_conversation[len(state["prompt"]) :]
```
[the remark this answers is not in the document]
willccbb: @cursoragent pls fix

PR #609, opened 2025-12-08
Error handling and propagation

mikasenghaas:
## Description

This PR implements exception handling as a first class citizen in verifiers. The main unlocks are:
- Different downstream applications (RL trainer, evals, synthetic data, etc.) can explicitly handle different error classes, e.g. masking/ zero reward/ dropping trajectories with errors
- Improved robustness, especially around tool calling (e.g. not crash on non-JSON decodable outputs)

Main implementation details:
- Define hierarchical error classes. Base error is `vf.Error`, currently with 3 children: `vf.ModelError` (raised if error while getting model response from API), `vf.ToolError` (contains `vf.ToolParseError` and `vf.ToolCallError`) and `vf.InfraError` (contains `vf.SandboxError`), can be expanded in the future as we see need (in my mind, these will be upstreamed somewhat analogous to envs; as we figure out and upstream env patterns, we will also discover and upstream error patterns)
- Environments may define new types of errors, which should subclass the base verifiers error classes. For example, `vf.SandboxEnv` defines `SandboxCreationError` and `SandboxNotReadyError`, and `vf.PythonEnv` defines `PythonWorkerNotReadyError` and `PythonWorkerRequestError`
- Define a new attribute `error: vf.Error | None` on `vf.State` which is used to mark which, if any, error has occurred during the rollout
- Defines a new stop condition `has_error` on the base `Environment` which is check in main multi-turn loop and piggybacks on the completion rendering/ cleanup logic running if triggered
- We try..catch for `vf.Error` in setup and in each turn in `MultiTurnEnv.rollout`, if an error occurs we populate the `error` field in state which will lead to termination of the rollout

Misc. changes:
- Adjusted `get_model_response` function signature to more closely mimic the v0.1.8 pattern--it now only takes `prompt_messages` and `state`
- Updated `ToolEnv`, `StatefulToolEnv` to explicitly catch tool call errors
- Updated `SandboxEnv` and `PythonEnv` to showcase catching infra errors
- Borrows some changes around configurability of PythonEnv from #602 which was used for debugging and the illustration examples
- Update `SandboxEnv` to not wait for creation on every `bash` tool call but only once (borrow pattern from `python` tool in `PythonEnv` which maintains `ready` flag in state)
- Update `math-python` to have more configurability via env args
- Adjusted and expanded tests

### Regular Case

We run a regular `math-python` rollout because it's multi-turn, uses sandboxes and tools

```bash
```

Screenshot 2025-12-09 at 3 47 19 PM

### Expected error

Expected tool errors, such as the model trying to use a package that is not installed in the environment does not raise. Instead the error message is shown to the model (same behavior as before)

```bash
# removing sympy from the list of pre-installed packages
uv run vf-eval math-python -n1 -r1 -v -a '{"pip_install_packages": "numpy scipy"}'
```

Screenshot 2025-12-09 at 3 45 04 PM

### Unexpected error

One example of an unexpected error that should abrupt the rollout and mark as errored (after sufficient retries ofc) are infrastructure errors, such as the sandbox not booting up (in time). We simulate this by setting a too low startup time which won't allow the Python worker to be ready. This will now early exit the rollout and mark the rollout with an error that can be caught as an infra error

```bash
# too low startup timeout to simulate infrastructure failure
uv run vf-eval math-python -n1 -r1 -v -a '{"max_startup_wait_seconds": 5}'
```

Screenshot 2025-12-09 at 3 47 54 PM

## TODOs/ Questions

- [x] Fix tests
- [x] Nail error class anatomy (is the wrapping of root error flexible enough?)
- [x] Should tool errors be defined in`ToolEnv` and sandbox errors in `SandboxEnv` or are they fine in global error
- [ ] Figure out whether we can easily catch an infra error if it is wrapped, e.g. in a tool call error (eg. in prime-rl we would like to handle any infra error in the "error tree")
- [ ] Improved logging for errors in vf-tui and vf-eval

## Type of Change

- [ ] Documentation update
- [ ] Test improvement

## Testing

## Checklist

### DIFF
--- a/verifiers/envs/environment.py
+++ b/verifiers/envs/environment.py
@@ -23,6 +23,7 @@ from typing import (
 from datasets import Dataset
 from openai import AsyncOpenAI, BadRequestError, OpenAI
 
+import verifiers as vf
 from verifiers.parsers.parser import Parser
 from verifiers.rubrics.rubric import Rubric
 from verifiers.types import (
@@ -471,6 +472,10 @@ class Environment(ABC):
         if await condition(state):
             state["is_completed"] = True
             state["stop_condition"] = condition.__name__
+            if state.get("stop_condition") == "has_error":
+                self.logger.error(
+                    f"An error has occurred: {state['error']!r}. Exiting..."
+                )
             return True
         return False
 
@@ -496,13 +501,10 @@ class Environment(ABC):
                 return True
         return False
 
+    @vf.stop(priority=100)  # high priority to always check for errors first
     async def has_error(self, state: State, **kwargs) -> bool:
-        """Checks if an error has occurred in the environment."""
-        has_error = state.get("error") is not None
-        if has_error:
-            print(state["error"])
-            self.logger.error(f"An error has occurred: {state['error']!r}. Exiting...")
-        return has_error
+        """Abrupts rollout early if an error has occurred."""
+        return state.get("error") is not None
 
     async def run_rollout(
         self,
--- a/verifiers/envs/environment.py
+++ b/verifiers/envs/environment.py
@@ -510,11 +510,6 @@ class Environment(ABC):
                 return True
         return False
 
-    @vf.stop(priority=100)  # high priority to always check for errors first
-    async def has_error(self, state: State, **kwargs) -> bool:
-        """Abrupts rollout early if an error has occurred."""
-        return state.get("error") is not None
-
     async def run_rollout(
         self,
         sem: AsyncContextManager,
--- a/verifiers/envs/environment.py
+++ b/verifiers/envs/environment.py
@@ -291,9 +291,10 @@ class Environment(ABC):
 
     async def get_model_response(
         self,
-        client: AsyncOpenAI,
-        model: str,
+        state: State,
         prompt: Messages,
+        client: AsyncOpenAI | None = None,
+        model: str | None = None,
         oai_tools: list[ChatCompletionToolParam] | None = None,
         sampling_args: SamplingArgs | None = None,
         message_type: MessageType | None = None,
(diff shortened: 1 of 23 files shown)


willccbb (approved):

verifiers/envs/environment.py
```diff
@@ -496,6 +510,11 @@ async def is_completed(self, state: State, **kwargs) -> bool:
                 return True
         return False
 
+    @vf.stop(priority=100)  # high priority to always check for errors first
```
willccbb: can we move this to multiturn_env for clarity/consistency?

verifiers/envs/stateful_tool_env.py
```diff
@@ -17,13 +17,11 @@ def __init__(
         self,
         tools: list[Callable] | None = None,
         max_turns: int = 10,
-        error_formatter: Callable[[Exception], str] = lambda e: f"{str(e)}",
```
willccbb, line 20: hmmm what's the need to remove this? would be breaking for some envs, and we want to distinguish between fatal and non-fatal errors. ToolErrors, in general, should probably be displayed to the agent, and user should be able to configure this

mikasenghaas: ah so my reasoning was:
- before: tool raises error -> stateful tool env catches and formats error into fake tool response
- after: tool raises errors -> abrupts rollouts. if tool has an "expected" error (e.g. python call is invalid and want to show stderr to llm) this has to be caught inside the tool and returned as regular tool response

so in essence, there's no need for having a formatter on errors in stateful tool env bc no error will ever be caught and handled in (stateful) tool env. only way i could see keeping it around would be by injecting it into all tools that any child of stateful tool env defines, but that seems a bit clunky?

willccbb: Hmmm don't think I agree with this as default behavior, and we have to be careful about any breaking changes for things that are already in user-configurable space. 

If an error occurs inside a tool function, IMO the *default* behavior should be that this error is shown to the model, as this often is the result of the model passing invalid args (which it should have an opportunity to retry). 

Proposal would be to not remove error_formatter from ToolEnv / StatefulToolEnv, still catch and render by default, and allow user to write their own fatal error logic on a per-tool basis

e.g. if a tool wants to raise a vf.InfraError which terminates the rollout, this is possible, but not default behavior. default should be interpreted as non-fatal + rendered to model

willccbb (approved):

PR #626, opened 2025-12-12
Support for token-in vLLM endpoint

mikasenghaas:
## Description

This PR implements integrates the custom token-in `/v1/chat/completions/tokens` endpoint from PRIME-RL's inference server (introdued in [#1422](https://github.com/PrimeIntellect-ai/prime-rl/pull/1422)) with verifiers so that PRIME-RL can do multi-turn RL without mismatches caused by retokenization.

The main changes are:
- Make `interleaved_rollouts` (and any other extra env kwargs) configurable via `vf-eval`
- If interleaved rollouts is configured, `get_model_response` will correctly set up prompt tokens, sampling args and the client to make a request to the custom endpoint

We decided on the following defaults for reliably building prompt tokens:
- Use vLLM for tokenization (can scale API server capacity with `--api-server-count` to not get bottlenecked by tokenization) 
- Will tokenize the `env_response` in isolation and compute suffix tokens (tokens added in between messages by chat template, but not produced by LLM) _once_ on dummy messages and cache for later usage. This should be safe in 99.9% of the cases.

## Examples

Default behavior is unaffected, e.g. running `math-python` against OAI API

```bash
```

To use the token-in prompt, start a custom vLLM server from PRIME-RL

```bash
uv run inference --model.name Qwen/Qwen3-4B-Instruct-2507 --enable-auto-tool-choice --tool-call-parser hermes --enable-log-requests
```

```bash
uv run vf-eval math-python -n1 -r1 -b http://localhost:8000/v1 -m Qwen/Qwen3-4B-Instruct-2507 -v -x '{"interleaved_rollouts": true}' 
```

## Type of Change

- [ ] Documentation update
- [ ] Test improvement

## Testing

## Checklist

## Additional Notes

---

> [!NOTE]
> Adds interleaved rollouts using a vLLM token-in endpoint with pre-tokenized prompts, and introduces CLI-configurable extra environment kwargs applied at runtime.
> 
> - **Core (Environment)**:
>   - Implement interleaved rollouts path in `get_model_response` using custom `/v1/chat/completions/tokens` with pre-tokenized `prompt_ids` and normalized sampling args.
>   - Add overlong-prompt error handler decorator and refactor arg resolution/sampling normalization.
>   - New setters: `set_kwargs`, `set_interleaved_rollouts` (with warning).
> - **Token utilities**:
>   - New `verifiers/utils/token_utils.py` with `tokenize_vllm`, `get_prompt_ids`, and `prepare_sampling_args_for_token_prompts` (cached suffix handling, overlap logic, tokens client copy).
> - **CLI/Config**:
>   - Add `--extra-env-kwargs` to `vf-eval`; plumb through `EvalConfig.extra_env_kwargs` and apply via `vf_env.set_kwargs` in `run_evaluation`.
> - **EnvGroup**:
>   - Add `set_interleaved_rollouts` to propagate to sub-envs.
> - **Types**:
>   - Make `State.client` and `State.model` required (non-optional).
> - **Tests**:
>   - Update `tests/test_eval_cli.py` to include `extra_env_kwargs` arg and validate sampling args precedence.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 75fa6954ac3d46b0b7bdda4581242633c0a66f58. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


verifiers/utils/eval_utils.py
```diff
@@ -110,6 +110,18 @@ async def run_evaluation(config: EvalConfig) -> GenerateOutputs:
     # load environment
     vf_env = vf.load_environment(env_id=config.env_id, **config.env_args)
 
+    # prepare token prompts
+    if config.use_token_prompts:
+        logger.warning(
+            "Configured to use token prompts. Currently, this is a hand-crated feature for PRIME-RL's vLLM server extension, and is not recommended for general use."
```
willccbb: *crafted

verifiers/envs/multiturn_env.py
```diff
@@ -62,6 +64,62 @@ async def get_prompt_messages(self, state: State) -> Messages:
             env_response = await self.env_response(messages, state)
             return concat_messages([messages, env_response])
 
+    async def get_prompt_ids(self, state: State, client: AsyncOpenAI) -> list[int]:
```
willccbb: can we have this live in response_utils, similar to the other parsing helpers? ideally wanna keep multiturn_env.py file super streamlined

verifiers/envs/multiturn_env.py
```diff
@@ -92,10 +150,22 @@ async def rollout(
         client: AsyncOpenAI,
         model: str,
         sampling_args: SamplingArgs | None = None,
+        use_token_prompts: bool | None = None,
```
willccbb: Can we have this inside of state so that signature doesn't need to change? Ideally the main rollout flow has very little conditional logic, and state is used to encapsulate all the metadata which affects routing

verifiers/envs/multiturn_env.py
```diff
@@ -104,7 +174,13 @@ async def rollout(
         while not await self.is_completed(state):
             try:
                 prompt_messages = await self.get_prompt_messages(state)
-                response = await self.get_model_response(state, prompt_messages)
+                if use_token_prompts:
```
willccbb: if use_token_prompts is inside state, could we just move all of this into get_model_response?

verifiers/envs/multiturn_env.py
```diff
@@ -92,10 +150,22 @@ async def rollout(
         client: AsyncOpenAI,
         model: str,
         sampling_args: SamplingArgs | None = None,
+        use_token_prompts: bool | None = None,
     ) -> State:
         """
         Generate a multi-turn rollout with the environment.
+
+        If use_token_prompts is set, the environment will prepare a token
```
willccbb: Not sure how I feel about this docstring, seems like overkill for the typical reader of the file

willccbb (changes requested):

verifiers/envs/environment.py
```diff
@@ -400,6 +403,96 @@ async def get_model_response(
                     raise vf.OverlongPromptError(e)
             raise vf.ModelError(e)
 
+    async def get_model_response_with_tokens(
```
willccbb: Can we move this all inside get_model_response, and perhaps break things out into clearer helper methods? Want to try and keep both environment.py and multiturn_env.py relatively streamlined + readable

verifiers/envs/environment.py
```diff
@@ -400,6 +403,96 @@ async def get_model_response(
...
+        message_type: MessageType | None = None,
+    ) -> ModelResponse:
+        """
+        Get model response from prompt ids.
+
+        Convenience function for wrapping (chat, completion) API calls.
+        Returns special error messages for context length issues.
+        """
+
+        # TODO: a good chunk of the logic below is duplicated from the get_model_response function, and should be abstracted out.
```
willccbb: Agree, would prefer having this folded inside + broken out into utils before merging

verifiers/envs/environment.py
```diff
@@ -880,6 +973,10 @@ def set_max_seq_len(self, max_seq_len: int | None) -> None:
         """Set the maximum sequence length for this environment."""
         self.max_seq_len = max_seq_len
 
+    def set_use_token_prompts(self, use_token_prompts: bool) -> None:
```
willccbb: If this is the main way that entrypoints (trainers, vf-eval) configure whether token prompts should be used, can we just have init_state set a bool in state and avoid passing around args for it?

verifiers/envs/multiturn_env.py
mikasenghaas: the function changed to `get_prompt_messages_and_ids`  and is now a 1-1 replacement for `get_prompt_messages`. as such, it should live on multi-turn env imo

verifiers/envs/multiturn_env.py
mikasenghaas: i put it inside state, but the signature still needs to change because this function does not take state as an input

verifiers/envs/multiturn_env.py
mikasenghaas: hmm, i think this would get quite messy. i think for now its cleaner to have the default path as is

```py
prompt_messages = await self.get_prompt_messages(state)
response = await self.get_model_response(state, prompt_messages)
```

and then `use_token_prompts=True` is an explicit override of these two messages and nothing else

```py
prompt_messages, prompt_ids = await self.get_prompt_messages_and_ids(state, client)
response = await self.get_model_response_with_tokens(state, prompt_messages, prompt_ids)
```

sure i could merge `get_prompt_messages` and `get_prompt_messages_and_ids` as well as `get_model_response` and `get_model_response_with_tokens` but i think this would be a) harder to understand and maintain and b) somewhat awkward because the signature is diff. in particular, `get_prompt_messages_and_ids` returns two things while `get_prompt_messages` returns one

verifiers/envs/multiturn_env.py
mikasenghaas: removed

verifiers/envs/environment.py
mikasenghaas: done

verifiers/envs/multiturn_env.py
mikasenghaas: okay, acc compromise: i did merge the logic into `get_model_response` because the a) the output type matches and b) there was awkward duplicated logic, so this is now all in one method with inline helpers for better readibility and no code redundancy

i still think we shouldn't merge `get_prompt_messages` and `get_prompt_messages_and_ids`

verifiers/envs/environment.py
mikasenghaas: done

verifiers/envs/environment.py
mikasenghaas: done

mikasenghaas: altho i do wanna add that im a bit confused why max seq len is the only arg that has a setter. i acc think this is quite a useful pattern in prime-rl in general to be able to set a "default arg" (e.g. client/ model) etc. because it would mean i can just set these in the orch main loop and dont have to pass around into helper funcs. but i think this is a discussion for another pr

mikasenghaas: fyi, i put the setter back because it is so so much nicer to setup args in this way rather than when calling .rollout() in PRIME-RL, would acc prefer also setting up client + model in this way

verifiers/envs/environment.py
```diff
@@ -304,108 +312,219 @@ async def get_model_response(
...
+                                        has_audio = True
+                                        break
+                            if has_audio:
+                                break
+                    except Exception:
+                        has_audio = False
+                    if has_audio and "modalities" not in sampling_args:
+                        sampling_args = {
+                            **sampling_args,
+                            "modalities": ["text"],
```
snimu: why only "text" if `has_audio`?

verifiers/envs/environment.py
```diff
@@ -304,108 +312,219 @@ async def get_model_response(
...
+            )
+
+            extra_body = sampling_args.pop("extra_body", {})
+            body = dict(
+                model=model,
+                messages=prompt,
+                tools=oai_tools,
+                tokens=prompt_ids,
+                **sampling_args,
+                **extra_body,
```
snimu: To clarify, is it intentional to quietly overwrite any part of the body if it's contained in the extra_body? e.g. if a different model is in the extra_body than in the model arg, we simply use the model from the extra_body?

verifiers/envs/multiturn_env.py
```diff
@@ -62,6 +66,175 @@ async def get_prompt_messages(self, state: State) -> Messages:
...
+                    # end of message token not found, so we don't need to add any suffix tokens
+                    return []
+
+            def find_largest_overlap(a: list[int], b: list[int]) -> int:
+                """Find the largest overlapping sequence between the end of a and beginning of b."""
+                if not a or not b:
+                    return 0
+
+                max_possible = min(len(a), len(b))
+                for overlap_len in range(1, max_possible + 1):
```
snimu: I think this should be `for overlap_len in reversed(range(1, max_possible + 1):` so it starts with the largest possible overlap; and then `if a_suffix == b_prefix: return overlap_len` below so it returns the right overlap size when there is overlap.

verifiers/envs/multiturn_env.py
```diff
@@ -62,6 +66,175 @@ async def get_prompt_messages(self, state: State) -> Messages:
...
+                    # these are typically chat template specific tokens, such as
+                    # eom tokens, newlines, etc.
+                    suffix_ids = compute_suffix_ids(
+                        dummy_messages_ids, dummy_content_ids[-1]
+                    )
+                    setattr(self, "cached_suffix_ids", suffix_ids)
+                else:
+                    suffix_ids = getattr(self, "cached_suffix_ids")
+                overlap_len = find_largest_overlap(prev_turn_ids, suffix_ids)
+                prev_turn_ids += suffix_ids[-overlap_len:]
```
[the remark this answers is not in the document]
snimu: The cursor comment about this is true, this should be `prev_turn_ids += suffix_ids[overlap_len:]` (without the minus)

snimu (changes requested):

verifiers/envs/multiturn_env.py
snimu: That works better in my tests

image

mikasenghaas: ah ops, good catch

verifiers/envs/environment.py
mikasenghaas: not sure, but i think this was like this before

verifiers/utils/response_utils.py
```diff
@@ -125,3 +134,109 @@ async def parse_response_messages(
...
+    **kwargs,
+) -> list[int]:
+    """Tokenize messages using a local tokenizer."""
+    from transformers import AutoTokenizer
+
+    global _TOKENIZER
+    if _TOKENIZER is None:
+        logger.debug(f"Lazily loading local tokenizer for {model}")
+        _TOKENIZER = AutoTokenizer.from_pretrained(model)
+    tokenizer = _TOKENIZER
```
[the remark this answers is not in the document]
mikasenghaas: this is fine

verifiers/utils/response_utils.py
```diff
@@ -125,3 +134,109 @@ async def parse_response_messages(
...
+    **kwargs,
+) -> list[int]:
+    """Tokenize messages using the vLLM /tokenize API."""
+
+    global _TOKENS_CLIENT
+    if _TOKENS_CLIENT is None:
+        logger.debug("Lazily copying OpenAI client for requests to /tokenize API")
+        url_without_v1 = str(client.base_url).replace("/v1/", "")
+        _TOKENS_CLIENT = client.copy(base_url=url_without_v1)
+    tokens_client = _TOKENS_CLIENT
```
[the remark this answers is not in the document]
mikasenghaas: this is fine for now, we only have one client in this code path

snimu (approved):
Looks really great to me :)

verifiers/envs/multiturn_env.py
```diff
@@ -62,6 +66,175 @@ async def get_prompt_messages(self, state: State) -> Messages:
             env_response = await self.env_response(messages, state)
             return concat_messages([messages, env_response])
 
+    async def get_prompt_messages_and_ids(
```
willccbb: Not happy with adding this function here, this file should always stay incredibly simple + readable + streamlined, strongly prefer that we allow this logic to be more hidden in the background

verifiers/envs/multiturn_env.py
```diff
@@ -16,6 +18,8 @@
 from verifiers.utils.response_utils import (
     parse_response_messages,
     parse_response_tokens,
+    tokenize_local,
```
willccbb: tokenize methods should be abstracted into other utils IMO

verifiers/scripts/eval.py
```diff
@@ -218,6 +218,26 @@ def main():
         default="",
         help="Name of dataset to save to Hugging Face Hub",
     )
+    parser.add_argument(
+        "--use-token-prompts",
+        "-utp",
+        action="store_true",
+        help="Use token prompts. Requires that the inference server supports token-in prompts.",
+    )
+    parser.add_argument(
```
willccbb: Why do we need both tokenize-method and exact-tokenization configurability? Ideally we should pick the "best" one and support it primarily (my preference would be vllm tokenization, and never materializing the local tokenizer inside env)

verifiers/scripts/eval.py
```diff
@@ -218,6 +218,26 @@ def main():
...
+        "--tokenize-method",
+        "-tm",
+        type=str,
+        default=None,
+        help="Whether to use local or remote (vLLM) tokenization. Only applicable if using token prompts. If None, will use vllm tokenization by default.",
```
willccbb: what are the option strings? "local" and "remote"?

verifiers/envs/environment.py
```diff
@@ -423,6 +539,9 @@ async def init_state(
         state["client"] = client
         state["model"] = model
         state["sampling_args"] = sampling_args
+        state["use_token_prompts"] = self.use_token_prompts
```
willccbb: Would rather we just have a global init_extras field which is populated into state more generically

verifiers/envs/environment.py
```diff
@@ -880,6 +1002,20 @@ def set_max_seq_len(self, max_seq_len: int | None) -> None:
...
+        """Set whether to use token prompts for this environment."""
+        self.use_token_prompts = use_token_prompts
+
+    def set_tokenize_method(
+        self, tokenize_method: Literal["local", "vllm"] | None
+    ) -> None:
+        """Set the tokenization method for this environment."""
+        self.tokenize_method = tokenize_method
+
+    def set_exact_tokenization(self, exact_tokenization: bool | None) -> None:
```
willccbb: What does "exact tokenization" mean here? This introduces 3 new binary fields, and I find it unlikely that the whole set product of this space is needed -- what are the dependencies between tokenize method + exact tokenization? Are all 4 pairs something we explicitly want to support? Which are a direct function of higher-upstream configs (e.g branching vs interleaved)?

willccbb:
@mikasenghaas  Another thought -- I'm not sure how much sense it makes to have the tokenizer pool managed at the verifiers layer, seems spiritually similar to inference DP which is auto-managed by vLLM / prime-rl... ideally there is always only a single tokenizer endpoint available, and if replication is needed to manage load, this can be behind the endpoint

verifiers/envs/environment.py
```diff
@@ -326,33 +340,98 @@ async def get_model_response(
...
+                        ):
+                            self.logger.debug("Caught overlong prompt.")
+                            raise vf.OverlongPromptError(e)
+                    # in all other case we raise a generic model error
+                    raise vf.ModelError(e)
+
+            return wrapper
+
+        @handle_overlong_prompt
+        async def get_model_response(
```
willccbb: can we call this something like get_model_response_with_messages to avoid collision?

verifiers/envs/environment.py
```diff
@@ -399,26 +478,72 @@ async def get_model_response(
...
+            )
+
+        client, model, oai_tools, sampling_args, message_type = resolve_optional_args(
+            client, model, oai_tools, sampling_args, message_type
+        )
+        sampling_args = normalize_sampling_args(sampling_args)
+        if self.interleaved_rollouts:
+            sampling_args = prepare_sampling_args_for_token_prompts(sampling_args)
+
+        if self.interleaved_rollouts and len(state["trajectory"]) > 0:
```
willccbb, line 523: niiiice, yeah this is quite clean, logic flow feels pretty easy to follow

verifiers/envs/environment.py
```diff
@@ -91,6 +98,13 @@ def __init__(
         self.env_id = env_id or ""
         self.env_args = env_args or {}
         self.max_seq_len = max_seq_len
+
+        self.interleaved_rollouts = interleaved_rollouts
+        if self.interleaved_rollouts:
+            self.logger.warning(
+                "Environment is configured to use interleaved rollouts. All model responses will after turn 1 will be pre-tokenized before being sent to the model. Currently, this is a hand-crafted feature for PRIME-RL's vLLM server extension, and is not recommended for general use."
```
snimu: "All model responses will after turn 1 will ..." -> double "will"

verifiers/envs/environment.py
```diff
@@ -326,33 +340,98 @@ async def get_model_response(
...
+            oai_tools = oai_tools or state["oai_tools"]
+            sampling_args = cast(
+                SamplingArgs, sampling_args or state["sampling_args"] or {}
+            )
+            message_type = message_type or self.message_type
+            return client, model, oai_tools, sampling_args, message_type
+
+        def normalize_sampling_args(sampling_args: SamplingArgs) -> SamplingArgs:
+            """
+            Norbalize sampling arguments. Mainly does 2 things:
```
snimu: "Norbalize" -> "Normalize"

mikasenghaas: 🥲

verifiers/utils/token_utils.py
```diff
@@ -0,0 +1,187 @@
...
+
+    # the env response is all messages after the previous turn
+    messages = concat_messages([prev_turn_prompt, prev_turn_completion])
+    env_response = prompt_messages[len(messages) :]
+
+    def compute_suffix_ids(lst: list[int], value: int) -> list[int]:
+        """Returns all tokens after the last occurrence of `value` in `lst`, if any."""
+
+        def find_last_index(lst: list[int], value: int) -> int:
+            for i in range(len(lst) - 1, -1, -1):
```
snimu: Tiny nitpick, but I feel like the following would be more readable:

```python
for i, entry in enumerate(reversed(lst)):
    if entry == value:
        return i
```

mikasenghaas: more readable, but worse worst-case complexity. `reversed` is a linear op

mikasenghaas: ig we only call this on dummy messages so should be fine

willccbb (approved):

PR #653, opened 2025-12-21
Implement monitor rubrics

mikasenghaas:
## Description

This PR implements the idea of `MonitorRubric`, a simple extensions of `Rubric` which only allows to register metrics, ie. reward functions with weight `0.0`, with simple helpers to build up metric logging across environment hierarchies. The idea is that monitor rubrics are typically tied to specific environments (e.g. `ToolEnv` has `ToolMonitorRubric` or `SandboxEnv` has `SandboxMonitorRubric`), and such monitor rubric are added in an additive fashion, such that metrics across the whole inheritance tree are respected and logged. For example, in the `math_python` environment, we have the custom `MathRubric` registered as the base rubric to determine the main reawrd. However, we may want to emit metrics for observability. In the `math-python` example, the inheritance tree is `PythonEnv(SandboxEnv(StatefulToolEnv(ToolEnv(MultiTurnEnv(...)))))`. At each level of this hierarchy, different metrics should be emitted, e.g.:
- `MultiTurnEnv`: `num_turns`
- `ToolEnv`: `total_tool_calls`, `{tool_name}_calls` (for each tool)
- `SandboxEnv`: `sandbox_ready_wait_time`, `sandbox_command_execution_time`
- `PythonEnv`: `python_ready_wait_time`

This is implemented with a monitor metric for each env. This means that any future env inherting from envs with a monitor rubric will get the logging of the parent envs for free, making it a very general pattern to help observability during env execution.

## Example

```bash
```

We see the following registered monitors from the respective environments:

Screenshot 2026-01-05 at 4 47 42 PM

## Type of Change

- [ ] Documentation update
- [ ] Test improvement

## Testing

## Checklist

## Additional Notes

---

> [!NOTE]
> Adds lightweight, additive monitoring metrics across the environment hierarchy and unifies rubric composition.
> 
> - Introduces `Environment.add_rubric` to compose rubrics (wraps into `RubricGroup` when needed)
> - Adds monitor rubrics:
>   - `MultiTurnMonitorRubric` → `num_turns`
>   - `ToolMonitorRubric` (in `ToolEnv`) → `total_tool_calls`, `{tool}_calls`
>   - `SandboxMonitorRubric` → `sandbox_ready_wait_time`, `sandbox_command_execution_time` (avg), with state tracking of `ready_wait_time` and per-command timings
>   - `PythonMonitorRubric` → `python_ready_wait_time`, with recorded worker `ready_wait_time`
> - Updates `SandboxEnv`/`PythonEnv` to record and log readiness/command timings; propagates state fields accordingly
> - Simplifies `math_python.load_environment` to return `PythonEnv` directly (no explicit tool rubric wiring)
> - Removes deprecated `ToolRubric` and its exports; updates tests to include `num_turns` in `EnvGroupRubric` metrics
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 573d68fd85b33c666578af7c9bf004f5239e7126. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


verifiers/envs/multiturn_env.py
```diff
@@ -22,10 +23,16 @@
 logger = logging.getLogger(__name__)
 
 
+class MultiTurnMonitorRubric(MonitorRubric):
```
willccbb: small nit -- what's the need for a distinct MultiTurnMonitorRubric class? All env classes currently extend MultiTurnEnv already, and build up the trajectory (even if single-turn)

willccbb: Risks other extensions (e.g. SandboxMonitorRubric) not being multi-turn compatible by default, which seems undesirable.

verifiers/envs/sandbox_env.py
```diff
@@ -89,6 +89,22 @@ def teardown(self, wait: bool = True) -> None:
 
 class SandboxState(TypedDict):
     ready: bool
+    ready_wait_time: float
+    command_execution_times: list[float]
+
+
+class SandboxMonitorRubric(vf.MonitorRubric):
```
willccbb: Is this serving a different role than MultiTurnMonitorRubric? is the idea to have many monitor rubrics which track new things introduced at different hierarchies?

verifiers/envs/environment.py
```diff
@@ -267,6 +267,12 @@ def add_task(example):
             dataset = dataset.map(add_task, **map_kwargs)
         return dataset
 
+    def add_rubric(self, rubric: Rubric) -> None:
+        if self.rubric is None:
+            self.rubric = rubric
+        else:
+            self.rubric = vf.RubricGroup(rubrics=[self.rubric, rubric])
```
willccbb: if self.rubric is already a RubricGroup, maybe we should do self.rubric.add_rubric ?

verifiers/envs/multiturn_env.py
```diff
@@ -22,10 +23,16 @@
...
+    def __init__(self):
+        super().__init__(state_keys=[("trajectory", "num_turns", len)])
```
willccbb: this feels a bit restrictive/unintuitive + not totally seeing the value of the state_keys approach. 

we can already do:
```
async def num_turns(state) -> float:
    return len(state['trajectory'])
```
which is fairly concise + avoids the need for a new pattern

could also be worth adding `add_metric` to Rubric which behaves just like `add_reward_func` but with default weight 0?

verifiers/envs/multiturn_env.py
mikasenghaas: hm, im not sure i understand fully, but the idea is that each env may create an associated monitor rubric, e.g.:
- `MultiTurnEnv` has `MultiTurnMonitorRubric`
- `ToolEnv` has `ToolMonitorRubric`
- `SandboxEnv` has `SandboxMonitorRubric`
- etc.
in this sense, `MultiTurnMonitorRubric` is just the associated monitor rubric class to the base multi turn env and will get registered by default with it. since all envs extend multi turn env, all envs will have this monitor metric registered. i agree that for for single turn envs, it might make sense to remove the turn count as it's a bit silly here

verifiers/envs/sandbox_env.py
mikasenghaas: yea, sandbox monitor rubric would only implement monitor metrics specified to the sandbox env-- i.e. currently `sandbox_ready_wait_time` and `sandbox_command_execution_time`. if every env in the env hierarchy, e.g. for `math-python` we have `PythonEnv(SandboxEnv(StatefulToolEnv(ToolEnv(MultiTurnEnv(...)))))`, register env-specific monitor metrics any env that inherits from any of these gets all the logging for free

mikasenghaas:
> Semi-related -- https://github.com/PrimeIntellect-ai/verifiers/pull/645 adds an option to bypass scoring, which currently would skip all Rubric classes. Do we want to treat MonitorRubric instances as fundamentally separate in the case of a RubricGroup?

ah yea, we probably always wanna run the monitor rubrics, even if scoring is disabled. any ideas on what would be a good way of distinguishing those?

verifiers/envs/multiturn_env.py
mikasenghaas: ah yea, agree -- this is cleaner. i see you alr added the `add_metric` method, gonna use that one.

verifiers/envs/tool_env.py
```diff
@@ -28,6 +75,8 @@ def __init__(
         }
         super().__init__(oai_tools=self.oai_tools, max_turns=max_turns, **kwargs)
 
+        self.add_rubric(ToolMonitorRubric(tools=self.tools))
```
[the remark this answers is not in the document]
mikasenghaas: ah, yea this is fair but i think this wasnt supported previously with the `ToolRubric`? adding would be easy to add, removing would be a bit harder bc we would have to remove a reward func from a rubric which we currently don't have support for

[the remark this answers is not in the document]
willccbb: ah we should probably fix this (can be separate PR), shouldn't be too hard

previously we weren't using tool_rubric that heavily, and it already forced you to pass the list of tools, so it wasn't too big of a problem to just have the added tool in that list as well

PR #739, opened 2026-01-16
optional DatasetBuilder pattern

willccbb:
## Description
Allows passing a Callable[[], Dataset] in place of Dataset for `dataset` / `eval_dataset`, lazy-loads when `get_dataset` / `get_eval_dataset` is first called.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Test improvement

## Testing

- [ ] All existing tests pass when running `uv run pytest` locally.
- [ ] New tests have been added to cover the changes

## Checklist
- [ ] My code follows the style guidelines of this project as outlined in [AGENTS.md](https://github.com/PrimeIntellect-ai/verifiers/blob/main/AGENTS.md)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published

## Additional Notes

---

> [!NOTE]
> Adds optional lazy dataset loading to environments, enabling deferred construction of `Dataset`s via a `DatasetBuilder` callable.
> 
> - Core: `Environment` now accepts `Dataset | DatasetBuilder` for `dataset`/`eval_dataset`, with new `build_dataset()`/`build_eval_dataset()` and formatting via `_format_dataset_source`; `get_*_dataset` triggers builds; preserves eager build for raw `Dataset`s. Stores `map_kwargs` and validates `message_type` earlier.
> - Group/Trainer: `EnvGroup` concatenation now builds sub-env datasets via `env.build_dataset()`/`build_eval_dataset()`; RL `orchestrator` filters with `env.get_dataset()` instead of accessing `env.dataset` directly.
> - Types/Exports: Add `vf.DatasetBuilder` type alias and export from `verifiers.__init__`.
> - Docs: Add "Lazy Loading with DatasetBuilder" section to `docs/environments.md` and `environments/AGENTS.md`; expand `RLMEnv` details.
> - Example env: Refactor `environments/alphabet_sort` to use `get_dataset_builder`, extract helpers, and return builder-backed `SortingEnv`; bump version to `0.1.11`.
> - Misc: Minor typing/casts in `data_utils.py`; small eval input simplification.
> 
[...]

PR #734, opened 2026-01-15
multi-env evals config

mikasenghaas:
## Description

This PR implements evaluating multiple environments in parallel via `vf-eval`. For more details check the updated [docs](https://github.com/PrimeIntellect-ai/verifiers/blob/3c361e5ca1f24cca98d8f53ac5800be624986068/docs/evaluation.md).

This PR is mainly concerned with the config system. Cosmetic updates will be shipped separately, e.g see #735 

## Examples

By default, we still evaluate a single env with no changes to the interface

```bash
uv run vf-eval gsm8k -n5 -r3
```

To configure multi-environment training, specify a comma-separated list of env ids

```bash
uv run vf-eval gsm8k,alphabet-sort -n5 -r3
```

_Note, that all environments use their default configuration. Since CLI arguments apply to _all_ enviroments one can only change values for all environments at the same time. To have more fine-grained configurability, check below._

To configure multi-environment training with (potentially) different arguments for each specify a path to a TOML config file

```bash
uv run vf-eval configs/evals/debug.toml -n5 -r3
```

```toml
# configs/local/vf-eval/debug.toml
[[env]]
id = "gsm8k"
num_examples = 1
rollouts_per_example = 1
```

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Test improvement

## Testing

- [ ] All existing tests pass when running `uv run pytest` locally.
- [x] New tests have been added to cover the changes

## Checklist
- [ ] My code follows the style guidelines of this project as outlined in [AGENTS.md](https://github.com/PrimeIntellect-ai/verifiers/blob/main/AGENTS.md)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published

## Additional Notes

---

> [!NOTE]
> Enables running multiple environments in one invocation with config-driven control, plus install utilities and extensive tests/docs.
> 
> - **Multi-env evals**: New `[[eval]]` TOML configs (`configs/eval/*.toml`) parsed by `load_toml_config()` with validation and precedence; CLI positional becomes `env_id_or_config` (single env or `.toml` path)
[...]

PR #744, opened 2026-01-19
env server/client

mikasenghaas:
## Description

> WIP. Based on #734 and #739.

This PR introduces the `EnvClient` and `EnvServer` which are a drop-in replacement for executing environments in a separate process (pool). This is especially useful for multi-env training (e.g. in `prime-rl`) and evals (e.g. via `vf-eval` or in online evals during training). A couple of notes on design decisions:
- The `EnvClient` mirrors the in-process `Environment` for the most common public facing methods, such as `run_rollout`, `run_group`, `generate` and `evaluate`
- The client/server pattern is protocol-agnostic, i.e. the client may communicate with the server via any protocol (gRPC/ZMQ/HTTP/...). _For now, only ZMQ is implemented_.

## Example

The env server pattern is integrated into `vf-eval` to sidecar an env server

```bash
uv run vf-eval gsm8k -n5 -r3 --use-env-server
```

### EnvServer

A `EnvServer` is initialized like a regular environment with an `env_id` and `env_args`

```python
env_server = ZMQEnvServer(
    env_id=args.env_id,
    env_args=args.env_args,
    address=address
)

try:
    await server.run()
finally:
    await server.close()
```

### EnvClient

A `EnvClient` communicates with a env server over the configured `address`

```python
env = ZMQEnvClient(address=address)

await env.run_rollout(...) # same as Environment.run_rollout
await env.run_group(...) # same as Environment.run_group
await env.evaluate(...) # same as Environment.evaluate
```

### Sidecar Pattern

To sidecar an env server (e.g. from `vf-eval`) simply wrap the `run_server` class method in a `Process` and connect the client to the same address

```python
env_server = Process(
    target=ZMQEnvServer.run_server,
    args=(config.env_id, config.env_args),
    kwargs=dict(address=address)
)
env_server.start()
env = ZMQEnvClient(address=address)

try:
   results = await env.evaluate(...)
finally:
  env_worker.terminate()
  env_worker.join(timeout=5)
  if env_worker.is_alive():
      env_worker.kill()
      env_worker.join()
```

## Breaking

- Pass `client_config: ClientConfig` instead of `client: AsyncOpenAI` to public-facing methods. This is because clients are not serializable, so there is no way to mirror the `Environment` API otherwise
- Can only limit concurrency for an entire rollout for now because the `AsyncContextManagers` are not serializable. We can prob define an env-level gen+score concurrency limits that is enforced across all calls but it's still breakng in the user for `run_rollout` and `run_group`

## TODOs

- [x] Typed inputs/outputs (this required lazily creating clients and removing them from the public-facing API)
- [x] Integrate with multi-env evals
- [x] Logging
- [x] Graceful server termination
- [ ] Unit tests
- [ ] Multi-process environment execution

## Type of Change

- [ ] Documentation update
- [ ] Test improvement

## Testing

## Checklist

### DIFF
--- a/verifiers/envs/environment.py
+++ b/verifiers/envs/environment.py
@@ -1,6 +1,7 @@
 import asyncio
 import atexit
 import functools
+import hashlib
 import inspect
 import json
 import logging
@@ -14,7 +15,6 @@ from datetime import datetime
 from pathlib import Path
 from typing import (
     TYPE_CHECKING,
-    AsyncContextManager,
     Awaitable,
     Callable,
     List,
@@ -25,7 +25,7 @@ from typing import (
 )
 
 from datasets import Dataset
-from openai import AsyncOpenAI, BadRequestError, OpenAI
+from openai import AsyncOpenAI, BadRequestError
 from openai.types.chat import ChatCompletion
 
 import verifiers as vf
@@ -34,6 +34,7 @@ from verifiers.rubrics.rubric import Rubric
 from verifiers.types import (
     ChatCompletionToolParam,
     ChatMessage,
+    ClientConfig,
     DatasetBuilder,
     GenerateMetadata,
     GenerateOutputs,
@@ -46,6 +47,7 @@ from verifiers.types import (
     State,
 )
 from verifiers.utils.async_utils import maybe_semaphore
+from verifiers.utils.client_utils import setup_client
 from verifiers.utils.error_utils import ErrorChain
 from verifiers.utils.eval_utils import make_dataset, save_rollout_results
 from verifiers.utils.message_utils import (
@@ -162,6 +164,8 @@ class Environment(ABC):
         self._stop_conditions: list[StopCondition] = []
         self._cleanup_handlers: list[RolloutCleanup] = []
         self._teardown_handlers: list[EnvironmentTeardown] = []
+
+        self._clients: dict[str, AsyncOpenAI] = {}
 
         self.__post_init__()
 
@@ -338,6 +342,16 @@ class Environment(ABC):
             )
         else:
             return self._format_completion_dataset(dataset, map_kwargs=self._map_kwargs)
+
+    def _get_client(self, client_config: ClientConfig) -> AsyncOpenAI:
+        config_hash = hashlib.sha256(
+            client_config.model_dump_json().encode()
+        ).hexdigest()
+        client = self._clients.get(config_hash)
+        if client is None:
+            client = setup_client(client_config)
+            self._clients[config_hash] = client
+        return client
 
     def _build_dataset(self) -> Dataset | None:
         """Build and cache the training dataset from source if needed."""
@@ -627,7 +641,7 @@ class Environment(ABC):
     async def init_state(
         self,
         input: RolloutInput,
-        client: AsyncOpenAI,
+        client_config: ClientConfig,
         model: str,
         sampling_args: SamplingArgs | None = None,
     ) -> State:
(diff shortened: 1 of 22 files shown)


verifiers/envs/environment.py
```diff
@@ -731,25 +745,21 @@ async def is_completed(self, state: State, **kwargs) -> bool:
     async def run_rollout(
         self,
         input: RolloutInput,
-        client: AsyncOpenAI,
+        client_config: ClientConfig,
```
samsja, line 748: wondering if we shouldn't have some kind backward compatbility and still accept client as input but failed in the case of runtime ?

verifiers/workers/client/__init__.py
```diff
@@ -0,0 +1,3 @@
+from verifiers.workers.client.zmq_env_client import ZMQEnvClient
```
samsja, line 1: wondering if client name is not missleading with open ai client. Do we expect user to instantiate a client or should be handle underthehood by a the load_enviornment

verifiers/workers/client/env_client.py
```diff
@@ -0,0 +1,60 @@
...
+from verifiers.types import (
+    ClientConfig,
+    GenerateOutputs,
+    RolloutInput,
+    SamplingArgs,
+    State,
+)
+
+
+class EnvClient(ABC):
```
samsja, line 13: I would use Protocol here over ABC unless we have a lot of shared code. But not deal breaker can also stick with abc

verifiers/envs/environment.py
mikasenghaas: yea, we could support this but it would always come at the cost of the `EnvClient` not being a drop-in replacement for an `Environment` because we cannot mirror the API with `client` as it's not serializable

verifiers/workers/client/__init__.py
mikasenghaas: we would expect users to instantiate a client/server. imo naming is pretty clear here, it's called exactly what it is: a client to interface with an environment

verifiers/workers/client/env_client.py
mikasenghaas: i will check if possible. i think this might work for the client, but unlikely for the server

verifiers/workers/client/__init__.py
mikasenghaas: env servers wrap `load_environment` (and hence `Environment`), clients are used to interface with those env servers

verifiers/envs/environment.py
samsja: fair

verifiers/workers/client/__init__.py
samsja: okay sg

verifiers/envs/environment.py
willccbb: would prefer at least initially not breaking the rollout API -- some integrations (e.g. Tinker) use this directly, we'll need to do some other changes here around non-OpenAI client types like Anthropic, which can get subsumed by a generic Client / ClientConfig type eventually.  

It's nice for users to be able to play around with envs in scripts/notebooks where the `rollout` method can be used directly, and IMO we should still support this with a generic OpenAI client.

My preference would be to handle it the same way we're doing DatasetBuilder, where we have a union type + keep the old var name back-compatibility, and check the type where relevant. We can allow certain code paths (prime-rl orchestrator, vf-eval) to fail if a generic client is used.

samsja: yeah I was also thinking about Union type here so that we can do both, we can also introduce an new method that only take client config and let this one for backward compatiblity

verifiers/workers/client/__init__.py
willccbb: which users here? IMO the spawning of clients/servers should always be automated by entrypoints like vf-eval / orchestrator. both can pull in info from configs, create their own clients, and spawn + connect to as many servers as needed

mikasenghaas: ah yea, im considering myself a user of verifiers when developing `prime-rl` haha. i agree, the user who is only running commands is blind to this and should never have to spawn an env server themselves

PR #788, opened 2026-01-26
anthropic client + interleaved thinking support 

mikasenghaas:
## Description

> Based on #774. Do not merge before.

This PR adds support for Anthropic clients and interleaved thinking 
- Accepts `AsyncOpenAI | AsyncAnthropic` in public-facing environment methods
- Internally we wrap those native clients with `vf.Client` which maps between a normalized message, response, and tool types and the client-native types
- New environment setter `Environment.set_interleaved_thinking` which passes the thinking flag to the client (e.g. OAI clients do a best-effort parsing of reasoning content)

### Tested Features

- [x] Single-turn + OAI completions
```bash
uv run vf-eval continuation-quality -n1 -r1 -d -v 
```
- [ ] Single-turn + Anthropic completions (didn't test bc didn't find a Anthropic model with completions compatiblity)
```bash
uv run vf-eval continuation-quality -n1 -r1 -d -v -m haiku
```
- [x] Single-turn + OAI chat completions
```bash
uv run vf-eval gsm8k -n1 -r1 -d -v 
```
- [x] Single-turn + Anthropic completions
```bash
uv run vf-eval gsm8k -n1 -r1 -d -v -m haiku
```
- [x] Multi-turn + OAI chat completions
```bash
uv run vf-eval alphabet-sort -n1 -r1 -d -v
```
- [x] Multi-turn + Anthropic messages
```bash
uv run vf-eval alphabet-sort -n1 -r1 -d -v -m haiku
```
- [x] Multi-turn + Tool + OAI chat completion
```bash
uv run vf-eval wiki-search -n1 -r1 -d -v
```
- [x] Multi-turn + Tool + Anthropic responses
```bash
uv run vf-eval wiki-search -n1 -r1 -d -v -m haiku
```
- [x] Multi-turn + OAI chat completions tokens (PRIME-RL vLLM server)
```bash
uv run vf-eval alphabet-sort -n1 -r1 -v -d -m Qwen/Qwen3-4B-Instruct-2507 -b http://localhost:8000/v1
```
- [x] Multi-turn + OAI chat completions tokens (PRIME-RL vLLM server)
```bash
uv run vf-eval alphabet-sort -n1 -r1 -v -d -m Qwen/Qwen3-4B-Instruct-2507 -b http://localhost:8000/v1 -x '{"interleaved_rollouts": true}'
```
- [x] Multi-turn + Tool + Interleaved Thinking + OAI chat completions
```bash
[...]

PR #860, opened 2026-02-08
abstract client + support for interleaved thinking

eligotts:
## Description
based off of pr [788](https://github.com/PrimeIntellect-ai/verifiers/pull/788) which added provider agnostic types for message/client/tool/response flow. 

General questions:
- right now both client implementations are rather large to handle the mixed types (new custom types and legacy oai types). How much backwards compatibility should these support?
- theres also a significant amount of backwards compat bloat in environments.py due to oai_tools vs tool_defs- how do we want to handle tools and what format they should be in?
- in multiturn_env.py, we now need to add an explicit check for 'completion' type in get_prompt_messages and render_completion  because parse_response_message is now chat-shaped. this is kind of ugly i think, so wondering what best route forward is here to support completion type environments
- just generally how this migration should handle mixed message types, how much support we should maintain for oai based types

**Other notes**
1. **Completion warning is expected**
- Completion envs intentionally store `prompt/completion` as `str`. (like continuation quality)
- `RolloutOutput` still types these as messages, so Pydantic warns.
- Behavior is correct; follow-up is widen type to `Messages | str` (should we do this?).

2. **`tool_defs` is canonical**
- Internal runtime now uses only `tool_defs` (provider-agnostic `Tool`).
- Legacy `oai_tools` is accepted only at boundaries, normalized once at init.
- This preserves old env inputs without dual-state bugs.

3. **Legacy aliases kept temporarily**
- `ChatMessage`, `ChatMessages`, `ModelResponse` remain as transition shims.
- Core runtime is migrated, but some experimental/integration modules still reference aliases.
- We’ll remove aliases after those paths finish migration.

4. **Explicit completion path in `MultiTurnEnv`**
- Old helpers previously hid chat/completion branching.
- After moving to unified response parsing, explicit completion logic is required.
- Added completion-specific assembly to keep multi-turn completion envs correct.

**Also fixed**
- Endpoint normalization now preserves `client_type`, so `-e configs/endpoints.py -m haiku` resolves to Anthropic correctly.

## Client Interface

`vf.Client` is the adapter layer that wraps a native SDK client and standardizes everything into a `vf.Response`.

Each client implementation defines four core methods:

1. `to_native_prompt`  
   Convert `vf.Messages` into provider-native prompt format.
2. `get_native_response`  
   Execute the provider-native API call.
3. `raise_from_native_response` (optional)  
   Map/raise provider-specific errors (for example overlong prompt behavior).
4. `from_native_response`  
   Convert provider-native output into unified `vf.Response`.

We intentionally moved to custom unified types (instead of continuing to normalize to OpenAI-only types) because some provider features do not map cleanly to OAI schemas. The current type system is provider-agnostic and supports multimodal/reasoning/tool patterns across clients.

### Implemented Clients

- `OAICompletionsClient`
- `OAIChatCompletionsClient`
- `OAIChatCompletionsTokenClient`
- `AnthropicMessagesClient`

(Anthropic raw completions are not supported in the current implementation, I think they deprecated)

This architecture is meant to be extensible to additional providers/API surfaces (including future OAI Responses-style adapters).

---

## Backward Compatibility

- Public environment APIs remain backward compatible.
- `tool_defs` is now canonical internally; legacy `oai_tools` is accepted at boundaries and normalized at init.
- We keep temporary legacy aliases (`ChatMessage`, `ChatMessages`, `ModelResponse`) as migration shims for non-migrated experimental/integration paths.
- `CustomBaseModel` keeps dict-like access to avoid breaking existing env code that still uses mapping-style message access.
- Completion-mode environments intentionally keep raw string prompt/completion behavior for compatibility (known non-fatal Pydantic serializer warning; follow-up planned to widen output typing).

---

### Tested

- [x] `uv run vf-eval continuation-quality -n1 -r1 -d -v`
- [x] `uv run vf-eval gsm8k -n1 -r1 -d -v`
- [x] `uv run vf-eval gsm8k -n1 -r1 -d -v -m haiku -e configs/endpoints.py`
- [x] `uv run vf-eval alphabet-sort -n1 -r1 -d -v -m haiku -e configs/endpoints.py`
- [x] `uv run vf-eval wiki-search -n1 -r1 -d -v -e configs/endpoints.py`
- [x] `uv run vf-eval math-python -n1 -r1 -v -d -m deepseek-reasoner -e configs/endpoints.py`

### Not Tested Yet

- [ ] Multi-turn + OAI chat completions tokens (PRIME-RL vLLM server)  
  `uv run vf-eval alphabet-sort -n1 -r1 -v -d -m Qwen/Qwen3-4B-Instruct-2507 -b http://localhost:8000/v1`
- [ ] Multi-turn + OAI chat completions tokens + interleaved rollouts (PRIME-RL vLLM server)  
  `uv run vf-eval alphabet-sort -n1 -r1 -v -d -m Qwen/Qwen3-4B-Instruct-2507 -b http://localhost:8000/v1 -x '{"interleaved_rollouts": true}'`

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Test improvement

## Testing

## Checklist

## Additional Notes

---

> [!NOTE]
> **High Risk**
> Touches core request/response plumbing (client selection, tool handling, and rollout state shape) and introduces a new provider adapter, so regressions could affect all environments and model backends despite added test coverage.
> 
> **Overview**
> This PR replaces direct SDK calls inside `Environment.get_model_response` with a provider-agnostic `vf.Client` adapter that converts `Messages`/`Tool` definitions to native provider requests and normalizes outputs into a unified `vf.Response` (including usage, tool calls, and optional reasoning content). It adds first-class Anthropic support via `AnthropicMessagesClient`, plus an OpenAI token-capable client (`OAIChatCompletionsTokenClient`) to support interleaved rollouts/thinking.
> 
> Tools are migrated from legacy `oai_tools` (OpenAI schema) to canonical `tool_defs` (`vf.Tool`), with strict validation and boundary normalization (including intercepted agent tool schemas and MCP tool conversion). Prompt rendering in OpenEnv environments now normalizes raw messages into typed `Messages`, and completion-mode env/tests are updated to treat stored prompts/completions as message lists.
> 
> Endpoint configuration is expanded and normalized to preserve an explicit provider `type`/`api_client_type` (e.g., Anthropic + DeepSeek variants), `pyproject.toml` adds the `anthropic` dependency, and extensive new tests cover client auth/overlong-prompt behavior, multimodal content parts, interception serialization, and endpoint client-type resolution.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 476fb12246de5d1516fed5f248de430014a5acb3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


verifiers/envs/experimental/rlm_env.py
```diff
@@ -62,19 +61,18 @@
...
 from verifiers.utils.data_utils import extract_boxed_answer
 from verifiers.utils.message_utils import concat_messages
 from verifiers.utils.response_utils import (
-    parse_is_truncated,
-    parse_response_messages,
+    parse_response_message,
     parse_response_tokens,
 )
-from verifiers.utils.tool_utils import convert_func_to_oai_tool
+from verifiers.utils.tool_utils import convert_func_to_tool
```
willccbb: can we have name be convert_func_to_tool_def?

verifiers/types.py
```diff
@@ -11,46 +12,225 @@
...
+from pydantic import BaseModel, ConfigDict, Field
+from pydantic import field_validator
 
-# typing aliases
-ChatMessage = ChatCompletionMessageParam
+# ──────────────────────────────────────────────────────────────────────
+# Client / message type literals
+# ──────────────────────────────────────────────────────────────────────
+ClientType = Literal["openai", "anthropic"]
 MessageType = Literal["chat", "completion"]
```
mikasenghaas, line 31: i know this is still from me lol-- but im not sure whether this distinction makes sense under the new client abstraction. what we have now:

 a client which maps generic message types -> native message  types (could be provider specific, eg. oai chat vs. anthropic message, or "message"-specific, eg. oai completion (str) vs. oai chat (dict))

thus, having the client and message type distinct like this doesn't really makes sense to me anymore. the main difficulty with this imo is that the message type is currently owned by the env (e.g. an env may be defined as a "completion" env) which (currently) implies that the completions api is used to get responses.

under the new client abstractions i would probably argue that this should not be allowed. an env defines rollout interactions in terms of generic types and does not dictate which client is used to get the response

verifiers/types.py
```diff
@@ -11,46 +12,225 @@
...
-ModelResponse = Completion | ChatCompletion | None
 
+
+# ──────────────────────────────────────────────────────────────────────
```
mikasenghaas: nit: can we remove the lines, i dont like em:*(

verifiers/types.py
```diff
@@ -11,46 +12,225 @@
...
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, Mapping):
+            return self.model_dump() == dict(other)
+        return super().__eq__(other)
+
+
+# ──────────────────────────────────────────────────────────────────────
+# Shared message types (provider-agnostic)
+# ──────────────────────────────────────────────────────────────────────
+class TextMessage(CustomBaseModel):
```
mikasenghaas, line 56: ah so we still define our own types? i thought we wanted to keep oai types and extend where necessary for now? if we do this, we should imo keep the naming as-is and just define a bunch of type aliases to oai types so that if we decide to move away from it we dont break imports for people

that said, if @willccbb is fine with new types, i am also fine with it. imo we have to define those sooner or later anyways so might as well break it now. but i would expect quite some edge cases for some time with this that we just have to accept and fix bit by bit

verifiers/types.py
```diff
@@ -11,46 +12,225 @@
...
+
+
+class Response(CustomBaseModel):
+    id: str
+    created: int
+    model: str
+    usage: Usage | None = None
+    message: ResponseMessage  # can call tools
+
+    @property
```
mikasenghaas: i think we do decide to go new types we should go full new types haha

verifiers/types.py
```diff
@@ -74,9 +254,9 @@ class TokenUsage(TypedDict):
 
 
 class TrajectoryStep(TypedDict):
-    prompt: Messages
-    completion: Messages
-    response: ModelResponse
+    prompt: Messages | str
```
mikasenghaas: the consistent thing would be to deprecate those because the conversion from our generic message type to the `str` type e.g. oai completions expects lives in the client now so the core vf logic should not have to deal with this. this would break the (small?) set of completion envs

verifiers/types.py
```diff
@@ -141,24 +321,24 @@ class RolloutOutput(dict):
...
+    oai_tools: list[Tool]
     token_usage: TokenUsage
 
 
 class State(dict):
     INPUT_FIELDS = ["prompt", "answer", "task", "info", "example_id"]
     # rollout inputs
     input: RolloutInput
-    client: AsyncOpenAI
+    client: VfClient
```
mikasenghaas: can we type this as `vf.Client`? (again, im aware this is from me lol just didn't get to it)

mikasenghaas (commented):
nice! i think the main open questions from my side are:
- where and how does client resolution happen? right now, bc we have oai completion, oai chat completion, anthropic completion and anthropic message, we basically choose a client based on "which client type passsed? -> oai or anthropic" and "which message type does the env have? -> str or dict". this fails once we introduce a client that does not support e.g. str prompts
- also i would personally be in favor of having an env dictate the message type. it is now client's concern to map generic inputs -> native inputs and native output -> generic outputs. we could think about black or whitelisting specific clients from the registry for an env (e.g. a tool env cannot use oai completions client bc it doesn't support tools) so then we could show standard errs instead of api errs. but beyond that i dont think the env should care abt this
- still not fully decided on which types to use (custom vs. extend oai). former is more elegant and future proof but latter more backwards compat

also i think we should not release this with the next release. i think there's alr enough things in for ppl to adapt to and if we were to go through with this as-is i would expect a lot of breaking changes (just from our custom type defs). we should prob test the typical user migration we expect on research-envs before releasing

mikasenghaas:
merged in #897

PR #955, opened 2026-02-24
best effort tito

eligotts:
## Description
Best effort tito. instead of assuming extension and looking back at strictly last trajectory step, we walk backward until we find a MESSAGES level prefix hit

tested on both wiki-search and eligottlieb/poker-multiagent, which was giving the loud failure previously with TITO as it does explicit rewriting of history (like context folding)

wiki search no regression:
Screenshot 2026-02-28 at 5 45 25 PM

tests on true branching envs (multi-agent setups):
Screenshot 2026-02-28 at 4 48 29 PM
what this shows is for each agent trajectory (extension per agent turn, but because all stuffed into same trajectory we have 'branching'), we are successfully retrieving the latest trajectory step and doing token in from that point.

## Type of Change

- [ ] Documentation update
- [ ] Test improvement

## Testing

## Checklist

## Additional Notes

---

> [!NOTE]
> **Medium Risk**
> Changes how token-stitching selects prior-turn tokens and adds a new fallback path; mistakes could cause silent behavior changes (more MITO calls) or incorrect token prefixes in production.
> 
> **Overview**
> Makes token-stitching (TITO) *best-effort*: `get_prompt_ids` now scans the trajectory backwards to find the **largest message-level prefix match** (normalizing message structures) instead of assuming the last step matches, and returns `None` when no match exists.
> 
> Updates `get_native_response` to **fall back to the standard `/chat/completions` path** when `get_prompt_ids` returns `None`, avoiding broken `/chat/completions/tokens` requests on history rewrites/context folding. Adds async tests covering largest-prefix selection, no-prefix behavior, and ensuring the correct route is used depending on whether prompt token IDs are available.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6ae6affd4f0e328eed80816c490a29b19193913b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/verifiers/clients/openai_chat_completions_token_client.py
+++ b/verifiers/clients/openai_chat_completions_token_client.py
@@ -1,4 +1,5 @@
-from typing import Optional, cast
+from collections.abc import Mapping
+from typing import Any, Optional, cast
 
 from openai import AsyncOpenAI, BaseModel
 from openai.types.chat import ChatCompletion
@@ -64,6 +65,10 @@ class OpenAIChatCompletionsTokenClient(OpenAIChatCompletionsClient):
                 prompt, model, sampling_args, tools
             )
         prompt_ids = await self.get_prompt_ids(state, prompt, tools)
+        if prompt_ids is None:
+            return await super().get_native_response(
+                prompt, model, sampling_args, tools
+            )
         extra_body = sampling_args.pop("extra_body", {})
         body = dict(
             model=model,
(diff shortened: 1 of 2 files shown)


verifiers/clients/openai_chat_completions_token_client.py
```diff
@@ -85,18 +90,61 @@ async def get_prompt_ids(
...
+        Returns None when no trajectory step has a message-level prefix match with
+        prompt_messages.
         """
-        prev_turn_tokens = state["trajectory"][-1]["tokens"]
-        assert prev_turn_tokens is not None
-        prev_turn_prompt_ids = prev_turn_tokens["prompt_ids"]
-        prev_turn_completion_ids = prev_turn_tokens["completion_ids"]
-        prev_turn_ids = prev_turn_prompt_ids + prev_turn_completion_ids
+
+        def normalize_for_comparison(value: Any) -> Any:
```
mikasenghaas, line 124: should we make this a general message_util? seems useful in other places too? also, vaguely remember we have a similar util to this alr but might be wrong

verifiers/clients/openai_chat_completions_token_client.py
```diff
@@ -85,18 +90,61 @@ async def get_prompt_ids(
...
+            if isinstance(value, Mapping):
+                return {
+                    str(key): normalize_for_comparison(val)
+                    for key, val in value.items()
+                }
+            if isinstance(value, list):
+                return [normalize_for_comparison(item) for item in value]
+            return value
+
+        async def find_largest_prefix_match_tokens() -> list[int] | None:
```
mikasenghaas, line 136: maybe a small docstring here as well

verifiers/clients/openai_chat_completions_token_client.py
mikasenghaas, line 205: i know unrelated to this pr but can i think we might be able to remove the suffix part since we tokenize the `env_response_ids = full_ids[len(prev_turn_ids) :]` and not tokenize env response ids in isolation

mikasenghaas (commented):
yeap, this works!

verifiers/clients/openai_chat_completions_token_client.py
```diff
@@ -106,18 +110,63 @@ async def get_prompt_ids(
...
+                    continue
+                best_prefix_len = prefix_len
+                best_step_tokens = step_tokens
+                if best_prefix_len == len(normalized_prompt_messages):
+                    break
+
+            if best_step_tokens is None:
+                return None
+            return best_step_tokens["prompt_ids"] + best_step_tokens["completion_ids"]
+
```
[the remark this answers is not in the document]
mikasenghaas: this is not valid right, because we expect the trajectory step list to be append only essentially right? so even if there is some anomalies in the message, it would be the same across all steps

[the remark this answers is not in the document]
eligotts: yeah i don't think we need to worry about this. the flow is native response (OAI types) -> custom types (ToolCall, AssistantMessage, etc) which get stored in TrajectoryStep. as long as `get_prompt_messages` doesn't too anything too crazy as is extending from _some_ existing TrajectoryStep (like for RLMs, we have the individual subagent trajectories. but for true compaction, we dont do this extension, so we wouldn't find any message prefix), those prompt messages go from custom types -> native prompt (OAI types), then in this code snippet we are comparing to previous trajectory steps which also get transformed to native types. so everything is going from custom types -> native types through to_native_response so we shouldn't really be missing anything.

verifiers/clients/openai_chat_completions_token_client.py
eligotts: yep looks like some circular logic, might be kinda confusing because env_response_ids now might contain suffix/delimiters of assistant message but functionally will be the same

PR #1027, opened 2026-03-17
remove redundant msg normalization + align `env_response` api

mikasenghaas:
## Description

Reduces calls to `normalize_messages` to minimal set: 
- Once and unconditionally in `init_state` (this is safe, because not on hot path)
- Then for all other call sites (e.g. after `get_prompt_messages` or `env_response`) we use `maybe_normalize_messages` which emits a warning before normalizing the message to nudge users towards using the types directly to avoid potential performance bottlenecks
- Changed the `env_response` return type from `Messages | str` to just `Messages` to be consistent with the rest of the API (e.g. `get_model_response`, `rollout`, etc just operate on `Messages` and `str -> vf.Messages` conversion is an example of (undesired) conversion to custom types) and thus be handled the same way as list of dicts
- Also fixes all our envs to use those patterns

## Type of Change

- [ ] Documentation update
- [ ] Test improvement

## Testing

## Checklist

## Additional Notes

---

> [!NOTE]
> **Medium Risk**
> Tightens the core `MultiTurnEnv`/`Environment.get_model_response` APIs to require `Messages` and introduces conditional normalization, which could break downstream custom envs/providers that still return raw dicts/strings or pass string prompts.
> 
> **Overview**
> Standardizes environment and model prompting to operate on typed `vf.Messages` only: `env_response()` and `get_model_response()` now take/return `Messages` (no more `str`), and built-in envs (`alphabet_sort`, `doublecheck`, `sentence_repeater`, `GymEnv`) were updated to emit `vf.UserMessage` objects instead of raw dicts.
> 
> To cut hot-path Pydantic overhead, `MultiTurnEnv` now uses `maybe_normalize_messages()` (new helper) to only normalize when needed and logs a **warning once** when callers still return raw dicts/strings; supporting `log_once`/`warning_once` utilities were added, and `parse_response_message()` now constructs `AssistantMessage` directly instead of round-tripping through `model_validate`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cd87b2c5ce14ea8b396844fea2220cf9eb78c6ad. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/verifiers/envs/environment.py
+++ b/verifiers/envs/environment.py
@@ -494,7 +494,7 @@ class Environment(ABC):
     async def get_model_response(
         self,
         state: State,
-        prompt: Messages,
+        prompt: Messages | str,
         client: Client | None = None,
         model: str | None = None,
         tool_defs: list[Tool] | None = None,
@@ -540,6 +540,18 @@ class Environment(ABC):
         )
 
         self._get_usage_tracker(state, create_if_missing=True)
+
+        if not isinstance(prompt, list) or not all(
+            isinstance(m, vf.Message) for m in prompt
+        ):
+            self.logger.warning(
+                "get_model_response() received raw dicts/strings instead of "
+                "vf.Message objects. This triggers normalize_messages() on every "
+                "call, which can cause unnecessary Pydantic validation overhaead. "
+                "Return vf.Message types (e.g. vf.UserMessage, vf.AssistantMessage) "
+                "from get_prompt_messages() to avoid this overhead."
+            )
+            prompt = normalize_messages(prompt)
 
         response = await client.get_response(
             prompt=prompt,
--- a/verifiers/envs/environment.py
+++ b/verifiers/envs/environment.py
@@ -494,7 +494,7 @@ class Environment(ABC):
     async def get_model_response(
         self,
         state: State,
-        prompt: Messages | str,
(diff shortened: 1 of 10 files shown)


verifiers/envs/environment.py
```diff
@@ -562,12 +562,10 @@ def resolve_optional_args(
             client, model, tool_defs, sampling_args
         )
 
-        normalized_prompt = normalize_messages(prompt, field_name="prompt")
```
eligotts, line 542: this is the only thing that MIGHT be a bit unsafe, all of our clients assume prompt is already a list of Messages of type `SystemMessage | UserMessage | AssistantMessage | ToolMessage | TextMessage` so if an env overrides `get_prompt_messages` like for any compaction type envs and returns any raw dicts this would break i think. could just do a check and hard fail if we want people to migrate to our stuff

willccbb (approved):
Approving in advance, but let's auto-convert or hard-fail if `get_prompt_messages` is overridden to return dicts. would prefer auto-convert i think, writing dicts is often the more ergonomic way to do it, but fine either way (could surface a helper fn + mention in docstring?).

eligotts (approved):

PR #1067, opened 2026-03-25
feat: composable Task/Agent/Environment architecture

hallerite:
## Summary

Abstracts the **task axis** while reusing the battle-tested `CliAgentEnv` as-is.

### The problem

Each environment (opencode_swe, opencode_lean, etc.) bundles task logic with agent logic. Adding a new task for an existing agent requires writing a full environment from scratch.

### The solution

Three new files in `verifiers/envs/experimental/composable/`:

```python
taskset = R2EGymTaskSet()
harness = opencode_harness(system_prompt="You are a coding agent...")
env = ComposableEnv(taskset=taskset, harness=harness, keep_sandbox_for_scoring=True)
```

### Core concepts

| Concept | What it is |
|---|---|
| `Task` | One fully-bound instance — `task.prompt`, `task.sandbox_spec`, `task.image` |
| `TaskSet` | Collection + behavior for tasks. Override `get_instruction`, `get_rubric` |
| `SandboxTaskSet` | TaskSet with sandbox. Override `get_sandbox_spec`, `setup(state)` |
| `SandboxSpec` | Per-instance sandbox requirements (image, CPU, memory, GPU type, timeout) |
| `Harness` | Agent-side config — install script, run command, declared paths |
| `ComposableEnv` | `CliAgentEnv` subclass, pure polymorphism, zero isinstance checks |

### Key design decisions

- **Uniform `state`-based API**: all TaskSet methods (`setup`, `validate_instance`) take just `state`. Sandbox context lives in state
- **Rubrics own all scoring**: each taskset provides `get_rubric()`. Rubrics run tests / read files / compute rewards in `score_rollout`. Use `keep_sandbox_for_scoring=True` so sandbox stays alive for rubrics. Enables retrying scoring independently of rollouts
- **Zero isinstance branching** in ComposableEnv — `TaskSet` base provides sensible defaults
- **`SandboxSpec`**: per-instance sandbox requirements with `gpu_type` support
- **`validate()`**: verify gold solutions at scale, replaces SolveEnv

### Changes to CliAgentEnv

- `get_sandbox_resources(state)` hook for per-instance resources
- `PROTECTED_ENV_VARS` class attribute
- `gpu_type` passthrough to `CreateSandboxRequest`
- `vm=True` auto-set when `gpu_count > 0`

### Validated E2E (GPT-5, 3 examples each)

| TaskSet | Ours | Main | Match? |
|---|---|---|---|
| Science | 0.333 | 0.333 | Identical |
| CP | 1.000 | 1.000 | Identical |
| SWE (R2E) | 0.000 | 0.667 | Same examples, GPT-5 variance |
| Math | 0.333 | 0.000 | Same examples, GPT-5 variance |
| Lean | 1.000 (GPT-5) | n/a | — |

**Companion PR:** PrimeIntellect-ai/research-environments#221

## Test plan

- [x] `pytest tests/test_composable_env.py` — 15 pass
- [x] `ruff check` + `ty check` — clean
- [x] E2E with real sandboxes + Prime inference API (all 5 envs)
- [x] All reviewer feedback from Mika and Xeophon addressed
- [x] Full param parity with main (verified against production TOML config)

---

> [!NOTE]
> **Medium Risk**
> Introduces a new experimental environment stack and changes `CliAgentEnv` sandbox creation to use a new `get_sandbox_resources()` hook (including `gpu_type`/VM selection), which could affect any existing subclasses relying on the old fixed resource fields or env-var behavior.
> 
> **Overview**
> Adds a new experimental **composable** architecture that splits *task definition* from *agent execution*: `Task`/`TaskSet`/`SandboxTaskSet` + `SandboxSpec` define per-instance instructions, sandbox requirements, env vars, and validation, while `Harness` defines how to install/run an agent; `ComposableEnv` wires them together by uploading instruction/system prompt files, exporting `AGENT_WORKDIR`, and delegating sandbox spec/resources to the task.
> 
> Updates `CliAgentEnv` to support per-rollout sandbox sizing via `get_sandbox_resources(state)` and to pass `gpu_type`/`vm`/custom timeouts into `CreateSandboxRequest`, plus adds protected env-var keys to prevent subclasses from overriding interception-critical variables. Documentation and a new `pytest` suite were added to cover the new APIs and behaviors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7aa6b0d49001e18ee1b1841c81f8dab8c0f4a758. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


verifiers/envs/experimental/composable/composable_env.py
```diff
@@ -0,0 +1,213 @@
...
+
+    Extends ``SandboxMonitorRubric`` so sandbox OOM/timeout metrics are
+    included in the same rubric.
+    """
+
+    def __init__(self, **kwargs: Any):
+        super().__init__(**kwargs)
+        self.add_reward_func(self.spec_reward)
+
+    async def spec_reward(self, state: State) -> float:
```
mikasenghaas: why is this called `spec_reward`?

verifiers/envs/experimental/composable/composable_env.py
```diff
@@ -0,0 +1,213 @@
...
+        if "rubric" not in kwargs:
+            kwargs["rubric"] = ComposableRubric()
+        super().__init__(run_command=run_command, **kwargs)
+
+        self.spec = taskset.spec
+        self.install_script = install_script
+        self.system_prompt_path = system_prompt_path
+        self.test_timeout = test_timeout
+
+        self.add_rubric(CliAgentMonitorRubric())
```
mikasenghaas: this is alr added by `CliAgentEnv` - should prob remove?

verifiers/envs/experimental/composable/task.py
```diff
@@ -0,0 +1,380 @@
...
+    ) -> bool:
+        """Verify this instance is solvable.
+
+        Applies the known-correct solution and checks that evaluation
+        passes.  Domain-specific: SWE applies the gold patch and runs
+        tests, Lean compiles the ground-truth proof, Harbor runs solve.sh.
+        Returns True if the instance is valid, False if broken.
+        Tasks without a gold solution should return True (assume valid).
+        """
+        ...
```
mikasenghaas: for tasks which do not require sandboxes the api seems to have a lot of unnecessary methods and method args

verifiers/envs/experimental/composable/task.py
```diff
@@ -0,0 +1,380 @@
...
+    def get_workdir(self) -> str:
+        return self.spec.get_workdir(self.info)
+
+    def __repr__(self) -> str:
+        return (
+            f"Task(spec={type(self.spec).__name__}, info_keys={list(self.info.keys())})"
+        )
+
+
+# ---------------------------------------------------------------------------
```
mikasenghaas: remove those

verifiers/envs/experimental/composable/task.py
```diff
@@ -0,0 +1,380 @@
...
+        return self.spec.get_prompt(info)
+
+    def get_image(self, info: dict) -> str:
+        return self.spec.get_image(info)
+
+    def get_workdir(self, info: dict) -> str:
+        return self.spec.get_workdir(info)
+
+    def get_env_vars(self) -> dict[str, str]:
+        return self.spec.get_env_vars()
```
mikasenghaas: the fact that we delegate so much from `Task` -> `TaskSpec` might indicate that `Task` should subclass `TaskSpec`?

verifiers/envs/experimental/composable/task.py
```diff
@@ -0,0 +1,380 @@
...
+        """
+        import asyncio
+        import logging
+        import time
+
+        from prime_sandboxes import CreateSandboxRequest
+        from verifiers.utils.threaded_sandbox_client import ThreadedAsyncSandboxClient
+
+        logger = logging.getLogger(__name__)
+        client = ThreadedAsyncSandboxClient(max_workers=concurrency)
```
mikasenghaas: max_workers should not be set to concurrency otherwise you can get bunch of overhead from the thread pool which each owns a client. the current scaling function we use on main is `min(max(1, c // 8), 50)` which i have found to be good

verifiers/envs/experimental/composable/composable_env.py
```diff
@@ -0,0 +1,213 @@
...
+            return self.docker_image
+
+    async def build_env_vars(self, state: State) -> dict[str, str]:
+        """Merge base env vars with Task-provided env vars."""
+        env_vars = await super().build_env_vars(state)
+        env_vars.update(self.spec.get_env_vars())
+        # Set AGENT_WORKDIR from task if available
+        info = state.get("info") or {}
+        try:
+            env_vars.setdefault("AGENT_WORKDIR", self.spec.get_workdir(info))
```
mikasenghaas: is this standard?

verifiers/envs/experimental/composable/composable_env.py
```diff
@@ -0,0 +1,213 @@
...
+        # 1. Task setup (repo checkout, venv links, proof file, etc.)
+        await self.spec.setup(self.sandbox_client, sandbox_id, state)
+
+        # 2. Upload task instruction for the agent to read
+        info = state.get("info") or {}
+        try:
+            prompt = self.spec.get_prompt(info)
+            instruction = ""
+            for msg in prompt:
+                content = (
```
mikasenghaas: can we not assume vf message type?

verifiers/envs/experimental/composable/composable_env.py
```diff
@@ -0,0 +1,213 @@
...
+                    == "user"
+                ):
+                    instruction += str(content)
+            if instruction:
+                await self.sandbox_client.execute_command(
+                    sandbox_id, "mkdir -p /task", timeout=10
+                )
+                await self.sandbox_client.execute_command(
+                    sandbox_id,
+                    f"cat > /task/instruction.md << 'INSTRUCTION_EOF'\n{instruction}\nINSTRUCTION_EOF",
```
mikasenghaas: also this seems like the agent should own?

verifiers/envs/experimental/composable/composable_env.py
```diff
@@ -0,0 +1,213 @@
...
+                    timeout=30,
+                )
+        except Exception as e:
+            self.logger.warning(f"Failed to upload instruction: {e}")
+
+        # 3. Upload system prompt if provided
+        if self.system_prompt_path:
+            await self.sandbox_client.execute_command(
+                sandbox_id, "mkdir -p /opencode", timeout=10
```
mikasenghaas: no opencode here

verifiers/envs/experimental/composable/composable_env.py
```diff
@@ -0,0 +1,213 @@
...
+        if not sandbox_id:
+            return
+
+        # Skip evaluation on infra errors
+        if state.get("error") and isinstance(state["error"], vf.InfraError):
+            state["reward"] = 0.0
+            return
+
+        # Store run_background_job so Task.evaluate() can run tests
+        state["_run_background_job"] = self.run_background_job
```
mikasenghaas: hmm this seems hacky

verifiers/envs/experimental/composable/composable_env.py
```diff
@@ -0,0 +1,213 @@
...
+                state,
+            )
+            if isinstance(reward, dict):
+                state["role_rewards"] = reward
+                state["reward"] = sum(reward.values()) / len(reward) if reward else 0.0
+            else:
+                state["reward"] = float(reward)
+        except Exception as e:
+            self.logger.warning(f"Task evaluation failed: {e}")
+            state["reward"] = 0.0
```
mikasenghaas: we actually move the scoring outside of `post_rollout` and inside the rubric -- rubrics can now inherit sandbox mixin and take ownership of rollout sandboxes and use them for scoring. if we can, would move towards this pattern as well because it allows us to retry just the scoring while keeping the rollout in tact later down the line

mikasenghaas (commented):
very nice! i do like this a lot. i think in terms of api im mainly concerned with the fact that the task api seems very tailored towards tasks that require sandbox and it seems a bit clunky to use if they done. would be nice if we can get this a bit more elegant so that a math qa taskset really is just 10 lines of code or so. otherwise mainly small nits in how it stacks onto the cli agent env

verifiers/envs/experimental/composable/task.py
```diff
@@ -0,0 +1,380 @@
...
+    The TaskSpec provides the shared behavior (image, setup, evaluate).
+    """
+
+    def __init__(self, spec: TaskSpec, dataset: Any, name: str = ""):
+        self.spec = spec
+        self._dataset = dataset
+        self.name = name
+
+    @property
+    def needs_sandbox(self) -> bool:
```
xeophon: i think this should not be a bool, but specs about the sandbox in terms of resources, incl. GPUs. 

and because this is rather clunky atm, i propose a sandboxspec, see https://github.com/PrimeIntellect-ai/verifiers/pull/1021/changes#diff-6f37472f45d2eb27469ba5abd7de7b9c7633adcb604b20aa502e5b849b187c35R17

verifiers/envs/experimental/composable/task.py
xeophon: why do we need taskspec, task and taskset? i think only task + taskset is fine

verifiers/envs/experimental/composable/task.py
```diff
@@ -0,0 +1,380 @@
...
+        sliced = self._dataset.select(range(min(n, len(self._dataset))))
+        return TaskSet(spec=self.spec, dataset=sliced, name=self.name)
+
+    # -- Validation ---------------------------------------------------------
+
+    async def validate_taskset(
+        self,
+        n: int | None = None,
+        concurrency: int = 10,
+        cpu_cores: int = 2,
```
xeophon: we should takes these values from the sandboxspec, see https://github.com/PrimeIntellect-ai/verifiers/pull/1067/changes#r2994430134

verifiers/envs/experimental/composable/composable_env.py
```diff
@@ -0,0 +1,213 @@
...
+    def __init__(self, **kwargs: Any):
+        super().__init__(**kwargs)
+        self.add_reward_func(self.spec_reward)
+
+    async def spec_reward(self, state: State) -> float:
+        """Return the reward computed by the Task during rollout."""
+        return float(state.get("reward") or 0.0)
+
+
+class ComposableEnv(CliAgentEnv):
```
xeophon, line 51: how would you be able to change the tools / disable some? something that opencodeenv allows atm

hallerite: you would do this at the harness level, but the harness would ofc have to support it.

passing tools from task to harness is a bit of scope creep rn, imo

verifiers/envs/experimental/composable/composable_env.py
hallerite: agreed. opened a PR for the sandbox sdk (PrimeIntellect-ai/prime#482) to make this cleaner here

verifiers/envs/experimental/composable/task.py
hallerite: after further consideration, I agree with you. I was thinking of an edge case where tasks in the same taskset need different ways for setup, but perhaps they should just be different tasksets then

verifiers/envs/experimental/composable/task.py
hallerite: agreed. will implement something like this.

verifiers/envs/experimental/composable/task.py
hallerite: you are right. tasks that need a sandbox and those that don't are fundamentally different and I guess it would be fine to differentiate between them on a type level, even without the clean shared interface (`evaluate` needs only state for tasks, but sandbox details for sandboxed tasks) and at the cost of `isinstance` in `ComposableEnv` for now.

verifiers/envs/experimental/composable/composable_env.py
hallerite: added this as future work in the README, to avoid scope creep, but should def tackle this v soon

verifiers/envs/experimental/composable/composable_env.py
hallerite: found an easier way to handle this generally

verifiers/envs/experimental/composable/composable_env.py
hallerite: nope, made it more general.

verifiers/envs/experimental/composable/composable_env.py
```diff
@@ -0,0 +1,237 @@
...
+            return {
+                "cpu_cores": spec.cpu_cores,
+                "memory_gb": spec.memory_gb,
+                "disk_size_gb": spec.disk_size_gb,
+                "gpu_count": spec.gpu_count,
+                "timeout_minutes": spec.timeout_minutes,
+            }
+        return super().get_sandbox_resources(state)
+
+    _PROTECTED_ENV_VARS = frozenset(
```
mikasenghaas: should those live in `CliAgentEnv`?

verifiers/envs/experimental/composable/composable_env.py
```diff
@@ -0,0 +1,237 @@
...
+        if isinstance(self.taskset, SandboxTaskSet):
+            await self.taskset.setup(self.sandbox_client, sandbox_id, state)
+
+        # 2. Upload instruction to harness-declared path
+        # Let programming errors (NotImplementedError, KeyError) propagate —
+        # only catch upload failures.
+        info = state.get("info") or {}
+        instruction = self.taskset.get_instruction(info)
+        if instruction.strip():
+            await self._ensure_parent_dir(sandbox_id, self.harness.instruction_path)
```
mikasenghaas: nit: could make this an arg like `upload_content(ensure_parent_dir: bool = True)` which creates the parent in the same command to save one sandbox roundtrip and to have cleaner code

verifiers/envs/experimental/composable/composable_env.py
```diff
@@ -0,0 +1,237 @@
...
+import verifiers as vf
+from verifiers.envs.experimental.cli_agent_env import CliAgentEnv
+from verifiers.envs.experimental.composable.harness import Harness
+from verifiers.envs.experimental.composable.task import SandboxTaskSet, TaskSet
+from verifiers.types import State
+
+logger = logging.getLogger(__name__)
+
+
+class ComposableRubric(vf.Rubric):
```
mikasenghaas: can we put `task.evaluate()` here?

mikasenghaas: i dont quite like the pattern that we have silently converged to of having rubrics be state readers. we should have a clear distinctino in terms of abstractions between what is solving the task and what is verifying the task. this is important for built-in agent timing traces (e.g. how long does it take to rollout vs. score), retries (e.g. retry just the scoring on transient errors without losing the rollout progress). all of this can ofc be hacked into the `post_rollout` hook but it doesn't generalize across envs at all

verifiers/envs/experimental/composable/composable_env.py
```diff
@@ -0,0 +1,237 @@
...
+            state["reward"] = 0.0
+            return
+
+        try:
+            if isinstance(self.taskset, SandboxTaskSet):
+                reward = await self.taskset.evaluate(
+                    self.sandbox_client, sandbox_id, state, timeout=self.test_timeout
+                )
+            else:
+                reward = await self.taskset.evaluate(state)
```
mikasenghaas: nit: wonder if we should swallow the extra args that sandbox taskset uses into kwargs for the regular taskset to avoid the branching. it's a little bit less explicit in which method takes which args but it would be the more "vf-like" way of doing things imo

verifiers/envs/experimental/composable/task.py
```diff
@@ -0,0 +1,323 @@
...
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Callable
+
+from verifiers.types import Messages, State
+
+
+@dataclass
```
mikasenghaas, line 35: nice! having smth like this is overdue -- should prob live in `SandboxMixin` tho? can keep here for this pr, but would like to wire this dataclass through our entire stack at some ponit

verifiers/envs/experimental/composable/task.py
```diff
@@ -0,0 +1,323 @@
...
+        """Plain text instruction for the agent.
+
+        This is the primary method to override. Used for:
+        - The instruction file written to the sandbox (e.g. /task/instruction.md)
+        - Building Task.prompt (wrapped as a user message)
+        """
+        raise NotImplementedError
+
+    async def evaluate(self, state: State) -> float:
+        return 0.0
```
mikasenghaas: should we raise here as well?

verifiers/envs/experimental/composable/task.py
```diff
@@ -0,0 +1,323 @@
...
+
+    # -- Combinators ---------------------------------------------------------
+
+    def filter(self, predicate: Callable[[dict], bool]) -> TaskSet:
+        return self._clone(self._dataset.filter(predicate))
+
+    def take(self, n: int) -> TaskSet:
+        return self._clone(self._dataset.select(range(min(n, len(self._dataset)))))
+
+    def _clone(self, dataset: Any) -> TaskSet:
```
mikasenghaas: this needed?

pyproject.toml
```diff
@@ -105,6 +105,9 @@ rl = [
 [tool.uv]
 preview = true
 
+[tool.uv.sources]
+prime-sandboxes = { git = "https://github.com/primeintellect-ai/prime.git", subdirectory = "packages/prime-sandboxes", rev = "8b38a92dd1e60f9ee0cb241f56a294bfa955a371" }
```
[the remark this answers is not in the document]
hallerite: @mikasenghaas should we make a release instead?

mikasenghaas (commented):
nice, very close!

verifiers/envs/experimental/composable/task.py
hallerite: credit goes to @xeophon :)

verifiers/envs/experimental/composable/task.py
hallerite: yes

verifiers/envs/experimental/composable/task.py
hallerite: good point, nope.

verifiers/envs/experimental/composable/composable_env.py
hallerite: would def save us a lot of branching. seems good to me.

verifiers/envs/experimental/composable/task.py
```diff
@@ -0,0 +1,323 @@
...
+        assert isinstance(self, SandboxTaskSet)
+
+        async def validate_one(i: int) -> dict:
+            row = ds[i]
+            info = row.get("info") or {}
+            state: State = {"info": info}  # type: ignore[assignment]
+
+            async with sem:
+                spec = self.get_sandbox_spec(info)
+                sb = await client.create(
```
xeophon: when you create a client and it fails, the whole batch will fail, as its not in the try blog

verifiers/envs/experimental/composable/composable_env.py
hallerite: totally agree, but we can improve this in another PR for all of our envs imo

verifiers/envs/experimental/composable/composable_env.py
hallerite: you are right, they should

verifiers/envs/experimental/composable/task.py
```diff
@@ -0,0 +1,323 @@
...
+
+@dataclass
+class SandboxSpec:
+    """Sandbox requirements for a task instance."""
+
+    image: str = "python:3.11-slim"
+    cpu_cores: int = 4
+    memory_gb: int = 4
+    disk_size_gb: int = 10
+    gpu_count: int = 0
```
xeophon, line 43: needs allowed gpu types!

verifiers/envs/experimental/composable/task.py
```diff
@@ -0,0 +1,323 @@
...
+                spec = self.get_sandbox_spec(info)
+                sb = await client.create(
+                    CreateSandboxRequest(
+                        name=f"validate-{i}",
+                        docker_image=spec.image,
+                        cpu_cores=spec.cpu_cores,
+                        memory_gb=spec.memory_gb,
+                        disk_size_gb=spec.disk_size_gb,
+                        gpu_count=spec.gpu_count,
+                        timeout_minutes=timeout_minutes,
```
xeophon: why not use spec.timeout_minutes here?

verifiers/envs/experimental/composable/task.py
```diff
@@ -0,0 +1,323 @@
...
+        ds = self.get_dataset()
+        total = min(n, len(ds)) if n is not None else len(ds)
+        is_sandbox = isinstance(self, SandboxTaskSet)
+
+        if not is_sandbox:
+            # No sandbox needed — just run validate_instance on each
+            results: list[dict] = []
+            for i in range(total):
+                row = ds[i]
+                state: State = {"info": row.get("info") or {}}  # type: ignore[assignment]
```
xeophon: do we only need info as information during rollouts?

verifiers/envs/experimental/composable/composable_env.py
```diff
@@ -0,0 +1,237 @@
...
+            except Exception as e:
+                self.logger.warning(f"Failed to collect agent logs: {e}")
+
+        await super().post_rollout(state)
+
+        if not sandbox_id:
+            return
+
+        if state.get("error") and isinstance(state["error"], vf.InfraError):
+            state["reward"] = 0.0
```
xeophon: - needs to set vf.State["error"]

tests/test_composable_env.py
```diff
@@ -0,0 +1,176 @@
...
+    async def evaluate(self, state):
+        return 1.0 if state.get("info", {}).get("id") == 0 else 0.0
+
+
+def _make_dataset(n=3):
+    from datasets import Dataset
+
+    return Dataset.from_dict(
+        {
+            "question": [f"q{i}" for i in range(n)],
```
xeophon: question is in info field, not top level

verifiers/envs/experimental/composable/composable_env.py
```diff
@@ -0,0 +1,237 @@
...
+        harness: Harness,
+        test_timeout: int = 900,
+        **kwargs: Any,
+    ):
+        kwargs["dataset"] = taskset.get_dataset()
+        if "rubric" not in kwargs:
+            kwargs["rubric"] = ComposableRubric()
+        super().__init__(run_command=harness.run_command, **kwargs)
+
+        self.taskset = taskset
```
xeophon, line 76: env_id = taskset.name

right now, this code:

```
taskset = MyTaskSet(ds, name="mytask")
env = ComposableEnv(taskset=taskset, harness=...)
```

results in `{"task": "default", ...}` instead of `{"task": "mytask", ...}`

verifiers/envs/experimental/composable/composable_env.py
mikasenghaas: we already have this for opencode envs tho, so this would be a regression then

verifiers/envs/experimental/composable/composable_env.py
mikasenghaas: why? it's alr set no?

verifiers/envs/experimental/composable/composable_env.py
hallerite: ah, I see. in that case, let me refactor this.

verifiers/envs/experimental/composable/composable_env.py
hallerite: `env_id` is used for module loading, can't set it from `taskset.name` & state["task"] still gets populated correctly via `_ensure_task` from `env_id`

verifiers/envs/experimental/composable/task.py
hallerite: agreed, might need `answer`, too.

verifiers/envs/experimental/composable/composable_env.py
hallerite: added in [7a7057f](https://github.com/PrimeIntellect-ai/verifiers/pull/1067/commits/7a7057f7ea76265ab92149efd00b84159c1f63ad)

verifiers/envs/experimental/composable/composable_env.py
mikasenghaas: but which field to use for this, `Harness` is just a dataclass, right?

verifiers/envs/experimental/composable/composable_env.py
```diff
@@ -0,0 +1,207 @@
...
+        A ``TaskSet`` or ``SandboxTaskSet``.
+    harness:
+        A ``Harness`` — the agent configuration.
+    """
+
+    def __init__(
+        self,
+        taskset: TaskSet,
+        harness: Harness,
+        test_timeout: int = 900,
```
mikasenghaas: `test_timeout` seems sandbox specific?

mikasenghaas: also its being overriden unconditionally from the sandbox task spec later? unless diff viewer is playing tricks on me

verifiers/envs/experimental/composable/composable_env.py
```diff
@@ -0,0 +1,207 @@
...
+    async def get_docker_image(self, state: State) -> str:
+        info = state.get("info") or {}
+        spec = self.taskset.get_sandbox_spec(info)
+        if spec:
+            return spec.image
+        if self.harness.sandbox_spec:
+            return self.harness.sandbox_spec.image
+        return self.docker_image
+
+    def get_sandbox_resources(self, state: State) -> dict[str, Any]:
```
mikasenghaas, line 99: wouldn't it make sense for return type to be sandbox spec here? and just `return spec` instead? or what is the point of fanning out into raw dicts here?

hallerite: the problem is `get_sandbox_resources` is on `CliAgentEnv`. I would like to make a second PR that is a bit more invasive soon that puts the `SandboxSpec` into a more central location, e.g. into `sandbox_mixin.py`

verifiers/envs/experimental/composable/composable_env.py
hallerite: yeah, fixed this

verifiers/envs/experimental/composable/task.py
```diff
@@ -0,0 +1,362 @@
...
+        self,
+        n: int | None = None,
+        concurrency: int = 10,
+    ) -> list[dict]:
+        """Validate instances. For sandbox tasks, creates sandboxes and runs validate_instance."""
+        import asyncio
+        import logging
+        import time
+
+        from prime_sandboxes import CreateSandboxRequest
```
snimu: if we lazily import them, why not after the `not is_sandbox` case?

snimu: CreateSandboxRequest and ThreadedAsyncSandboxClient that is

verifiers/envs/experimental/composable/task.py
```diff
@@ -0,0 +1,362 @@
...
+
+        logger = logging.getLogger(__name__)
+        ds = self.get_dataset()
+        total = min(n, len(ds)) if n is not None else len(ds)
+        is_sandbox = isinstance(self, SandboxTaskSet)
+
+        if not is_sandbox:
+            # No sandbox needed — just run validate_instance on each
+            results: list[dict] = []
+            for i in range(total):
```
snimu: why is this not run via `asyncio.gather`?

verifiers/envs/experimental/composable/task.py
```diff
@@ -0,0 +1,362 @@
...
+                            "valid": False,
+                            "elapsed": time.time() - t0,
+                            "error": str(e),
+                        }
+                    )
+            return results
+
+        # Sandbox path
+        client = ThreadedAsyncSandboxClient(
+            max_workers=min(max(1, concurrency // 8), 50)
```
snimu, line 248: why do we hard-limit the `max_workers`?

verifiers/envs/experimental/composable/task.py
```diff
@@ -0,0 +1,362 @@
...
+            f"Validating {total} instances from {self.name} (concurrency={concurrency})"
+        )
+        t0 = time.time()
+        try:
+            results = await asyncio.gather(*[validate_one(i) for i in range(total)])
+        finally:
+            client.teardown()
+        elapsed = time.time() - t0
+        passed = sum(1 for r in results if r["valid"])
+        logger.info(f"Validation: {passed}/{total} valid ({elapsed:.0f}s)")
```
snimu: converting passed/total into a pass-rate (in percent or 0-1) would be very helpful in this message

verifiers/envs/experimental/composable/harness.py
```diff
@@ -0,0 +1,58 @@
...
+        Default sandbox resources when the task doesn't provide a
+        SandboxSpec (e.g. math + OpenCode — the agent needs a sandbox
+        but the task doesn't specify one).
+    """
+
+    install_script: str | None = None
+    run_command: str = ""
+    system_prompt: str | None = None
+    instruction_path: str = "/task/instruction.md"
+    system_prompt_path: str = "/task/system_prompt.txt"
```
snimu, line 55: i know this is extremely pedantic but i'd prefer if the `system_prompt_path` were right below the `system_prompt` and above the `instruction_path`

verifiers/envs/experimental/composable/composable_env.py
```diff
@@ -0,0 +1,206 @@
...
+        super().__init__(run_command=harness.run_command, **kwargs)
+
+        self.taskset = taskset
+        self.harness = harness
+
+    # -- CliAgentEnv hooks --------------------------------------------------
+
+    def _get_spec(self, state: State) -> Any:
+        """Get SandboxSpec, cached on state to avoid redundant calls."""
+        cached = state.get("_sandbox_spec")
```
snimu, line 83: why do we use caching here at all? in what scenario is `self.taskset.get_sandbox_spec(info)` so expensive that we need it? i thought it was just returning static information?

verifiers/envs/experimental/composable/task.py
hallerite: https://github.com/PrimeIntellect-ai/verifiers/pull/1067#discussion_r2993890198

verifiers/envs/experimental/composable/composable_env.py
hallerite: in the case where we may need to make a network call, for example. this method is called thrice, so better to have it cached imo

snimu: makes sense

verifiers/envs/experimental/composable/README.md
```diff
@@ -0,0 +1,151 @@
...
+**SandboxSpec** — describes sandbox requirements per instance (image, CPU, memory, GPU type, timeout).
+
+**Harness** — agent-side configuration. Declares how to install and run an agent binary, and where it expects to find task-provided content (instruction, system prompt).
+
+**ComposableEnv** — a `CliAgentEnv` subclass that wires a TaskSet + Harness. Inherits all interception machinery unchanged.
+
+## Usage
+
+```python
+from swe_tasksets import R2EGymTaskSet
```
snimu, line 22: where are swe_tasksets and opencode_harness coming from? if they exist they should be linked, if not then it should be made clear that these are illustrative libraries

hallerite: they are in research-environments. I guess we could link them, but maybe wait until they are merged?

verifiers/envs/experimental/composable/task.py
```diff
@@ -0,0 +1,362 @@
...
+@dataclass
+class SandboxSpec:
+    """Sandbox requirements for a task instance."""
+
+    image: str = "python:3.11-slim"
+    cpu_cores: int = 4
+    memory_gb: int = 4
+    disk_size_gb: int = 10
+    gpu_count: int = 0
+    gpu_type: str | None = None
```
xeophon, line 44: i would prefer a `gpu_types: list[str] | None`, as it matches Harbor (see [gpu_types](https://harborframework.com/docs/tasks#environment:~:text=List%20of%20acceptable%20GPU%20types%20(e.g.%2C%20%5B%27H100%27%2C%20%27A100%27%2C%20%27T4%27%5D).%20None%20means%20any%20GPU%20type%20is%20acceptable.)) and would allow things like ["H100", "H200"] for all Hopper GPUs

hallerite: our SDK only takes a single gpu_type atm, hence why I have it as a single string

hallerite: as soon as the sdk supports this!

PR #1068, opened 2026-03-25
feat: add renderers package

hallerite:
## Summary

Adds `packages/renderers/` — a standalone package that owns message ↔ token conversion as an alternative to vLLM's Jinja chat templates and the existing TITO/MITO server-side machinery. Every sampled assistant turn keeps its exact tokens across the rollout boundary; the trainer sees bit-for-bit what vLLM produced.

> **Status: experimental.** Available alongside the production `openai_chat_completions_token` (TITO) client, which remains the recommended path for production training. Renderers offer stronger token-preservation guarantees but only ship hand-coded support for a subset of models, and corner cases are still being shaken out. The package is text-only — `Qwen3VLRenderer` works against the Qwen3-VL tokenizer for text conversations only; multimodal training should continue to use MITO.

**Full design, motivation, and examples of the failure modes this fixes:** [`packages/renderers/README.md`](./packages/renderers/README.md).

## Renderer matrix

| Template family | Renderer class | Models |
|---|---|---|
| Qwen chatml | `Qwen3Renderer` / `Qwen35Renderer` / `Qwen36Renderer` / `Qwen3VLRenderer` | Qwen3, Qwen3.5, Qwen3.6, Qwen3-VL (text-only) |
| GLM (next-turn markers) | `GLM5Renderer` / `GLM45Renderer` | GLM-5, GLM-5.1, GLM-4.5, GLM-4.7 |
| MiniMax chatml-variant | `MiniMaxM2Renderer` | MiniMax-M2 / M2.5 |
| Kimi im-role format | `KimiK2Renderer` / `KimiK25Renderer` | Kimi K2, K2.5, K2.6 |
| DeepSeek V3 tool-section | `DeepSeekV3Renderer` | DeepSeek V3 |
| Nemotron chatml + XML tools | `Nemotron3Renderer` | Nemotron-3 Nano / Super |
| OpenAI harmony | `GptOssRenderer` | GPT-OSS |
| Fallback (`apply_chat_template`) | `DefaultRenderer` | anything else, with optional `tool_parser` / `reasoning_parser` |

## Architecture

```
messages → renderer.render_ids() → [token IDs]
                                       ↓
                              vLLM /v1/generate
                                       ↓
[completion IDs] → renderer.parse_response() → ParsedResponse(content, reasoning_content, tool_calls)
```

The Renderer Protocol:
- `render()` / `render_ids()` — messages → tokens with per-token message attribution for loss masking.
- `parse_response()` — completion tokens → structured message via token-ID boundary scanning (no regex on decoded text).
- `get_stop_token_ids()` — turn-close tokens.
- `bridge_to_next_turn()` — extends `prev_prompt + prev_completion` with the new turn's tokens; returns `None` if the renderer can't prove prefix-stability (caller falls back to fresh render).

## Key design decisions

- **Per-renderer bridges, hand-coded.** No shared `chatml_bridge` / `glm_bridge` helper — that approach rendered `[dummy_assistant, *new_messages]` and diffed against `[dummy_assistant]` to extract extension tokens, which broke on templates that treated the dummy as an invalid prefix (GLM-5.1 wraps the last assistant with empty `<think></think>`, harmony's assistant uses different channels historical vs latest, Kimi auto-injects a default system). Each renderer's bridge now hand-emits the new-turn tokens by calling the same per-role inline helpers that `render()` uses, so the two paths can't silently diverge. Two small shared primitives remain: `trim_to_turn_close` (scan `prev_completion` for a template-specific close token; on truncation, append the canonical close so the bridge still extends) and `reject_assistant_in_extension` (bridges refuse to re-tokenize model-sampled assistant content).

- **Truncation handling.** Hand-coded renderers always synthesize the canonical turn-close (`<|im_end|>`, `<|endoftext|>`, harmony's `<|end|>`, …) when vLLM hits `max_tokens` and the prior completion has no close token, so the next prompt still extends the prior step's tokens verbatim; the synthetic token lands in the merged sample's `prompt_ids` (mask=False) and never enters loss or KL. `DefaultRenderer.bridge_to_next_turn` returns `None` unconditionally — it wraps an unknown Jinja template and can't prove the extension contract holds — so the caller falls back to a fresh re-render.

- **Pluggable parsers for `DefaultRenderer`.** Hand-coded renderers bake parsing in. `DefaultRenderer` takes optional `tool_parser=` / `reasoning_parser=` kwargs wired to registries in `renderers.parsers`. Built-ins today: `qwen3`, `qwen3.5`, `glm`, `deepseek_v3` for tools; `think` for reasoning.

## RendererClient + `/v1/generate`

Adds `RendererClient` (verifiers client type `"renderer"`) — renders messages client-side, POSTs raw token IDs to vLLM's `/v1/generate`, parses completions back into structured responses. Multi-turn rollouts reuse the prior step's exact tokens through `bridge_to_next_turn`; no re-rendering of sampled content.

A `RendererPool` offloads sync tokenization to threads so concurrent rollouts tokenize in parallel instead of blocking the event loop.

## Test plan

- [x] `packages/renderers/tests/test_render_ids.py` — multi-model parity matrix vs `apply_chat_template` (1 documented xfail for an upstream Jinja bug on Qwen3-VL `content=None`).
- [x] `packages/renderers/tests/test_roundtrip.py` — render → parse round-trip per renderer: content, reasoning, single and multiple tool-calls.
- [x] `packages/renderers/tests/test_bridge.py` — bridge contract invariants per hand-coded renderer: extends prev verbatim, rejects assistant-role extension, synthesizes close on truncation, extension contains the new-message content.
- [x] `packages/renderers/tests/test_incremental.py` — unit coverage of `trim_to_turn_close` + `reject_assistant_in_extension` edge cases.
- [x] `packages/renderers/tests/test_parsers.py` / `test_parse_response.py` / `test_parse_response_robustness.py` — parsing on truncated / malformed output; includes regression test for `parse_qwen3` JSON-decode-error fallback.
- [x] `tests/test_renderer_e2e.py` — end-to-end TITO rollout with scripted vLLM; asserts token preservation and multi-turn bridge extension.
- [x] E2E in prime-rl [#2278](https://github.com/PrimeIntellect-ai/prime-rl/pull/2278): wordle multi-turn `samples_per_rollout` = 1.00, reward + KL track main. Qwen3.5-35B-A3B + mini-swe-agent-plus: 0 breaks vs main's 32 in the same step (see README §4 for the concrete break modes).

---

> [!NOTE]
> **Medium Risk**
> Large, mostly additive change that introduces new tokenization/parsing/bridging logic and a new dependency; errors here can subtly affect rollout correctness and RL training data integrity, though existing client paths remain available.
> 
> **Overview**
> Introduces a new standalone `packages/renderers` package that implements a `Renderer` protocol, per-model message→token renderers (Qwen/GLM/Kimi/DeepSeek/MiniMax/Nemotron/GPT-OSS) with `bridge_to_next_turn` support, and a fallback `DefaultRenderer` that wraps `tokenizer.apply_chat_template` with optional tool/reasoning parsers.
> 
> Adds a renderer-backed inference path (`renderers.client.completions_request`) that sends `prompt_token_ids` to vLLM’s `/generate` endpoint, parses completion token IDs back into structured outputs, and supports parallel tokenization via `RendererPool`.
> 
> Updates docs (`evaluation.md`, `training.md`, `faqs.md`, `reference.md`) to document the new `renderer` `client_type`, expand client type listings/descriptions, and clarify RL training tradeoffs between MITO/TITO/renderer approaches.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ee6cdb5e2fd3a8e0e6b4ed0fb219085c44567062. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/renderers/renderers/client.py
```diff
@@ -0,0 +1,210 @@
...
+        )
+
+    pool = renderer if isinstance(renderer, RendererPool) else None
+
+    # -- Prepare: tokenize prompt --
+    def _prepare(r: Renderer):
+        prepared_prompt_ids = (
+            list(prompt_ids)
+            if prompt_ids is not None
+            else r.render_ids(messages, tools=tools, add_generation_prompt=True)
```
eligotts, line 68: Does this path need to honor chat-template kwargs from sampling_args like `enable_thinking`? if sampling args have `enable_thinking=False` we should probably pipe that through so then then renderer will automatically add the close `</think>` tag

packages/renderers/renderers/client.py
```diff
@@ -0,0 +1,210 @@
...
+        prompt_ids, stop_token_ids = _prepare(renderer)
+
+    # -- Build request body --
+    body: dict[str, Any] = {
+        "model": model,
+        "prompt_token_ids": prompt_ids,
+        "stop_token_ids": stop_token_ids,
+    }
+
+    for key in ["temperature", "top_p", "seed", "n"]:
```
eligotts: why only these keys? could there be other sampling args we need?

packages/renderers/renderers/client.py
```diff
@@ -0,0 +1,210 @@
...
+        stop_token_ids = r.get_stop_token_ids()
+        return prepared_prompt_ids, stop_token_ids
+
+    if pool is not None:
+        prompt_ids, stop_token_ids = await _run_pooled(pool, _prepare)
+    else:
+        prompt_ids, stop_token_ids = _prepare(renderer)
+
+    # -- Build request body --
+    body: dict[str, Any] = {
```
eligotts, line 79: for VLM are we doing renderers? if so need to also pass image features

verifiers/clients/renderer_client.py
```diff
@@ -0,0 +1,670 @@
...
+            args["max_completion_tokens"] = args.pop("max_tokens")
+
+        prompt_ids = await _get_incremental_prompt_ids(
+            renderer=renderer,
+            prompt=prompt,
+            state=kwargs.get("state"),
+            tools=tools,
+        )
+
+        return await completions_request(
```
eligotts, line 484: what about kwargs["extra_headers"] ? should we pass those through?

packages/renderers/renderers/gpt_oss.py
```diff
@@ -0,0 +1,538 @@
...
+    def _render_tool_result(
+        self,
+        msg: Message,
+        msg_idx: int,
+        *,
+        emit_special,
+        emit_text,
+    ) -> None:
+        """Emit a tool result message in Harmony commentary format."""
+        tool_name = msg.get("name") or "unknown"
```
eligotts: RendererClient parses the GPT-OSS assistant tool call into ToolCall(name=...), but ToolEnv returns ToolMessage with only content/tool_call_id, and _to_renderer_message forwards only role/content/tool_call_id. As a result every GPT-OSS tool result rendered through this path becomes functions.unknown to=assistant. This affects both full renders and bridge_to_next_turn

verifiers/clients/renderer_client.py
```diff
@@ -0,0 +1,670 @@
...
+    return fn(renderer)
+
+
+def _get_value(obj: Any, key: str, default: Any = None) -> Any:
+    if isinstance(obj, Mapping):
+        return obj.get(key, default)
+    return getattr(obj, key, default)
+
+
+def _normalize_for_comparison(value: Any, _key: str | None = None) -> Any:
```
eligotts, line 114: need to add in here the @mikasenghaas change about content being `""` or `None` , choosing a unified route, see https://github.com/PrimeIntellect-ai/verifiers/commit/237e15b2f9c6318e99d0d666ca93262735625b18

packages/renderers/renderers/client.py
hallerite: not for now.

packages/renderers/renderers/client.py
hallerite: should probably mirror the other clients

verifiers/clients/renderer_client.py
hallerite: probably yes

packages/renderers/renderers/gpt_oss.py
hallerite: Fixed in d5198706 — added `_attach_tool_call_names` that walks the message list and recovers the function name from the prior assistant `tool_calls` (matched by `tool_call_id`), then attaches `name` on the renderer dict before render. Wired into both `to_native_prompt` and `_step_rendered_messages` so the fix applies to first turns and to incremental-bridge paths.

Validated end-to-end on GPT-OSS-20b: rendered tokens decode to `functions.calculator to=assistant` instead of `functions.unknown`. Pinned by 3 unit tests in `tests/test_renderer_client.py` and a parametrized parity test across all (model, renderer) pairs in `packages/renderers/tests/test_render_ids.py::test_tool_response_with_name`.

packages/renderers/renderers/client.py
hallerite: Fixed in e6c93674 — broadened the allowlist to the full set vLLM /generate accepts (`temperature`, `top_p`, `top_k`, `min_p`, `seed`, `n`, `repetition_penalty`, `min_tokens`, `prompt_logprobs`, `priority`, `cache_salt`). Top-level `sampling_args` wins, with `extra_body` as fallback for OpenAI-SDK-style callers.

verifiers/clients/renderer_client.py
hallerite: Fixed in e6c93674 — `extra_headers` from `sampling_args` is now passed through to `client.post` via `RequestOptions.headers`. Added in `packages/renderers/renderers/client.py::completions_request`, so the passthrough applies to every caller (renderer client, plus any direct caller).

packages/renderers/renderers/client.py
hallerite: Deferring this to a follow-up — properly piping `enable_thinking` (and other chat-template kwargs) needs a Renderer Protocol change since it is currently a constructor-time arg on the renderers that support it (DeepSeek V3, GLM4.5, GLM5, Kimi K2/K2.5, MiniMax-M2). Will spec a `chat_template_kwargs` argument on `render_ids` separately so it threads through the per-call path.

packages/renderers/renderers/gpt_oss.py
```diff
@@ -0,0 +1,538 @@
...
+            return None
+
+        previous_ids = list(previous_prompt_ids) + list(previous_completion_ids)
+        # Harmony stops cleanly on <|return|> or <|call|>; content beyond
+        # those means vLLM hit max_tokens. Synthesise <|end|> — the
+        # non-terminal close — so the template stays well-formed.
+        if previous_ids[-1] not in {self._return, self._call}:
+            if not self.synthesize_close_on_truncation:
+                return None
+            previous_ids = previous_ids + [self._end]
```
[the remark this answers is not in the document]
hallerite: Fixed in cd2d9911 — switched `GptOssRenderer.bridge_to_next_turn` to use `trim_to_turn_close` (the canonical helper that the other hand-coded renderers use). It scans only within `previous_completion_ids`, so an empty completion no longer indexes into the prompt; if no close is found and `synthesize_close_on_truncation` is on, it appends `<|end|>` after the prompt cleanly.

packages/renderers/renderers/client.py
```diff
@@ -0,0 +1,210 @@
...
+        )
+    except BadRequestError as exc:
+        _log_overlong_prompt_diagnostic(
+            prompt_ids=prompt_ids,
+            messages=messages,
+            max_tokens=max_tokens,
+            exc=exc,
+        )
+        raise
+    choice = data.get("choices", [{}])[0]
```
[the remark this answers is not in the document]
hallerite: Fixed in cd2d9911 — `data.get("choices") or [{}]` so an existing-but-empty list also falls back. The `[{}]` default of `dict.get` only kicks in when the key is missing entirely.

packages/renderers/renderers/kimi_k2.py
```diff
@@ -0,0 +1,456 @@
...
+                    msg, oi, content, emit_special=emit_special, emit_text=emit_text
+                )
+
+            else:
+                # Unknown role: use system-style formatting
+                emit_special(self._im_system, oi)
+                emit_text(role, oi)
+                emit_special(self._im_middle, oi)
+                emit_text(content, oi)
+                emit_special(self._im_end, i)
```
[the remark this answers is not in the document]
hallerite: Fixed in cd2d9911 — `i` → `oi` so the `<|im_end|>` token is attributed to the caller-relative message index, matching every other `emit_*` in the same block.

packages/renderers/renderers/gpt_oss.py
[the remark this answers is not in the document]
hallerite: Fixed in cd2d9911 — `GptOssRenderer.bridge_to_next_turn` now uses `trim_to_turn_close`, so trailing tokens after a stop marker are trimmed back to the last close boundary instead of being copied into the next turn's prompt. Same change addresses the empty-completion edge case in #3140102313.

packages/renderers/renderers/kimi_k25.py
```diff
@@ -0,0 +1,987 @@
...
+                )
+
+            else:
+                # Unknown roles use system-style formatting
+                emit_special(self._im_system, i)
+                emit_text(role, i)
+                emit_special(self._im_middle, i)
+                emit_text((msg.get("content") or ""), i)
+                emit_special(self._im_end, i)
+                emit_text("\n", i)
```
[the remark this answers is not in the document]
hallerite: Fixed in bccc4886 — and went further than the index fix. The renderer was a stale K2 copy-paste; reconciled it with the K2.5 chat template end-to-end:

- Dropped auto-system injection (template doesn't auto-inject)
- Removed every `\n` after `<|im_end|>` (template uses `{%- -%}` to strip whitespace)
- Added the template's hist/suffix split (history assistants render `<think></think>`, suffix preserves `reasoning_content`)
- Tool-decl now uses the TS encoder (K2.5's tokenizer auto-computes `tools_ts_str`); fixed a separate bug where `_encode_tools_typescript` passed the outer `{"type":"function","function":{...}}` dict to a function expecting the inner one
- Tool result body emits `## Return of {id}\n` unconditionally (template interpolates empty string when missing)
- Parser partitions on `<think>` first to avoid leaking pre-tag tokens into `reasoning_content`

K2.5 and K2.6 are now in the conftest matrix and pass all 778 renderer-package tests (incl. the new `test_message_indices_in_range` that would have caught the original index-shift class of bug).

packages/renderers/renderers/minimax_m2.py
```diff
@@ -0,0 +1,406 @@
...
+
+    def __init__(
+        self,
+        tokenizer: PreTrainedTokenizer,
+        *,
+        enable_thinking: bool = True,
+        default_system: str = _DEFAULT_SYSTEM,
+    ):
+        self._tokenizer = tokenizer
+        self._enable_thinking = enable_thinking
```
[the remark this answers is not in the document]
hallerite: Fixed in bccc4886 — removed the dead `enable_thinking` constructor kwarg from `MiniMaxM2Renderer`. It was stored on the instance but never read; generation prompt unconditionally emitted `<think>\n`. If we want per-call thinking control later it should go through `chat_template_kwargs` on the Renderer Protocol (deferred per #3150020314), not a construction-time flag that silently no-ops.

packages/renderers/renderers/kimi_k25.py
```diff
@@ -0,0 +1,994 @@
...
+    """
+    if not tools:
+        return ""
+    functions = []
+    for tool in tools:
+        if tool.get("type") != "function":
+            continue
+        func_def_dict = tool.get("function") or {}
+        if not func_def_dict:
+            continue
```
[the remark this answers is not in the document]
hallerite: Fixed in b11b9ca0 — encoder now accepts both shapes: unwraps `tool["function"]` when the OpenAI envelope is present, otherwise treats the dict as a flat `ToolSpec`. Non-function envelope types (e.g. `"_plugin"`) are still skipped.

The bug slipped through because the tests use the OpenAI envelope format while production callers pass `ToolSpec`. Test fixture wasn't representative.

docs/evaluation.md
```diff
@@ -102,7 +102,7 @@ env.set_concurrency(256)
 | `--model` | `-m` | `openai/gpt-4.1-mini` | Model name or endpoint alias |
 | `--api-base-url` | `-b` | `https://api.pinference.ai/api/v1` | API base URL |
 | `--api-key-var` | `-k` | `PRIME_API_KEY` | Environment variable containing API key |
-| `--api-client-type` | — | `openai_chat_completions` | Client type: `openai_chat_completions`, `openai_completions`, `openai_chat_completions_token`, or `anthropic_messages` |
+| `--api-client-type` | — | `openai_chat_completions` | Client type: `openai_chat_completions`, `openai_completions`, `openai_chat_completions_token`, `renderer`, or `anthropic_messages` |
```
mikasenghaas: should it be `renderer` or `renderers`?

verifiers/types.py
```diff
@@ -426,6 +427,11 @@ class ClientConfig(BaseModel):
 
     client_idx: int = 0
     client_type: ClientType = "openai_chat_completions"
+    renderer: str = "auto"
```
mikasenghaas, line 431: hm in wondering if we should bundle those somehow? so like a discriminated union of client types with some shared args but mostly disjoint

verifiers/clients/renderer_client.py
```diff
@@ -0,0 +1,749 @@
...
+    # reasoning_parser, pool_size) so that different parser configs or pool
+    # sizes for the same model don't collide.
+    _shared_pools: ClassVar[
+        dict[tuple[str, str, str | None, str | None, int], RendererPool]
+    ] = {}
+    _shared_pools_lock: ClassVar[threading.Lock] = threading.Lock()
+
+    def __init__(
+        self,
+        config: ClientConfig,
```
mikasenghaas: ya i think at this point we can justify having client-specific configs like `RendererClientConfig` which are registered as discriminated union (same pattern as e.g. optimizer config in prime-rl)

verifiers/clients/renderer_client.py
```diff
@@ -0,0 +1,749 @@
...
+        renderer: Renderer | None = None,
+        pool_size: int = _DEFAULT_POOL_SIZE,
+    ):
+        super().__init__(config)
+        self._renderer = renderer
+        # ClientConfig.renderer_pool_size wins over the constructor default so
+        # callers can tune pool size via config without subclassing.
+        cfg_size = getattr(config, "renderer_pool_size", None)
```
mikasenghaas: then we can remove this `getattr` madness

verifiers/clients/renderer_client.py
```diff
@@ -0,0 +1,749 @@
...
+        renderer_name = self._config.renderer if self._config is not None else "auto"
+        renderer_model = (
+            self._config.renderer_model_name
+            if self._config is not None and self._config.renderer_model_name is not None
+            else model
+        )
+        tool_parser = self._config.tool_parser if self._config is not None else None
+        reasoning_parser = (
+            self._config.reasoning_parser if self._config is not None else None
+        )
```
mikasenghaas: same here, would remove a lot of line noise

verifiers/clients/renderer_client.py
```diff
@@ -0,0 +1,749 @@
...
+
+        with self._shared_pools_lock:
+            if cache_key not in self._shared_pools:
+
+                def factory(
+                    _name=renderer_name,
+                    _model=renderer_model,
+                    _tool_parser=tool_parser,
+                    _reasoning_parser=reasoning_parser,
+                ) -> Renderer:
```
mikasenghaas: can we not make this `factory() -> Renderer`

verifiers/clients/renderer_client.py
```diff
@@ -0,0 +1,749 @@
...
+                reasoning_content=reasoning_content,
+                finish_reason=finish_reason,
+                is_truncated=finish_reason == "length",
+                tokens=tokens,
+                tool_calls=tool_calls,
+            ),
+        )
+
+
+# ── Helpers ─────────────────────────────────────────────────────────
```
mikasenghaas, line 94: can we inline them? more in the style of the other clients. if smth is shared across components can prob extract into `verifiers.clients.utils`

verifiers/clients/openai_chat_completions_client.py
```diff
@@ -161,6 +164,8 @@ async def close(self) -> None:
     async def to_native_prompt(
         self, messages: Messages
     ) -> tuple[OpenAIChatMessages, dict]:
+        messages = maybe_normalize_messages(messages, field_name="prompt")
```
mikasenghaas: why this?

packages/renderers/renderers/base.py
```diff
@@ -0,0 +1,567 @@
...
+
+logger = logging.getLogger("renderers.base")
+
+
+# ---------------------------------------------------------------------------
+# Message types — strong typing for the conversation data model
+# ---------------------------------------------------------------------------
+
+
+class TextPart(TypedDict):
```
mikasenghaas, line 17: hmm is there anything functionally different from the vf types here? or is this purely for package isolation?

packages/renderers/renderers/base.py
```diff
@@ -0,0 +1,567 @@
...
+# Renderer data types
+# ---------------------------------------------------------------------------
+
+
+@dataclass
+class RenderedTokens:
+    """Result of rendering messages to tokens.
+
+    Each token carries an index into the original message list so callers can
+    build per-token loss masks without re-rendering.  Tokens from structural
```
mikasenghaas, line 85: ah interesting, are those trained on by default? bc right now e.g. `<think>` is part of the prompt (chat template) and sometimes the model is expected to generate it, so basically whether or not we train on it depends on the chat template. is this behavior preserved here?

packages/renderers/renderers/base.py
```diff
@@ -0,0 +1,567 @@
...
+    tool_calls: list[dict[str, Any]] | None = None
+
+
+@dataclass
+class RenderedConversation:
+    """Exact token state for a rendered conversation."""
+
+    prompt_ids: list[int]
+    completion_ids: list[int] = field(default_factory=list)
+    completion_logprobs: list[float] = field(default_factory=list)
```
mikasenghaas, line 108: would it make sense to already collapse `prompt_ids` and `completion_ids` into `messages_id` or just `token_ids`

mikasenghaas: i think in multiturn the definition of the prompt is somewhat irrelevant

packages/renderers/renderers/base.py
```diff
@@ -0,0 +1,567 @@
...
+            "deepseek_v3": DeepSeekV3Renderer,
+            "kimi_k2": KimiK2Renderer,
+            "kimi_k25": KimiK25Renderer,
+            "nemotron3": Nemotron3Renderer,
+            "gpt_oss": GptOssRenderer,
+        }
+    )
+
+
+def create_renderer_pool(
```
mikasenghaas, line 327: this is duplicated in the renderer client no?

packages/renderers/README.md
```diff
@@ -0,0 +1,212 @@
...
+- **Tool-call parsing** — `ToolParser` implementations scan token ids for the model's special delimiter tokens (e.g. token id `151657` for `<tool_call>` on Qwen3). A regex-on-decoded-text parser can mistake a literal `"<tool_call>"` inside user content for a real tool-call opener; matching by id can't, because regular text never tokenizes to the special-token id.
+- **Reasoning parsing** — `ReasoningParser` implementations split `<think>…</think>` with a tested whitespace-preservation contract (bit-exact newline handling around the closing tag — a spot where we had a real bug that silently broke the extension property at the think boundary).
+
+### What we gain
+
+- **RL correctness.** A prompt/completion split we control, which is exactly what `bridge_to_next_turn` relies on to keep rollouts from fragmenting under truncation or re-tokenization.
+- **Testable parity.** Per-model renderers are plain Python. We can render the same conversation through the renderer and through HF's `apply_chat_template` and assert token-level parity. Every edge case (empty thinking, multiple tool calls, truncated turns) becomes a unit test instead of undefined behavior buried inside Jinja.
```
mikasenghaas, line 24: LFGGG

packages/renderers/README.md
```diff
@@ -0,0 +1,212 @@
...
+- **Escape hatch.** Anything without a hand-coded renderer falls back to `DefaultRenderer` (a generic `apply_chat_template` wrapper), which mirrors the previous TITO path.
```
mikasenghaas, line 25: what sort of guarantees, if any, can we give for this fallback? what are the pros and cons compare to fallback to mito?

packages/renderers/README.md
```diff
@@ -0,0 +1,212 @@
...
+
+`ParsedResponse` is `(content, reasoning_content, tool_calls)`.
+
+Round-trip invariant: rendering `[user, assistant(content=X, reasoning=Y, tool_calls=[T])]` to token ids, slicing out the assistant completion, and feeding it through `parse_response` returns an equivalent structured assistant message (modulo field formatting). Tested per-renderer in `tests/test_roundtrip.py`.
+
+### Picking a renderer
+
+```python
+from transformers import AutoTokenizer
+from renderers.base import create_renderer
```
mikasenghaas: should we export this into  top-level for cleaner import?

packages/renderers/README.md
```diff
@@ -0,0 +1,212 @@
...
+
+tok = AutoTokenizer.from_pretrained("Qwen/Qwen3-8B")
+r = create_renderer(tok, renderer="auto")   # → Qwen3Renderer
+```
+
+`renderer="auto"` matches `tokenizer.name_or_path` against `MODEL_RENDERER_MAP` by **exact match**. Prefix matching is intentionally off: two models with the same architecture can ship different chat templates (base vs instruct, fine-tune renames), and prefix routing would silently pick a renderer that doesn't produce template-parity output. Fine-tunes must pass `renderer=<name>` explicitly; unknown names fall back to `DefaultRenderer`.
+
+Available hand-coded renderers: `qwen3`, `qwen3_vl`, `qwen3.5`, `glm5`, `glm4.5`, `minimax-m2`, `deepseek_v3`, `kimi_k2`, `kimi_k25`, `nemotron3`, `gpt_oss`, `default`.
```
mikasenghaas, line 61: would maybe be cool if we could have a small cli which can check if <model_name> is compatible with <render> by running the unit tests for that model<>renderer combo? would be quite useful for quick checks/ integratins maybe?

packages/renderers/README.md
```diff
@@ -0,0 +1,212 @@
...
+
+The bridge is what produces the tokens needed to uphold the extension property in accordance with the chat template the model was trained on.
+
+### Per-renderer bridges
+
+Each hand-coded renderer implements `bridge_to_next_turn` directly for its model's chat template — no shared generic helper, just Python that knows what tokens the template would insert between turns. Qwen3's bridge knows about `<|im_start|>role\n … <|im_end|>\n`; GLM's bridge knows that turns end when the next role marker appears; DeepSeek V3, Kimi K2/K2.5, Nemotron-3, GPT-OSS, MiniMax each have their own. On a clean stop, vLLM's `completion_ids` already includes the template's close token; on truncation, the renderer synthesizes the canonical close (`<|im_end|>`, `<|endoftext|>`, or the equivalent for that model) so the extension invariant still holds, and the synthetic close is masked out of the loss because the model didn't produce it.
```
mikasenghaas, line 96: legendary

packages/renderers/renderers/base.py
hallerite: renderers does not consume verifiers, hence it can't use verifiers types

packages/renderers/renderers/base.py
hallerite: I think so, but I'll flag it to check this again

packages/renderers/README.md
hallerite: yeah, let me be more clear about this

packages/renderers/renderers/base.py
hallerite: I think we should do this in a follow-up PR, but yes. I think it's time.

docs/evaluation.md
hallerite: I think `renderer`, because the renderer is an object and we have one renderer client

verifiers/clients/renderer_client.py
hallerite: Yeah, but I would do this in a follow-up PR, too

packages/renderers/README.md
hallerite: Yeah, but would do this in an extra PR. Just to get this in sooner.

packages/renderers/renderers/base.py
hallerite: indeed

verifiers/clients/openai_chat_completions_client.py
hallerite: good catch

mikasenghaas (approved):

eligotts (approved):

PR #1101, opened 2026-04-03
Add max_total_tokens parameter to MultiTurnEnv

samsja:
## Summary
- Adds `max_total_tokens` parameter to `MultiTurnEnv` to limit total output tokens across all turns in a multi-turn rollout
- Uses the existing `StateUsageTracker` to accumulate token usage
- Disabled by default (`max_total_tokens=-1`)

## Usage
```python
env = MyMultiTurnEnv(max_total_tokens=1000)  # Stop after 1000 output tokens
```

Works for all environments that inherit from `MultiTurnEnv` (ToolEnv, StatefulToolEnv, SandboxEnv, etc.).

---

> [!NOTE]
> **Medium Risk**
> Adds a new built-in termination condition to multi-turn rollouts based on accumulated output token usage, which can change when/why rollouts stop across all `MultiTurnEnv` subclasses. Behavior is disabled by default but misconfiguration or missing usage tracking could lead to unexpected early/late termination.
> 
> **Overview**
> Adds an optional `max_total_completion_tokens` limit to `MultiTurnEnv`, including a new built-in stop condition `max_total_completion_tokens_reached` that ends a rollout once cumulative output tokens (via `get_state_usage`) meet the configured threshold.
> 
> Updates the API/docs to document the new stop condition and exposes `set_max_total_completion_tokens()` as a configuration hook.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5d66162a18da1dbd9f394aeb6fb7f169279d7bbe. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/verifiers/envs/multiturn_env.py
+++ b/verifiers/envs/multiturn_env.py
@@ -35,12 +35,16 @@ class MultiTurnMonitorRubric(vf.Rubric):
 
 
 class MultiTurnEnv(vf.Environment):
-    def __init__(self, max_turns: int = -1, max_total_tokens: int = -1, **kwargs):
+    def __init__(self, max_turns: int = -1, **kwargs):
         super().__init__(**kwargs)
         self.max_turns = max_turns
-        self.max_total_tokens = max_total_tokens
+        self.max_total_completion_tokens: int = -1
 
         self.add_rubric(MultiTurnMonitorRubric())
+
+    def set_max_total_completion_tokens(self, max_total_completion_tokens: int) -> None:
+        """Set the maximum total completion tokens for this environment."""
+        self.max_total_completion_tokens = max_total_completion_tokens
 
     @abstractmethod
     async def env_response(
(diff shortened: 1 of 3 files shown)


snimu (approved):

verifiers/envs/multiturn_env.py
```diff
@@ -35,9 +35,10 @@ async def num_turns(self, state: State) -> int:
 
 
 class MultiTurnEnv(vf.Environment):
-    def __init__(self, max_turns: int = -1, **kwargs):
+    def __init__(self, max_turns: int = -1, max_total_tokens: int = -1, **kwargs):
```
mikasenghaas: this is in user space, we cannot configure this for arbitrary envs from prime-rl. rn, a user would have to expose `max_total_tokens` in `load_environment` and manually pass it through to the env constructor

prob want a solution that can hook into any env (not make this user space)

verifiers/envs/multiturn_env.py
```diff
@@ -38,9 +38,14 @@ class MultiTurnEnv(vf.Environment):
     def __init__(self, max_turns: int = -1, **kwargs):
```
mikasenghaas, line 38: wondering if we should demote `max_turns` to an "internal" arg as well, feels like we should handle it the same way as `max_seq_len` and `max_total_completion_tokens`

mikasenghaas (approved):
only nits, generally happy with this -- should ship! and do companion pr to prime-rl to make this configurable in `EnvConfig` which then auto-populates `extra_env_kwargs`

verifiers/envs/multiturn_env.py
eligotts: i feel like `max_turns` is more tied to the real logic of a rollout? Should still definitely let an env builder set it if they want to

PR #1194, opened 2026-04-19
fix: make interception proxy streaming resilient to upstream cuts

rasdani:
Makes the interception proxy's streaming response resilient to upstream cuts: 10s SSE keepalive comments keep idle streams warm, a per-chunk asyncio.sleep(0) forces an event-loop yield so content+close can't race the transport flush under warmup-burst contention, and transport exceptions at prepare/write/write_eof are surfaced as StreamInterrupted into state["error"] (builds on #1191) so rollouts reschedule instead of looking like clean zero-turn completions.
Validated on job 5814 (opencode-swe, tasks_per_minute=256, pristine frpc+prime-tunnel): 0 silent turns=0-agent_completed and 0 StreamInterrupted in 19 min / 771 rollouts; ablations confirm both mechanisms are load-bearing.


verifiers/utils/interception_utils.py
```diff
@@ -224,30 +227,99 @@ async def _handle_streaming_response(
...
+                # originally surfaced this effect (reverted in 4f900a23).
+                await asyncio.sleep(0)
 
         except asyncio.CancelledError:
             logger.debug(f"[{rollout_id}] Streaming cancelled")
         except Exception as e:
-            logger.error(f"[{rollout_id}] Streaming error: {e}")
+            waited_s = time.monotonic() - start
+            logger.debug(
+                f"[{rollout_id}] Streaming error after {waited_s:.1f}s: {e}"
```
mikasenghaas: we have a nice `print_time` util for this, just fyi

rasdani: Good call — switched to `print_time` in 0f7df95.

mikasenghaas (approved):

PR #1216, opened 2026-04-20
fix: preserve inner task names and raise on unknown task in EnvGroup

### LINKED ISSUE #1008: EnvGroup nested in outer EnvGroup silently misroutes all tasks to envs[0]
## Bug: Nested EnvGroup task routing is silently broken

When an `EnvGroup` is wrapped inside another `EnvGroup` (as prime-rl does when it wraps user envs), the outer group overwrites the `task` column to a single name. The inner `EnvGroup` then fails to route because its `env_map` uses the original task names — and **silently falls back to `envs[0]`** for rollouts, or returns `reward=0.0` for scoring.

This means **only the first sub-environment's rubric ever executes**, and the user gets no error or warning that the rest of their tasks are being silently misrouted or scored as zero.

## Impact

If a user builds a multi-task environment using `EnvGroup` (e.g., 17 tasks with different rubrics) and registers it as a single `[[orchestrator.env]]` in prime-rl, only the first task type gets meaningful reward signal. The model trains with zero reward on all other task types. There is no error or warning — the only symptom is that wandb metrics only show the first rubric's reward name.

### 1. Task name overwriting (`EnvGroup.__init__`, lines 172-182)

```python
for env, name in zip(self.envs, self.env_names):
    add_task = make_add_task_fn(name)
    env_dataset = env.build_dataset()
    if env_dataset is not None:
        if "task" in env_dataset.column_names:
            env_dataset = env_dataset.remove_columns(["task"])  # destroys inner task names
        env_dataset = env_dataset.map(add_task, **map_kwargs)   # overwrites with outer name
```

When the sub-env is itself an `EnvGroup`, its dataset already has meaningful per-task names. The outer `EnvGroup` unconditionally destroys these and replaces them all with a single name.

### 2. Silent fallback in `get_env_for_task` (line 318-319)

```python
def get_env_for_task(self, task: str) -> vf.Environment:
    return self.env_map.get(task, self.envs[0])  # silent fallback!
```

When the inner `EnvGroup` receives the outer name (which doesn't exist in its `env_map`), it silently falls back to `envs[0]`. All rollouts go to the first sub-environment. No warning is logged.

### 3. Scoring returns zeros for unknown tasks (`EnvGroupRubric.score_group`, lines 86-94)

```python
task = states[0].get("task", "default")
env = self.env_map.get(task)
if env is None:
    self.logger.warning(f"No environment found for task '{task}'")
    for state in states:
        state["reward"] = 0.0  # all tasks scored as zero
```

The inner `EnvGroupRubric` can't find the outer name in its `env_map`, so all states get `reward=0.0`.

### A. Preserve inner task names when sub-env is an EnvGroup

In `EnvGroup.__init__`, when a sub-env is itself an `EnvGroup`, preserve its internal task names instead of overwriting them. Register each inner task name in the outer `env_map` pointing to the sub-env:

```python
for env, name in zip(self.envs, self.env_names):
    add_task = make_add_task_fn(name)
    env_dataset = env.build_dataset()
    if env_dataset is not None:
        if isinstance(env, EnvGroup) and "task" in env_dataset.column_names:
            # Preserve inner EnvGroup's task names for correct routing
            for inner_name in env.env_names:
                self.env_map[inner_name] = env
            # Don't overwrite task column — inner tasks are already set
        else:
            if "task" in env_dataset.column_names:
                env_dataset = env_dataset.remove_columns(["task"])
            env_dataset = env_dataset.map(add_task, **map_kwargs)
        datasets.append(env_dataset)
```

Same pattern for `eval_dataset` handling.

This way:
- Inner task names are preserved in the dataset
- Outer `env_map` maps each inner task name to the inner `EnvGroup`
- Routing works: outer gets a task name, routes to inner `EnvGroup`, inner routes to correct sub-env
- `EnvGroupRubric` also gets the updated `env_map`, so scoring routes correctly
- `results_df.task.nunique() > 1`, enabling per-task logging in orchestrators like prime-rl

### B. Remove silent fallback in `get_env_for_task`

```python
def get_env_for_task(self, task: str) -> vf.Environment:
    env = self.env_map.get(task)
    if env is None:
        available = list(self.env_map.keys())
        raise ValueError(
            f"No environment found for task '{task}'. "
            f"Available tasks: {available}"
        )
    return env
```

The current fallback to `envs[0]` silently masks routing failures. An explicit error makes misconfigurations immediately visible.

### 1. Update `test_get_env_for_task` — unknown task should raise, not fallback

```python
def test_get_env_for_task(self, mock_openai_client):
    # ... (same setup as current) ...
    env_group = EnvGroup(envs=[env1, env2], env_names=["math", "code"])

    assert env_group.get_env_for_task("math") == env1
    assert env_group.get_env_for_task("code") == env2
    # Unknown task should raise, not silently fallback
    with pytest.raises(ValueError, match="No environment found for task"):
        env_group.get_env_for_task("unknown")
```

### 2. Add `test_nested_env_group_preserves_inner_tasks`

```python
def test_nested_env_group_preserves_inner_tasks(self, mock_openai_client):
    """Test that wrapping an EnvGroup in another EnvGroup preserves inner task names."""
    env1 = SingleTurnEnv(
        client=mock_openai_client,
        model="test-model",
        dataset=Dataset.from_dict({"question": ["q1"], "answer": ["a1"]}),
        rubric=Rubric(),
    )
    env2 = SingleTurnEnv(
        client=mock_openai_client,
        model="test-model",
        dataset=Dataset.from_dict({"question": ["q2"], "answer": ["a2"]}),
        rubric=Rubric(),
    )

    inner_group = EnvGroup(envs=[env1, env2], env_names=["math", "code"])
    outer_group = EnvGroup(envs=[inner_group], env_names=["my_env"])

    # Inner task names should be preserved in the dataset
    dataset = outer_group.get_dataset()
    tasks = dataset["task"]
    assert "math" in tasks
    assert "code" in tasks
    assert "my_env" not in tasks

    # Routing should work through both levels
    assert outer_group.get_env_for_task("math") == inner_group
    assert outer_group.get_env_for_task("code") == inner_group
```

### 3. Add `test_nested_env_group_rubric_scoring`

```python
@pytest.mark.asyncio
async def test_nested_env_group_rubric_scoring(self, mock_openai_client, make_input):
    """Test that scoring routes correctly through nested EnvGroups."""
    def math_reward(completion, **kwargs):
        return 0.8

    def code_reward(completion, **kwargs):
        return 0.6

    env1 = SingleTurnEnv(
        client=mock_openai_client,
        model="test-model",
        dataset=Dataset.from_dict({"question": ["q1"], "answer": ["a1"]}),
        rubric=Rubric(funcs=[math_reward], weights=[1.0]),
    )
    env2 = SingleTurnEnv(
        client=mock_openai_client,
        model="test-model",
        dataset=Dataset.from_dict({"question": ["q2"], "answer": ["a2"]}),
        rubric=Rubric(funcs=[code_reward], weights=[1.0]),
    )

    inner_group = EnvGroup(envs=[env1, env2], env_names=["math", "code"])
    outer_group = EnvGroup(envs=[inner_group], env_names=["my_env"])

    # Score a "code" task — should route through outer -> inner -> env2's rubric
    state = State(input=make_input(prompt="Test", answer="ans", task="code"))
    state["completion"] = "Test completion"
    state["trajectory"] = []
    state["timing"] = {"generation_ms": 0.0, "scoring_ms": 0.0, "total_ms": 0.0, "start_time": 0.0}
    state["is_completed"] = False
    state["stop_condition"] = None
    state["oai_tools"] = []
    state["reward"] = None
    state["metrics"] = None

    await outer_group.rubric.score_rollout(state)

    assert state["reward"] == 0.6  # code_reward, not math_reward
    assert state["metrics"]["code_reward"] == 0.6
    assert state["metrics"]["math_reward"] == 0.0
```

## Reproduction

Any environment that returns an `EnvGroup` from its `load_environment()` and is registered as a single `[[orchestrator.env]]` in prime-rl will hit this. The inner task names get overwritten, routing breaks silently, and only the first sub-env's rubric ever scores anything.

tashapais:
## Summary

Two related fixes for `EnvGroup` task routing:

**1. Nested EnvGroup silently misroutes all tasks to envs[0]**

When an `EnvGroup` is wrapped inside another `EnvGroup` (as prime-rl does when registering user envs), the outer group was unconditionally overwriting the inner task names with a single outer name. The inner `EnvGroup` then could not find those outer names in its own `env_map` and silently fell back to `envs[0]` for every rollout, and scored everything as 0.0.

Fix: when a sub-env is itself an `EnvGroup`, register each of its task names in the outer `env_map` pointing to the inner group, and leave the inner task column in the dataset unchanged. Routing then flows `outer -> inner group -> correct sub-env` as expected.

**2. Silent fallback masks misconfiguration**

`get_env_for_task` was returning `envs[0]` for any unrecognized task. This made it impossible to notice routing failures. It now raises `ValueError` with the list of available task names.

## Test plan

- [ ] `test_get_env_for_task` updated: unknown task raises `ValueError` instead of returning `envs[0]`
- [ ] `test_nested_env_group_preserves_inner_tasks`: wrapping an EnvGroup preserves inner task names in dataset and env_map
- [ ] All existing `TestEnvGroup` and `TestEnvGroupRubric` tests still pass

Closes #1008

---

> [!NOTE]
> **Medium Risk**
> Changes core rollout routing behavior in `EnvGroup` (now raises on unknown tasks and alters nested task mapping), which could surface previously-hidden misconfigurations and break callers relying on fallback behavior.
> 
> **Overview**
> **EnvGroup task routing is now strict and nesting-safe.** `EnvGroup` preserves inner `task` labels when an `EnvGroup` is wrapped inside another `EnvGroup`, expands the outer `env_map` to include the inner task names, and rejects task-name collisions at construction time.
> 
> **Misconfiguration is no longer silent.** `get_env_for_task` now raises a `ValueError` listing available tasks rather than defaulting to `envs[0]`; tests were updated/added for unknown-task errors, nested routing, and name-collision handling, and docs were updated to describe the new behavior (including `env_names` in the reference).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 50dbd92e1e6314dc9dbb6bd8b32bf9f6e97d445d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/assets/lab/environments/AGENTS.md
+++ b/assets/lab/environments/AGENTS.md
@@ -786,7 +786,25 @@ combined = vf.EnvGroup(
 )
 ```
 
-The group concatenates all sub-environment datasets, tagging each row with a `task` column that routes rollouts to the appropriate environment for generation and scoring. Metrics from all environments are tracked together. 
+The group concatenates all sub-environment datasets, tagging each row with a `task` column that routes rollouts to the appropriate environment for generation and scoring. Metrics from all environments are tracked together.
+
+Passing an unknown task name to `get_env_for_task` raises a `ValueError` listing the available task names, making misconfiguration immediately visible rather than silently misrouting to the first sub-environment.
+
+An `EnvGroup` can itself be used as a sub-environment inside another `EnvGroup`. The outer group automatically inherits the inner group's task names and routes through both levels, so each inner task name maps correctly to its environment:
+
+```python
+inner = vf.EnvGroup(
+    envs=[math_env, code_env],
+    env_names=["math", "code"],
+)
+
+# prime-rl wraps user envs this way; task names are preserved through both levels
+outer = vf.EnvGroup(envs=[inner], env_names=["my_env"])
+# outer.get_env_for_task("math") -> inner -> math_env
+# outer.get_env_for_task("code") -> inner -> code_env
+```
+
+Task names must be unique across all levels of nesting; a collision raises `ValueError` at construction time.
 
 ## Performance
 
--- a/docs/environments.md
+++ b/docs/environments.md
@@ -780,7 +780,25 @@ combined = vf.EnvGroup(
 )
 ```
 
-The group concatenates all sub-environment datasets, tagging each row with a `task` column that routes rollouts to the appropriate environment for generation and scoring. Metrics from all environments are tracked together. 
+The group concatenates all sub-environment datasets, tagging each row with a `task` column that routes rollouts to the appropriate environment for generation and scoring. Metrics from all environments are tracked together.
+
+Passing an unknown task name to `get_env_for_task` raises a `ValueError` listing the available task names, making misconfiguration immediately visible rather than silently misrouting to the first sub-environment.
+
+An `EnvGroup` can itself be used as a sub-environment inside another `EnvGroup`. The outer group automatically inherits the inner group's task names and routes through both levels, so each inner task name maps correctly to its environment:
+
+```python
+inner = vf.EnvGroup(
+    envs=[math_env, code_env],
+    env_names=["math", "code"],
+)
+
+# prime-rl wraps user envs this way; task names are preserved through both levels
+outer = vf.EnvGroup(envs=[inner], env_names=["my_env"])
+# outer.get_env_for_task("math") -> inner -> math_env
+# outer.get_env_for_task("code") -> inner -> code_env
+```
+
+Task names must be unique across all levels of nesting; a collision raises `ValueError` at construction time.
 
 ## Performance
 
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -471,11 +471,15 @@ OpenEnv integration that runs OpenEnv projects in Prime Sandboxes using a prebui
 ```python
 env_group = vf.EnvGroup(
     envs=[env1, env2, env3],
-    names=["math", "code", "qa"]  # optional
+    env_names=["math", "code", "qa"]  # optional
 )
 ```
 
-Combines multiple environments for mixed-task training.
+Combines multiple environments for mixed-task training. Each row in the concatenated dataset is tagged with a `task` column that routes rollouts and scoring to the correct sub-environment.
+
+`get_env_for_task(task)` raises `ValueError` for unknown task names (listing available tasks) rather than silently falling back to `envs[0]`.
+
+An `EnvGroup` can be nested inside another `EnvGroup`. The outer group automatically inherits the inner group's task names so routing works through both levels. Task names must be unique across all nesting levels; a collision raises `ValueError` at construction time.
 
 ---
 
--- a/environments/AGENTS.md
+++ b/environments/AGENTS.md
@@ -786,7 +786,25 @@ combined = vf.EnvGroup(
 )
 ```
 
-The group concatenates all sub-environment datasets, tagging each row with a `task` column that routes rollouts to the appropriate environment for generation and scoring. Metrics from all environments are tracked together. 
+The group concatenates all sub-environment datasets, tagging each row with a `task` column that routes rollouts to the appropriate environment for generation and scoring. Metrics from all environments are tracked together.
+
+Passing an unknown task name to `get_env_for_task` raises a `ValueError` listing the available task names, making misconfiguration immediately visible rather than silently misrouting to the first sub-environment.
+
+An `EnvGroup` can itself be used as a sub-environment inside another `EnvGroup`. The outer group automatically inherits the inner group's task names and routes through both levels, so each inner task name maps correctly to its environment:
+
+```python
+inner = vf.EnvGroup(
+    envs=[math_env, code_env],
+    env_names=["math", "code"],
+)
+
+# prime-rl wraps user envs this way; task names are preserved through both levels
+outer = vf.EnvGroup(envs=[inner], env_names=["my_env"])
+# outer.get_env_for_task("math") -> inner -> math_env
+# outer.get_env_for_task("code") -> inner -> code_env
+```
+
+Task names must be unique across all levels of nesting; a collision raises `ValueError` at construction time.
 
 ## Performance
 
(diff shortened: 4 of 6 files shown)


tashapais:
Fixes #1008. Two changes: (1) when a sub-env is itself an EnvGroup, its inner task names are now registered in the outer env_map (pointing to the inner group) rather than being overwritten with the outer name -- routing then flows correctly through both levels; (2) get_env_for_task raises ValueError on unknown tasks instead of silently falling back to envs[0].

tashapais:
Addressed all three Bugbot findings in the latest commit (cebd14c):

1. **Name collision (medium):** inner task names from a nested EnvGroup now raise `ValueError` at construction if they conflict with a sibling env's name, instead of silently overwriting it.

2. **Stale outer name (low):** the outer name (e.g. `"my_env"`) is now removed from `env_map` after its inner task names are expanded, so `get_env_for_task("my_env")` raises cleanly rather than returning the inner group which would then immediately raise for the same name.

3. **Docs (low):** `docs/environments.md` updated to cover the `ValueError` on unknown tasks, nested `EnvGroup` support with a code example, and the name-uniqueness requirement.

@willccbb @mikasenghaas happy to make any adjustments.

tashapais:
Addressed the new bugbot findings in commit 50dbd92:

1. **Unconditional pop (medium):** `pop(name)` is now conditional -- it only removes the outer name if that name is not itself one of the inner task names, preventing it from undoing a registration just made.

2. **Duplicated logic (low):** Extracted `_register_nested_env_map` helper so the collision check, registration, and conditional pop live in one place. Both the train and eval dataset branches now call the helper.

3. **reference.md not updated (low):** `docs/reference.md` EnvGroup entry now documents the `ValueError` on unknown tasks, nested `EnvGroup` support, and the name-uniqueness requirement.

mikasenghaas (commented):
nice, i quite like this. we don't need this for prime-rl anymore since we moved away from `EnvGroup` for multi-env training but I still think supporting arbitrary nesting of `EnvGroup` might be nice. thoughts? @willccbb

PR #1218, opened 2026-04-21
AgentError: rollout_id, sandbox_id, ...

snimu:
Adds `sandbox_id`, `rollout_id`, `example_id`, and if present `instance_id` to `AgentError`

### DIFF
--- a/verifiers/envs/experimental/cli_agent_env.py
+++ b/verifiers/envs/experimental/cli_agent_env.py
@@ -148,19 +148,16 @@ class CliAgentEnv(SandboxMixin, vf.MultiTurnEnv):
 
     def make_agent_error(self, state: State, message: str) -> AgentError:
         """Create an AgentError with rollout-specific sandbox context when available."""
-        context_parts: list[str] = []
-        sandbox_id = state.get("sandbox_id")
-        if sandbox_id:
-            context_parts.append(f"sandbox_id={sandbox_id}")
-        state_input = state.get("input")
-        instance_id = (
-            state_input.get("instance_id") if isinstance(state_input, dict) else None
-        )
+        context_parts = [
+            f"sandbox_id={state['sandbox_id']}",
+            f"rollout_id={state['rollout_id']}",
+            f"example_id={state['example_id']}",
+        ]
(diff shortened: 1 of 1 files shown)


verifiers/envs/experimental/cli_agent_env.py
```diff
@@ -146,6 +146,19 @@ def __init__(
...
 
+    def make_agent_error(self, state: State, message: str) -> AgentError:
+        """Create an AgentError with rollout-specific sandbox context when available."""
+        context_parts = [
+            f"sandbox_id={state['sandbox_id']}",
+            f"rollout_id={state['rollout_id']}",
+            f"example_id={state['example_id']}",
+        ]
+        state_info = state["input"].get("info", {})
+        instance_id = state_info.get("instance_id")
```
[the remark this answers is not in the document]
snimu: `Environment.init_state()` should normalize this iiuc.

verifiers/envs/experimental/cli_agent_env.py
```diff
@@ -146,6 +146,19 @@ def __init__(
 
     TUNNEL_CHECK_INTERVAL = 60.0  # seconds between server-side liveness checks
 
+    def make_agent_error(self, state: State, message: str) -> AgentError:
```
rasdani: this func doesn't access `self`, so can be stand-alone utility
either at top-level or in some utils file.

verifiers/envs/experimental/cli_agent_env.py
rasdani: admittedly `instance_id` is leaking a bit swe task implementation detail into the env here.
especially since it might not always be present.

might coerce it in to `example_id` or smth, but this left to be done in swe tasksets. so can keep this for now.

rasdani (approved):
one nit, then lgtm.

verifiers/envs/experimental/cli_agent_env.py
snimu: I'll leave the decision on instance_id to you; i can remove it or leave it, whichever you think makes more sense

PR #1245, opened 2026-04-25
swe tasksets: add 5min sandbox scoring buffer

mikasenghaas:
## Summary

- Sandbox TTL was set to `ceil(timeout_seconds / 60)` in `rlm_swe.py` and other env wrappers, tying it exactly to the rollout-level deadline.
- When both fire simultaneously the scoring rubric runs against a dead container and its sandbox-poll loop hangs indefinitely.
- Adds `SCORING_BUFFER_MINUTES = 5` in `swe_tasksets.py` and applies it at every `SandboxSpec` construction across the 7 SWE backends (`swe_bench`, `r2e_gym`, `multi_swe`, `openswe`, `swe_lego`, `swe_rebench_v2`, `swe_smith`).

## Motivation

Hit a wedge on the GLM-5.1 RLM run (5968): step-75 eval reached 96-98% then stalled for 1h+. Per-worker post-mortem showed the env workers all stuck after a `swe_bench` rubric WARNING:

```
Test execution failed: Sandbox <id> failed (TIMEOUT): Sandbox exceeded maximum runtime
Test execution failed: Read file timed out after 30s: /tmp/job_<id>.exit
```

The sandbox TTL fires at the same instant as the rollout deadline (`2700s` rollout = `ceil(2700/60) = 45min` sandbox), so when the rollout hits its cap and the rubric immediately calls into the sandbox to run `test_patch`, the sandbox is already dying. The rubric's poll then never returns and the worker holds up the eval queue.

A 5-minute headroom guarantees the rubric has a live container to score against. A separate fix (bounded retry in the rubric polling loop) would be a defense-in-depth follow-up.

### DIFF
--- a/verifiers/envs/experimental/composable/tasksets/swe/multi_swe.py
+++ b/verifiers/envs/experimental/composable/tasksets/swe/multi_swe.py
@@ -8,6 +8,7 @@ from typing import Any
 
 import verifiers as vf
 from verifiers.envs.experimental.composable import SandboxSpec, SandboxTaskSet
+from verifiers.envs.experimental.composable.tasksets.swe.swe_tasksets import SCORING_BUFFER_MINUTES
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


mikasenghaas:
Closing — moving this change to research-environments instead, where the SWE rlm-* environments live.

PR #1279, opened 2026-05-03
Make `renderers` optional and add PyPI publish workflow

willccbb:
### Motivation
- Avoid installing the local `packages/renderers` package by default so consumers only opt in when they need rendering support.
- Provide a standard, automated PyPI release flow for the `renderers` package so it can be published independently from the root package.

### Description
- Remove `"renderers"` from the root `dependencies` list in `pyproject.toml` and add an optional extra `renderers = ["renderers"]` under `[project.optional-dependencies]` so users can `pip install verifiers[renderers]` to opt in.
- Keep the local editable source reference in `tool.uv.sources` pointing at `packages/renderers` so the package remains available for local development.
- Add a new GitHub Actions workflow `.github/workflows/publish-renderers.yml` that validates `renderers-v*` tags, checks the tag against `packages/renderers/pyproject.toml`, builds with `uv build packages/renderers`, and publishes with `uv publish --token "$PYPI_TOKEN" packages/renderers/dist/*`.

### Testing
- Ran `uv run pre-commit install` and the install completed successfully.
- Verified `pyproject.toml` can be parsed with `python - <<'PY' import tomllib; tomllib.load(open('pyproject.toml','rb'))` which returned successfully.
- Ran `uv run pre-commit run --all-files` and all configured checks passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f7079af69883269eafc193886da11b)

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes dependency packaging/import behavior and adds a new automated PyPI release workflow; issues would mainly surface as install/import failures or mis-published renderer releases.
> 
> **Overview**
> Adds a new GitHub Actions workflow to publish the `renderers` package to PyPI on `renderers-v*` tags (or manual dispatch), including a tag-to-`packages/renderers/pyproject.toml` version check and `uv` build/publish steps.
> 
> Makes `renderers` an opt-in dependency: removes it from core install, introduces the `renderers` extra (`verifiers[renderers]`), and switches `RendererClient` to lazy-import with clearer install guidance/errors when the extra isn’t present. Docs are updated to instruct installing the extra before using `client_type="renderer"`.
> 
> Bumps `packages/renderers` to `0.1.5`, adds runtime deps (`openai`, `tiktoken`, `jinja2`, `numpy`) and dev deps (`pytest`, `pytest-asyncio`), and updates lockfiles to reflect publishing/consuming `renderers` from PyPI rather than a local editable source.
> 
[...]

PR #1282, opened 2026-05-03
feat(renderers): switch client to vLLM 0.20 /inference/v1/generate

hallerite:
## Summary

vLLM 0.20 ships a unified tokens-in / tokens-out endpoint at `/inference/v1/generate` that supersedes the bespoke `/v1/generate` handler prime-rl shipped on top of vllm 0.19. Migrate verifiers' `RendererClient` onto the new endpoint and pin the renderers package to its lean rewrite.

Companion PRs:
- **PrimeIntellect-ai/renderers#1** — lean `generate()` rewrite (formerly this PR's `packages/renderers/`; now lives in its own repo)
- **PrimeIntellect-ai/prime-rl#2408** — consumer migration onto the new endpoint

### Renderers pin (`pyproject.toml` + `uv.lock`)

- `[tool.uv.sources]` pins `renderers` to `PrimeIntellect-ai/renderers@40bc2a6` (the head of the companion PR — the lean `generate()` rewrite).
- `packages/renderers/` deleted from the verifiers tree; the package is no longer vendored.
- Version constraint bumped to `renderers>=0.1.6` in `[project]` deps and the `renderers` extra. Once `renderers-v0.1.6` publishes to PyPI, drop `[tool.uv.sources]` and let it resolve from PyPI directly.
- Drops the `uv pip install -e packages/renderers` CI hack — no longer needed once renderers resolves through `[tool.uv.sources]`.
- Deletes `.github/workflows/publish-renderers.yml` — the publish flow now lives in the renderers repo (PrimeIntellect-ai/renderers#2).

### `RendererClient` adapter (`verifiers/clients/renderer_client.py`)

- `get_native_response` builds a `sampling_params` dict from the caller's flat sampling_args / `extra_body` and calls the new `generate(...)` with named args. This is the right place for the OpenAI-shaped → lean adaptation; the renderers package itself no longer carries OpenAI-SDK conventions.
- `from_native_response` reads `request_id` instead of `id`; `Usage` is reconstructed from token-list lengths (the new endpoint doesn't return a usage block).

## Test plan

- [x] `tests/test_renderer_client.py` + `tests/test_renderer_e2e.py` updated for the new `request_id` / `sampling_params` shapes — 42/42 pass against the external pin.
- [x] e2e renderer rollout against a live vllm 0.20 server (prime-rl#2408 + this PR's client): 20-step `multi_reverse_text` RL run, 2688 calls to `/inference/v1/generate`, eval Avg@4 = 0.83 — identical numbers to the fat-API version.
- [x] `ruff format --check` / `ruff check` clean.
- [x] `ty check verifiers` passes (0 errors).
- [x] Rebased onto current `main` (picks up #1279 — renderers as optional / PyPI publish workflow).

## Notes

- VLMs are still blocked by prime-rl's `validate_renderer_vs_vlm` config validator. The new endpoint already supports MM features end-to-end; lifting the ban needs the renderer client to build features client-side (HF processor → `MultiModalKwargsItem` → base64 msgpack). That's a separate PR — easier to review on its own once this lands.

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes the inference request/response contract (`generate` call shape, `request_id`, and usage reconstruction) and removes the vendored `packages/renderers` implementation in favor of a pinned external dependency.
> 
> **Overview**
> Switches `RendererClient` over to `renderers.client.generate`, adapting OpenAI-style sampling args into a `sampling_params` dict, passing through `cache_salt`/`priority`/headers, and updating response handling to use `request_id` and reconstruct `usage` from token lengths.
> 
> Removes the in-repo `packages/renderers` implementation (and its `publish-renderers` GitHub workflow) and pins `renderers` via `pyproject.toml` (`renderers>=0.1.6` plus a git `tool.uv.sources` override) so the renderer code is consumed as an external package.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b494fb75e94acf58da6846a2b312fba3e9871a2e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


verifiers/clients/renderer_client.py
eligotts, line 114: unrelated to this pr but realized we have drifted from our current token client when dealing with content as None vs "", see [here](https://github.com/PrimeIntellect-ai/verifiers/blob/877e58296aa00214361460a8760d8a0d942d729e/verifiers/clients/openai_chat_completions_token_client.py#L176).
we should be robust to tool call parser giving us back either empty string or none in content on tool calls, that should still count as message level prefix hit

PR #1346, opened 2026-05-11
feat(renderer-client): thread multimodal sidecar through rollout + transport

hallerite:
## Summary

Threads the renderer's `MultiModalData` sidecar (pixel_values, placeholder ranges, mm_hashes) from the renderer through `/inference/v1/generate` to vLLM and onto trajectory tokens — so the new renderer-only multimodal path in prime-rl can drop the orchestrator-side `AutoProcessor` and image cache entirely.

Companion PRs:
- renderers: [PrimeIntellect-ai/renderers#17](https://github.com/PrimeIntellect-ai/renderers/pull/17)
- prime-rl: [PrimeIntellect-ai/prime-rl#2473](https://github.com/PrimeIntellect-ai/prime-rl/pull/2473)

## What changes

**Rollout (`renderer_client.py`)**
- `_get_incremental_prompt_ids` now returns `RenderedTokens | None` (`token_ids` + `multi_modal_data`); text-only renderers' raw `list[int]` is normalized via `as_rendered_tokens` so callers unpack uniformly
- `_step_multi_modal_data(step)` recovers the prior turn's `mm_data` from the trajectory step and forwards it to `bridge_to_next_turn` so the new turn's placeholder runs cover every earlier-turn image
- `RendererClient.create_completion` unpacks the bridged result into `(prompt_ids, multi_modal_data)` and forwards both to `generate`
- `parse_response_tokens` copies `response.multi_modal_data` onto `ResponseTokens`
- Pool dispatch: `is_multimodal(renderer)` (cached `bool`) replaces the runtime-checkable `isinstance` walk on the hot path; pool methods are called directly (the pool implements the protocol structurally) and offloaded via `asyncio.to_thread` when the renderer is a pool

**Types (`types.py`)**
- `ResponseTokens.multi_modal_data: Any | None`
- `TrajectoryStepTokens.multi_modal_data: NotRequired[Any]`

Typed as `Any` to avoid a hard import dep on `renderers`.

**Transport (`utils/serve_utils.py`)**
- Custom msgpack encoder gains torch tensor / numpy ndarray / dataclass support; tensors encode as `{__torch_tensor__: True, dtype, shape, data}` with raw bytes payload (torch imported lazily — text-only consumers don't pay for it)
- `decode_tensor_payload` / `walk_decode_tensors` rehydrate on the receiving side

**Save (`utils/save_utils.py`)**
- `is_json_serializable` keeps its honest JSON-only contract; `state_to_output` bypasses the gate for the `trajectory` column (msgpack-transported)
- `_strip_intermediate_mm_data` drops `tokens.multi_modal_data` from all but the last trajectory step before transport — bridge merges prior turns' `mm_data` into each new turn, so naively shipping is O(N²) bytes per N-turn rollout

## Test plan

- [ ] Text-only RL: no `mm_data` → fast path unchanged
- [ ] Multimodal RL: `mm_data` reaches vLLM via `multi_modal_data` and the trainer via `TrajectoryStepTokens["multi_modal_data"]`
- [ ] Bridge: previous-turn images carried forward, placeholder count matches the combined token sequence
- [ ] Transport: msgpack round-trip preserves tensor shapes / dtypes
- [ ] JSONL save: trajectories with `multi_modal_data` accepted, no spurious "not JSON-serializable"; no O(N²) duplication

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes renderer client request/response plumbing and serialization/transport of trajectory data (msgpack), which could affect rollout tokenization compatibility and saved state size/format for multimodal runs.
> 
> **Overview**
> Adds end-to-end support for renderer-emitted multimodal sidecar data (`MultiModalData`) by threading it through `RendererClient` bridging, `/inference/v1/generate` calls, response parsing, and into `TrajectoryStepTokens`/docs.
> 
> Updates trajectory saving/transport to handle non-JSON multimodal payloads: msgpack encoding now supports numpy arrays/torch tensors/dataclasses with decode helpers, and `trajectory` output is delta-encoded per step to avoid O(N²) duplication of cumulative multimodal items (with new tests covering compaction and multiset hash semantics).
> 
> Bumps `renderers` dependency to `>=0.1.8.dev0` (switching from a pinned git rev) and adjusts renderer-client tests to the new `RenderedTokens` return shape and renderer id naming.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb56caa0cc0df373f9a0c7111f11f1ed89a787d8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


verifiers/clients/renderer_client.py
```diff
@@ -94,15 +97,15 @@ def _record_bridge(success: bool) -> None:
...
+    """Run sync renderer work on a thread iff ``renderer`` is a pool.
 
-        return await asyncio.to_thread(_work)
-    return fn(renderer)
+    Pool methods can block on the internal queue/lock; we offload to keep
+    the event loop responsive. A bare ``Renderer`` runs inline.
+    """
+    if isinstance(renderer, RendererPool):
+        return await asyncio.to_thread(fn)
+    return fn()
```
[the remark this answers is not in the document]
hallerite: This is incorrect on the runtime claim. `RendererPool` structurally implements the `Renderer` (and `MultimodalRenderer`) protocol — see `renderers/base.py:454-474`. `bridge_to_next_turn`, `render`, `render_ids`, `parse_response`, and `get_stop_token_ids` are all proxied via:

```python
def bridge_to_next_turn(self, *args, **kwargs):
    with self.checkout() as r:
        return r.bridge_to_next_turn(*args, **kwargs)
```

`checkout()` (`base.py:426-437`) provides the locking:
- `size == 1`: `with self._lock:` (`threading.Lock`)
- `size > 1`: `queue.Queue.get()` / `put()` (blocking, thread-safe — at most one caller holds a given slot)

So `pool.bridge_to_next_turn(...)` does not `AttributeError`, and concurrent calls cannot conflict on the same underlying renderer instance. The whole point of the refactor was to make the pool a drop-in `Renderer` for callers, so `_maybe_offload` no longer needs to reach into `checkout()` itself — that responsibility moved into the pool's delegating methods.

`is_multimodal(pool)` is also fine: it reads `pool.is_multimodal` (cached `bool` set in `__init__`), no protocol walk on the hot path.

pyproject.toml
```diff
@@ -126,10 +126,12 @@ prime-sandboxes = false
...
-# Pinned to renderers main until the next PyPI release lands; drop after.
-# fe67f9f = renderers main: PR #4 squash-merge — construction-time
-# preserve_*_thinking flags on create_renderer / create_renderer_pool.
-renderers = { git = "https://github.com/PrimeIntellect-ai/renderers.git", rev = "fe67f9f" }
+# Pinned to renderers feat/multimodal-vlm until that PR (#17) lands and
+# a new PyPI release is published. Provides RenderedTokens,
+# MultiModalData, PlaceholderRange, as_rendered_tokens, and the
+# `previous_multi_modal_data` kwarg on Renderer.bridge_to_next_turn
+# that this branch consumes. Drop after merge + release.
+renderers = { git = "https://github.com/PrimeIntellect-ai/renderers.git", branch = "feat/multimodal-vlm" }
```
willccbb: Merged the renderers PR, can we release renderers + replace this? Should never have a git source pinned in verifiers, PyPI won't respect it + we always want vf to be releasable

willccbb (changes requested):

verifiers/utils/save_utils.py
```diff
@@ -263,6 +272,26 @@ def state_to_output(
...
+            # Renderer multimodal rollouts accumulate mm_data on every step
+            # (bridge_to_next_turn merges previous_multi_modal_data into the
+            # new turn). Each subsequent step's mm_data is a superset of the
+            # prior step's, so shipping mm_data on every step duplicates
+            # every image O(N²) bytes for an N-turn rollout. Only the last
+            # step's sidecar is read by the trainer — strip mm_data from
+            # earlier steps before transport. Bridge accesses mm_data only
+            # within the env-worker rollout loop, which has already
+            # finished by the time state_to_output runs.
+            value = _strip_intermediate_mm_data(value)
```
eligotts: this assumes pure extension right? how do we handle branching/compaction rollouts?

pyproject.toml
hallerite: Done — released [renderers v0.1.7](https://pypi.org/project/renderers/0.1.7/) to PyPI and dropped the git source in caf4f6cc. `uv.lock` now resolves `renderers` from `registry = "https://pypi.org/simple"`.

verifiers/utils/save_utils.py
hallerite: great catch. I think 9aafa65 fixes this

willccbb (approved):

pyproject.toml
```diff
@@ -74,7 +74,7 @@ dev = [
     "aiohttp>=3.9.0",
     "python-dotenv>=1.0.0",
     "nltk",
-    "renderers>=0.1.6",
+    "renderers>=0.1.8.dev0",
```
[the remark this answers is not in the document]
willccbb: @cursoragent PyPI can still resolve, right? Both repos are maintained by our org, we trust the dev0 as stable for our purposes + accept the maintenance burdens, the tag is just flagging potential instability for end users working with it directly.

PR #1369, opened 2026-05-14
[codex] Add wiki-search typed message regression test

tim0120:
## Summary

Adds a regression test for the wiki-search judge reward path with typed verifier messages, including an assistant message whose content is `None`.

The published `wiki-search` environment package can fail this path by assuming completions are dicts with string content, which silently zeroes rewards when `Rubric` catches the reward exception. Current main already passes by handing the original completion to `JudgeRubric`; this test guards that behavior.

## Validation

- `uv run pytest tests/test_wiki_search_env.py` (skips when optional wiki-search deps are absent)
- `uv run --with chromadb --with datasets pytest tests/test_wiki_search_env.py`

### DIFF
--- a/tests/test_wiki_search_env.py
+++ b/tests/test_wiki_search_env.py
@@ -1,0 +1,49 @@
+import importlib.util
+from pathlib import Path
+
+import pytest
+
+import verifiers as vf
+
+pytest.importorskip("chromadb")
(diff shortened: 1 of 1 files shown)


tim0120:
Closing this PR. The related experiment changes should stay local for now.

PR #1427, opened 2026-05-20
Serialize retryable rollout error policy

rasdani:
## Summary
- Add optional `is_retryable` to serialized `ErrorInfo` payloads.
- Populate it from verifiers' default retry semantics: retry when the causal error chain contains `InfraError` or `InvalidModelResponseError`, including subclasses such as sandbox/tunnel/browser sandbox errors and empty model responses.
- Preserve `is_retryable` when reserializing existing `ErrorInfo` mappings through `state_to_output`.
- Update `maybe_retry` to use `is_retryable_error(error: BaseException | ErrorInfo, ...)` for both live exceptions and serialized `ErrorInfo`.
- Remove the old serialized-error string reconstruction path. Serialized errors now retry only when `is_retryable` is `true`.

## Coordinated Consumer
This is the producer-side change for PrimeIntellect-ai/prime-rl#2579. prime-rl receives rollout failures after verifiers serializes them into `ErrorInfo` dictionaries, so it cannot safely recover Python subclass relationships from error-name strings.

With this field, verifiers remains the source of truth for retry semantics and prime-rl can consume the serialized retry decision directly. prime-rl intentionally does not enumerate subclass names or fall back to string parsing; missing `is_retryable=true` is treated as terminal/no-reschedule.

## Why
Class-name strings preserve logging detail, but they do not preserve Python subclass relationships like `SandboxError <: InfraError`. Serializing the retryability decision keeps downstream consumers aligned with verifiers `maybe_retry` semantics and prevents plain model/provider errors from being mistaken for retryable infra failures.

Existing serialized `ErrorInfo` records remain valid because `is_retryable` is optional. The compatibility behavior is intentionally conservative: older serialized errors without the flag do not retry across the serialized boundary.

## Test Plan
- `uv run pytest tests/test_error_chain.py`
- `uv run ty check verifiers/utils/error_utils.py verifiers/utils/async_utils.py`
- `uv run ruff check verifiers/utils/error_utils.py verifiers/utils/async_utils.py tests/test_error_chain.py`
- `uv run ruff format --check verifiers/utils/error_utils.py verifiers/utils/async_utils.py tests/test_error_chain.py`

---

> [!NOTE]
> **Medium Risk**
> Changes retry behavior across the serialized-error boundary: previously some serialized errors could retry via class-name string parsing; now retries for serialized errors only happen when `ErrorInfo.is_retryable` is set, which could alter rollout scheduling/resilience if producers/consumers are mismatched.
> 
> **Overview**
> Adds an optional `is_retryable` flag to serialized `ErrorInfo` and populates it using verifiers’ default retry policy (error chain contains `InfraError` or `InvalidModelResponseError`).
> 
> Updates `maybe_retry` to use a new `is_retryable_error()` helper for both live exceptions and serialized `ErrorInfo`, and removes the prior string-based reconstruction path for serialized errors (missing `is_retryable=True` is treated as non-retryable). `state_to_output` now preserves `is_retryable` when re-serializing existing `ErrorInfo` mappings and otherwise emits errors via the shared `error_info()` serializer.
> 
> Extends tests to cover default retryability marking and the stricter behavior for serialized errors without the flag.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e710e01e62f4d7d62780bfa677564437a0c38cfd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_error_chain.py
+++ b/tests/test_error_chain.py
@@ -4,9 +4,9 @@ import verifiers as vf
 from verifiers.utils.error_utils import (
     ErrorChain,
     error_info,
-    error_info_to_exception,
     get_vf_error_chain,
     is_retryable_error,
+    is_retryable_error_info,
 )
 
 
@@ -183,9 +183,7 @@ class TestErrorChain:
         assert info["is_retryable"] is False
         assert is_retryable_error(outer) is False
 
-    def test_error_info_to_exception_uses_retryable_flag_for_serialized_subclasses(
-        self,
-    ):
+    def test_is_retryable_error_info_uses_serialized_retryable_flag(self):
         """Serialized subclass names do not need to enumerate their base classes."""
         info = {
             "error": "SandboxError",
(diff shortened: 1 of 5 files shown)


rasdani:
@codex review

PR #1742, opened 2026-06-18
fix: cap concurrent CPU-bound scoring scripts on the subprocess runtime

mikasenghaas:
## Summary

At an eval's start the whole in-flight batch finishes generation at roughly the same time and fires
its per-rollout scoring at once. On the **subprocess** runtime each math/gsm8k/aime score is a fresh
process that cold-imports `sympy` (`math-verify`) — CPU- and memory-heavy. N≫cores of them launching
together oversubscribe the cores and stack memory; the per-rollout scoring wall-clock — which the 30s
`scoring_timeout` bounds — balloons past the timeout for the unlucky rollouts. Once the herd clears,
scoring drops back to ~normal.

This adds a `cpu_bound` flag to the runtime exec API and uses it to cap how many such scripts run at
once on the host:

- **`Runtime.run` / `run_program` / `run_uv_script`** gain `cpu_bound: bool = False`.
- The **subprocess runtime** caps concurrently-running `cpu_bound` subprocesses at the available core
  count — a process-wide `asyncio.Semaphore` sized to `len(os.sched_getaffinity(0))`, override with
  `VF_SUBPROCESS_CPU_SLOTS`. The herd then drains through the cores instead of thrashing all at once,
  and peak concurrent verifiers (≈ peak memory) is bounded.
- The long-lived, **I/O-bound harness program** does *not* set the flag, so rollout concurrency is
  untouched — only the short CPU-bound scoring scripts are throttled.
- **Isolated runtimes** (docker / modal / prime) accept the flag and ignore it — their scripts run in
  a sandbox, off the host.
- Tags the three sympy-based verifiers `cpu_bound=True`: **gsm8k-v1**, **aime24-v1**, **math_env-v1**.

## Verification

Repro: fire N concurrent gsm8k verifiers (each cold-imports sympy) on the subprocess runtime,
measuring per-call wall-clock (what the scoring timeout sees), `cpu_bound=False` vs `True`.

On an idle 30-core / 283 GB box (no memory pressure, fast CPU — so it can't hit the 30s wall, but the
oversubscription effect is clear):

```
  n=256  cpu_bound=False  p50= 2.5s  p90= 3.4s  max= 3.5s
  n=256  cpu_bound=True   p50= 1.8s  p90= 3.1s  max= 3.4s
  n=512  cpu_bound=False  p50= 4.8s  p90= 7.6s  max= 7.8s
  n=512  cpu_bound=True   p50= 3.2s  p90= 5.5s  max= 6.1s
```

Pinned to 8 cores (simulating the eval sharing a node with vLLM + the trainer) the uncapped case
lands right on the 30s cliff — the user's exact symptom — and the cap roughly halves it:

```
  n=256  cpu_bound=False  p50= 9.7s  p90=13.6s  max=13.8s
  n=256  cpu_bound=True   p50= 4.2s  p90= 7.5s  max= 8.4s
  n=512  cpu_bound=False  p50=17.4s  p90=28.7s  max=29.3s   <- right at the 30s scoring_timeout
  n=512  cpu_bound=True   p50= 8.1s  p90=14.5s  max=16.2s
```

The cap reaches the semaphore through the **production default path** too
(`RetryingRuntime(SubprocessRuntime)` → `run_uv_script` → `run_program` → `inner.run`): with the cap
pinned to 1, four concurrent `cpu_bound=True` calls serialize (~4×) while `cpu_bound=False` calls run
in parallel.

## Breaking

- `Runtime.run` / `run_program` / `run_uv_script` gain a trailing `cpu_bound: bool = False` parameter
  (additive, defaulted). A **custom `Runtime` subclass** that overrides `run` must add the parameter to
  match the ABC — the base `run_uv_script` now calls `self.run(..., cpu_bound)`. The four built-in
  runtimes are updated.

> [!NOTE]
> ### Cap concurrent CPU-bound scoring scripts on the subprocess runtime
> - Adds a process-wide `asyncio.Semaphore` to [`SubprocessRuntime`](https://github.com/PrimeIntellect-ai/verifiers/pull/1742/files#diff-f36f9994618d7a28baaf498b61eb6f129dcfda77500d165d18537426dfcdd6e2) that limits concurrent CPU-bound subprocesses to the number of CPU cores (or `VF_SUBPROCESS_CPU_SLOTS` env var override).
> - Extends the `cpu_bound: bool = False` flag through the full runtime hierarchy: `Runtime.run`, `run_program`, `run_uv_script`, and all implementations in Docker, Modal, Prime (accepted but unused), and subprocess runtimes.
> - AIME24, GSM8K, and MATH scoring/validation calls to `run_uv_script` now pass `cpu_bound=True` to opt into throttling.
> - Behavioral Change: on the subprocess runtime, concurrent verifier executions are now capped at core count by default; other runtimes are unaffected.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 20f4883.</sup>
> 
>

### DIFF
--- a/environments/aime24_v1/aime24_v1/taskset.py
+++ b/environments/aime24_v1/aime24_v1/taskset.py
@@ -62,6 +62,7 @@ class AIME24Taskset(vf.Taskset[AIME24Task, AIME24Config]):
         result = await runtime.run_uv_script(
             VERIFY,
             args=[task.answer, prediction or "", str(self.config.math_verify_timeout)],
+            cpu_bound=True,
         )
         if result.exit_code != 0:
             raise vf.ProgramError(f"verify.py failed: {result.stderr.strip()[-500:]}")
--- a/environments/gsm8k_v1/gsm8k_v1/taskset.py
+++ b/environments/gsm8k_v1/gsm8k_v1/taskset.py
@@ -51,7 +51,7 @@ class GSM8KTaskset(vf.Taskset[GSM8KTask, GSM8KConfig]):
             trace.assistant_messages[-1].content if trace.assistant_messages else ""
         )
         result = await runtime.run_uv_script(
-            VERIFY, args=[task.answer, prediction or ""]
+            VERIFY, args=[task.answer, prediction or ""], cpu_bound=True
         )
         if result.exit_code != 0:
             raise vf.ProgramError(f"verify.py failed: {result.stderr.strip()[-500:]}")
@@ -63,7 +63,7 @@ class GSM8KTaskset(vf.Taskset[GSM8KTask, GSM8KConfig]):
         answer as a well-formed `#### N` prediction and require a 1.0 score — catching rows the
         verifier can't parse or grade (the model-free counterpart of the `correct` reward)."""
         result = await runtime.run_uv_script(
-            VERIFY, args=[task.answer, f"#### {task.answer}"]
+            VERIFY, args=[task.answer, f"#### {task.answer}"], cpu_bound=True
         )
         if result.exit_code != 0:
             raise vf.ProgramError(f"verify.py failed: {result.stderr.strip()[-500:]}")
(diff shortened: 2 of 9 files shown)


mikasenghaas:
Scrapping this approach — the throttle concern shouldn't surface in env code.

PR #1766, opened 2026-06-19
feat: persistent runtime pool + warm in-runtime scoring

mikasenghaas:
## Summary

Adds a **persistent runtime** mode + **warm in-runtime scoring** so the heavy per-rollout costs (sandbox/container provisioning, and re-`import`ing a verifier's deps every rollout) are paid **once per run** instead of once per rollout.

- **`persistent` on the base runtime config** (`BaseRuntimeConfig.persistent`, inherited by subprocess/docker/prime/modal; CLI `--harness.runtime.persistent true`). A persistent runtime is taken from an eval/train-level **`RuntimePool`**, reused across rollouts, and torn down only at the end of the run.
- **Wired into `Environment.serving()`** alongside the existing shared-tools / interception pools, and injected into every `Rollout` via `episode()` — so it covers **both** the eval CLI and the env server. The env server's `serving()` spans the whole run, so on **training** the pool lives for the entire run.
- **`Rollout`** acquires/releases a pooled runtime instead of `make_runtime`/`stop` (ephemeral path unchanged). Acquire happens inside the rollout's `try`, so a provisioning failure is captured on the trace like a normal `start()` failure.
- **`Runtime.reset()`** clears the per-rollout workspace between reuses (subprocess recreates `/tmp/<name>`; docker/prime/modal empty the workdir) so reuse stays isolated; the provisioned resource **and** warm workers survive. Persistent mode therefore suits tasksets whose per-rollout state is workspace-local (e.g. gsm8k / math).
- **`run_uv_script(..., warm=True)`** — the subprocess runtime routes to a long-lived worker that loads the script as a module **once** (its heavy top-level imports paid once) and answers many `args → stdout` calls. Gated on `persistent` (a worker would otherwise die with its rollout). Scripts opt in by exposing `main(argv) -> str` while staying `uv run`-able cold via a `if __name__ == "__main__": print(main(sys.argv[1:]))` footer.
- **gsm8k-v1 / aime24-v1 / math-env-v1** opt in: each `verify.py` converted to the dual-mode `main(argv)` shape (the math ones return instead of `sys.exit`, so a reused worker survives the early-out paths) and the `correct` reward passes `warm=True`.
- Split the runtime factory (`RuntimeConfig` / `make_runtime` / `runtime_is_local`) into `runtimes/factory.py` so the pool can build runtimes without importing the package (avoids a cycle); `runtimes/__init__` re-exports.

## Why two layers

The provisioning win (skip re-creating a container/sandbox per rollout) needs only the **pool**, and helps remote runtimes most (the tb2 bench measured 2.5–7.4s sandbox+tunnel provisioning per rollout). The *import-once* win needs a **warm worker**, which only survives if the runtime persists — so the pool is the foundation and warm scoring rides on it. For subprocess, provisioning is ~free (`start`/`stop` is 0.2 ms); the entire per-rollout cost is the fresh-process `import math_verify` (~0.22 s), which the warm worker removes.

## Verification

`uv run python bench/persistent_runtime.py <N> <concurrency>` — runs gsm8k's real `verify.py` through the framework API (no model generation, to isolate runtime/scoring overhead). `ephemeral` = `make_runtime + start + run_uv_script(cold) + stop` per call (today's per-rollout scoring path); `persistent` = `pool.acquire + run_uv_script(warm=True) + release`.

| N | concurrency | ephemeral | persistent | speedup |
|---|---|---|---|---|
| 256  | 128 | 3.95s (15.5 ms/call) | 1.21s (4.7 ms/call) | **3.3x** |
| 1000 | 128 | 15.31s (15.3 ms/call) | 1.27s (1.27 ms/call) | **12.0x** |
| 4000 | 128 | 63.28s (15.8 ms/call) | 2.16s (0.54 ms/call) | **29.4x** |

The win grows with N: ephemeral stays pinned at ~15.3 ms/call (the per-call import), while persistent's one-time worker import (one per pooled runtime) amortizes away — so it peaks when `N ≫ concurrency` (a long training run).

Correctness: a small gsm8k-v1 eval with `--harness.runtime.persistent true` vs default produces identical rewards (`reward=1.000`, 0 errors), logs rollouts as `(pooled)`, and tears the pool down at the end (`runtime pool: tearing down N persistent runtime(s)`). Warm worker output verified correct across edge cases (e.g. `1,000`, malformed → `0.0`). `ruff` clean; `tests/v1/test_configs.py` green; full test suite collects.

## Scope / follow-ups

- Warm workers are currently implemented on the **subprocess** runtime. **Persistence** works on every runtime (and on remote runtimes also skips the per-rollout sandbox+tunnel provisioning). Warm workers on docker/prime/modal (a small in-runtime HTTP service reached via the existing `expose`/`host_endpoint` plumbing — structurally like a tool server) are a natural follow-up.
- Persistent mode reuses a runtime across rollouts/tasks; the per-rollout workspace is `reset`, but anything a taskset installs **outside** the workspace persists — so it's opt-in and best for workspace-local state. Tasks with per-task images get a separate pool per image (pool keyed by resolved config).

## Breaking

None. `persistent` defaults to `False` (ephemeral, today's behavior); `warm` defaults to `False`. `run_uv_script` gains an optional trailing `warm` kwarg.

### DIFF
--- a/bench/persistent_runtime.py
+++ b/bench/persistent_runtime.py
@@ -1,0 +1,90 @@
+"""Benchmark the persistent runtime + warm scoring against the per-rollout default.
+
+Isolates the runtime/scoring overhead (no model generation, which would mask it): runs gsm8k's
+`verify.py` N times at a fixed concurrency, both ways, through the real framework API.
+
+  ephemeral:  make_runtime + start() + run_uv_script(cold) + stop() per call
+              — exactly today's per-rollout scoring path.
+  persistent: RuntimePool.acquire() + run_uv_script(warm=True) + release() per call
+              — runtimes (and their warm workers) reused; the math_verify import is paid once
+              per pooled runtime, not per call.
+
+Usage: uv run python bench/persistent_runtime.py [N=1000] [CONCURRENCY=128]
+"""
+
+import asyncio
+import sys
+import time
+from pathlib import Path
+
+from verifiers.v1.runtimes import RuntimePool, SubprocessConfig, make_runtime
+
+VERIFY = (
+    Path(__file__).resolve().parents[1] / "environments/gsm8k_v1/gsm8k_v1/verify.py"
+).read_bytes()
+GOLD, PRED = "42", "#### 42"
+
+
+async def ephemeral(n: int, concurrency: int) -> float:
+    sem = asyncio.Semaphore(concurrency)
+
+    async def one() -> None:
+        async with sem:
+            rt = make_runtime(SubprocessConfig())
+            await rt.start()
+            try:
+                r = await rt.run_uv_script(VERIFY, args=[GOLD, PRED])
+                assert r.stdout.strip().endswith("1.0"), r
+            finally:
+                await rt.stop()
+
+    t0 = time.time()
+    await asyncio.gather(*(one() for _ in range(n)))
+    return time.time() - t0
+
+
+async def persistent(n: int, concurrency: int) -> float:
+    sem = asyncio.Semaphore(concurrency)
+    cfg = SubprocessConfig(persistent=True)
+
+    async def one(pool: RuntimePool) -> None:
+        async with sem:
+            rt = await pool.acquire(cfg)
+            try:
+                r = await rt.run_uv_script(VERIFY, args=[GOLD, PRED], warm=True)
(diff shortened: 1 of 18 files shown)


mikasenghaas:
issue was nfs, not code

PR #1819, opened 2026-06-21
Decode large env replies off the event loop

xeophon:
## Overview

Keep env-server reply bytes compact on the asyncio event loop, then unpack and validate large trace responses in a worker thread. A client-level decode lock limits expansion to one large representation at a time, while small health and info responses retain their direct synchronous path.

## Why

The receiver previously unpacked msgpack before completing each request future, and the request coroutine then performed Pydantic validation on the same event loop. Large queued traces could therefore block unrelated coroutines while several expanded raw object graphs were retained by their futures.

Routing compact bytes through the receiver keeps request matching lightweight. Large replies acquire one decode slot and perform msgpack expansion plus typed validation together in a worker. The slot is released when the worker physically finishes, so caller cancellation cannot permit overlapping giant decodes. Wire format, typed responses, timeout cleanup, and public client methods remain unchanged.

## Performance

The reproduction used three concurrent, distinct 25,934,052-byte replies (77,802,156 bytes total). Each reply contained 1,000 message nodes with 2,000 token IDs, masks, and log probabilities per node. Values below are medians across three separate processes on macOS.

| Measurement | Before | After | Difference |
|---|---:|---:|---:|
| Wall time | 367.596 ms | 372.706 ms | +5.111 ms (+1.39%) |
| Maximum event-loop gap | 367.585 ms | 68.015 ms | −299.571 ms (81.50% saved) |
| Peak RSS increase during decode | 516.719 MiB | 469.000 MiB | −47.719 MiB (9.23% saved) |
| Retained RSS increase after release/GC | 142.969 MiB | 121.078 MiB | −21.891 MiB (15.31% saved) |
| RSS with all typed results retained | 751.734 MiB | 751.969 MiB | effectively unchanged |

The final typed representations necessarily occupy the same memory. The reduction comes from avoiding simultaneous expanded msgpack intermediates. The new path adds one reusable worker thread and preserves a maximum decoder concurrency of one.

---

> [!NOTE]
> **Medium Risk**
> Changes async timing and cancellation behavior for large env-server responses; correctness depends on shielded background decode and semaphore release on all paths, but scope is limited to the ZMQ client.
> 
> **Overview**
> **`EnvClient`** no longer unpacks msgpack in the ZMQ receive loop; pending futures now carry raw reply **`bytes`**, and **`_request`** decides where decoding happens.
> 
> **Health** and **info** responses still **`msgpack.unpackb`** and **`model_validate`** on the event loop. Rollout/trace-style replies acquire a client-wide **`BoundedSemaphore(1)`**, run unpack + Pydantic validation in **`asyncio.to_thread`**, and **`asyncio.shield`** the decode task so caller cancellation cannot overlap large expansions; the slot releases only when the worker finishes.
> 
> Public wire format, timeouts, and client methods are unchanged; under concurrent large replies, decode is serialized (possible extra latency) in exchange for shorter event-loop stalls and lower peak intermediate memory.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7e575251006d888d4d982185b3761b196d400aab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Decode large env replies off the event loop in `EnvClient`
> - `_receive_loop` now sets raw `bytes` on pending futures instead of decoding with `msgpack.unpackb` at receive time, moving decode work out of the hot path.
> - `_request` offloads msgpack unpacking and Pydantic validation for non-health/info responses to a background thread via `asyncio.to_thread`, with concurrency limited to 1 via a `BoundedSemaphore`.
> - The decode task is shielded from cancellation via `asyncio.shield`; if the caller is cancelled, a done callback observes any exception before re-raising to avoid unobserved-exception noise.
> - `HealthResponse` and `InfoResponse` are still decoded inline on the event loop.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7e57525.</sup>
> 
>

### DIFF
--- a/verifiers/v1/serve/client.py
+++ b/verifiers/v1/serve/client.py
@@ -50,8 +50,9 @@ class EnvClient:
         self.socket.setsockopt(zmq.SNDHWM, 0)
         self.socket.setsockopt(zmq.RCVHWM, 0)
         self.socket.connect(address)
-        self._pending: dict[str, asyncio.Future] = {}
+        self._pending: dict[str, asyncio.Future[bytes]] = {}
         self._receiver: asyncio.Task | None = None
+        self._decode_lock = asyncio.Lock()
 
     def _ensure_receiver(self) -> None:
         if self._receiver is None:
@@ -65,7 +66,7 @@ class EnvClient:
                 break
             future = self._pending.pop(request_id_bytes.decode(), None)
             if future is not None and not future.done():
-                future.set_result(msgpack.unpackb(data, raw=False))
+                future.set_result(data)
 
     async def _request(
         self,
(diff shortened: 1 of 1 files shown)


verifiers/v1/serve/client.py
```diff
@@ -77,18 +78,38 @@ async def _request(
...
+            await self._decode_lock.acquire()
+            decoding = asyncio.create_task(
+                asyncio.to_thread(
+                    lambda: response_type.model_validate(
+                        msgpack.unpackb(data, raw=False)
+                    )
+                )
+            )
+            # The worker owns the slot so caller cancellation cannot overlap decodes.
+            decoding.add_done_callback(lambda _: self._decode_lock.release())
```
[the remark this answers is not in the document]
xeophon: Addressed in 7e5752510. `asyncio.Lock` is not task-owned, but this now uses `BoundedSemaphore(1)` to model the decode capacity slot explicitly. The completion callback releases the slot only after the background decode finishes, including when the caller is cancelled.

verifiers/v1/serve/client.py
```diff
@@ -77,18 +78,38 @@ async def _request(
...
         await self.socket.send_multipart(
             [request_id.encode(), request.method.encode(), payload]
         )
         try:
-            raw = await asyncio.wait_for(future, timeout)
+            data = await asyncio.wait_for(future, timeout)
         except (asyncio.TimeoutError, asyncio.CancelledError):
             self._pending.pop(request_id, None)
             raise
-        response = response_type.model_validate(raw)
```
mikasenghaas, line 91: could we not just do `asyncio.to_thread` on this?

xeophon: We do for the large-reply path below: both `msgpack.unpackb` and `model_validate` run inside `asyncio.to_thread`. A bare `await asyncio.to_thread(...)` is not enough here because concurrent replies could then expand together, and cancelling the await does not stop an already-running worker. The single capacity slot stays held until that worker actually finishes, which keeps peak expansion at one. Health/info remain synchronous because those replies are tiny.

mikasenghaas: heard

mikasenghaas (approved):

PR #1862, opened 2026-06-24
feat: add web_search mode to the codex harness

mikasenghaas:
## Summary

- Add an opt-in `web_search` field (`disabled` / `cached` / `live`) to `CodexHarnessConfig`, passed to the CLI as `-c web_search=<mode>`. `None` (the default) leaves Codex's own default, so the change is non-breaking.
- Codex's web search is a **server-side Responses hosted tool** (`{"type": "web_search"}`) that the model provider's `/responses` endpoint executes. Prime inference executes it — verified by a direct `/responses` call carrying the hosted tool, which returns real web-search results for `deepseek/deepseek-v4-flash`. Use `--harness.web_search live` to give a Codex rollout live web access.

## Notes

- This enables web search for models that speak OpenAI-structured tool calls through the provider. A model that emits a *non-OpenAI native* tool-call format as plain text (rather than structured calls) won't have its calls consumed by Codex — a model/provider concern, independent of this change.

> [!NOTE]
> ### Add `web_search` mode configuration to the Codex harness
> Adds an optional `web_search` field to `CodexHarnessConfig` in [harness.py](https://github.com/PrimeIntellect-ai/verifiers/pull/1862/files#diff-904c37fe437bb3b42e0238258ed85a8c6cede02461216bcf9c7417c5f3b72b7b), accepting `'disabled'`, `'cached'`, or `'live'`. When set, `CodexHarness.setup` appends `-c web_search=<value>` to the spawned `codex` process arguments, controlling its server-side web search behavior. When unset, behavior is unchanged.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized dfef38f.</sup>
> 
>

### DIFF
--- a/verifiers/v1/harnesses/codex/harness.py
+++ b/verifiers/v1/harnesses/codex/harness.py
@@ -9,6 +9,7 @@ session secret) is read from an env var.
 
 import logging
 import shlex
+from typing import Literal
 
 from verifiers.v1.clients import RolloutContext
 from verifiers.v1.harness import Harness, HarnessConfig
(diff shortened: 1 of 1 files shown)


mikasenghaas:
Closing — not needed for wideseek-v1. The env is harness-agnostic and validated via the rlm/default harnesses; codex already sends the `web_search` hosted tool (cached) to custom providers by default, so this only added explicit `live`/`cached`/`disabled` control. Can revisit as a standalone codex enhancement if wanted.

PR #1873, opened 2026-06-25
fix: guarantee final-assistant text is a str in v1 example env scoring

mikasenghaas:
## Summary

`AssistantMessage.content` is typed `str | None` (and is `None` for a tool-call-only final turn — providers return `content: null`), so the common scoring idiom

```python
trace.assistant_messages[-1].content if trace.assistant_messages else ""
```

guards the empty-list case but not the `None`-content case → it can yield `None`, not a str.

Normalize the v1 example tasksets to:

```python
(trace.assistant_messages[-1].content or "") if trace.assistant_messages else ""
```

so the final-assistant text is a **guaranteed str at the assignment**, independent of any downstream guard. 7 files. (Companion to the same sweep in research-environments.)

> [!NOTE]
> ### Fix final assistant message content to always be a string in v1 example environment scoring
> Across seven v1 environment tasksets, reward methods accessed `trace.assistant_messages[-1].content` directly, which could be `None` and cause type errors during scoring. Each affected file now coalesces the content to an empty string using `(trace.assistant_messages[-1].content or "") if trace.assistant_messages else ""`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 80814af.</sup>
> 
>

### DIFF
--- a/environments/code_golf_v1/code_golf_v1/taskset.py
+++ b/environments/code_golf_v1/code_golf_v1/taskset.py
@@ -29,7 +29,9 @@ SYSTEM = (
 
 def extract_program(trace: vf.Trace) -> str:
     """The python from the last assistant message's code block (or its raw text)."""
-    text = trace.assistant_messages[-1].content if trace.assistant_messages else ""
+    text = (
+        (trace.assistant_messages[-1].content or "") if trace.assistant_messages else ""
+    )
     match = re.search(r"```(?:python)?\n(.*?)```", text or "", re.DOTALL)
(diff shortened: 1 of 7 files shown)


mikasenghaas:
Closing — on review all 7 example envs touched here already coerce the final-assistant content to a str at the callsite (`or ""` / `isinstance(...) else ""`), so this change was pure redundancy. The genuinely-unguarded sites are only in research-environments (unscramble_v1, verbatim_copy_v1, ifeval_v1), fixed in PrimeIntellect-ai/research-environments#580.

PR #1880, opened 2026-06-26
feat: drive user simulator as an explicit harness feature

### LINKED ISSUE #1871: User simulator must be an explicit harness feature (transparent injection forks the graph with tools)
## Summary

A v1 rollout that uses **both task tools and a user simulator** under a harness driven by the transparent interception injection forks its message graph at every user-turn boundary (multiple `Trace.branches`, with `sampled=False` duplicate assistant nodes). Single-turn and tool-less-user-sim rollouts are unaffected. Surfaced porting BFCL v3 (multi-turn function calling) to a v1 taskset; reproduces with `deepseek/deepseek-v4-flash`.

Not a token issue — this is eval (`client.type=eval`, all node `token_ids` empty); it's a **message-level hash fork**.

## Root cause

The interception loop (`verifiers/v1/interception/server.py:306-404`) drives the user-sim transparently: on a model turn with **no tool calls**, it injects the next user turn and **re-prompts the model internally** (line 401: `prompt = [*prompt, response.message, *user_messages]`), looping until the model emits tool calls — then returns **only** that tool-calling response to the (sim-unaware) harness (line 383). The injected turns live only in the loop-local `prompt` and are discarded when the handler returns.

The default harness is a single `while True` loop (`harnesses/default/program.py`) that appends only what `chat()` returns + tool results it computes. So after a tool round it re-prompts with its **own** view, which **omits the injected user turn and the preceding no-tool "final answer"**. `prepare_turn` (line 325) matches the prefix only up to the last tool result, then forks — committing the tool-execution continuation as new `sampled=False` nodes parallel to the sim's `sampled=True` path.

### Why this only bites tools + user-sim

- **Tool-less user-sim** (alphabet_sort, color_codeword, textarena): the model never emits tool calls, so the interception loops entirely internally and the harness makes **one** `chat()` call — it never re-prompts, so nothing divergent ever re-enters the graph. Verified: `alphabet_sort_v1` on deepseek-flash, 2 user turns → **branches = 1** (linear).
- **Tools, no user-sim**: no injection, nothing to omit.
- **Tools + user-sim** (BFCL v3 multi-turn — first env to combine both): tool calls *force* a return to the harness so it can run the tool, and that round-trip is where the harness's sim-blind conversation re-enters the graph and forks it.

Reproduced graph (BFCL `multi_turn_base`, reasoning off):

```
id=3 Tool                                  <- fork parent
id=4 Assistant sampled=1 (final, no tools)    sim path (dead leaf)
id=5 User      sampled=0 (injected turn)
id=6 Assistant sampled=1 (tool call)          ^ leaf, no children
id=7 Assistant sampled=0 (same tool call)  parent=3  <- harness continuation (omits id4/id5)
id=8 Tool      parent=7
...
```

A secondary contributor with reasoning models: `message_hash` (`graph.py:206`) includes `reasoning_content` + `provider_state` (deepseek puts the reasoning in `provider_state`), but the harness re-sends prior assistants via a lossy `message.model_dump()` that drops them — so even a matched prefix forks. Adds extra branches on top of the primary cause (likely also affects **pure-tools** multi-turn — worth checking independently).

## Resolution: a user simulator must be an *explicit, supported* harness feature

There's no way to make the harness's conversation correct while keeping it sim-unaware: an OpenAI chat-completions response can only carry back one assistant message, so an injected user turn has nowhere to flow. The transparent-injection design is fundamentally at odds with a harness that runs its own tool loop. So **interactive user-sim should be a capability a harness opts into and implements**, not a transparent interception trick.

A harness that supports it makes the no-tool-call response a **turn boundary**: it fetches the next user turn from the framework, appends it, and continues its own loop — so its conversation is complete and the graph stays linear by construction. The interception then becomes a **pure 1:1 proxy** of model calls (no internal inject-and-re-prompt).

Proposed mechanism: expose the sim as an aux endpoint on the interception (alongside `/state`, `/task`), e.g. `/user` — POST the assistant's content, get back the next user message(s) or empty (=done). Each supporting harness adapts it to its own loop:

- **default / bash harness** (we own): on a no-tool response, POST `/user`; if non-empty, extend `messages` and continue; else break.
- **codex harness**: Codex headless already supports multi-turn, so it can implement the same contract — see below.

Tool-less user-sim envs can stay on the current internal-loop path (it works) or move to `/user` for one code path; I'd unify on `/user`.

Harnesses that *don't* implement the feature simply don't support interactive user-sim (declared, not silently corrupting) — but the ones we care about all can.

## How codex does it

The v1 codex harness currently runs one-shot `codex exec` (`harnesses/codex/harness.py:90`), which is why it leans on transparent injection today (and also why it hits the codex streaming-scoring race). Codex headless can drive multi-turn explicitly:

- **`codex exec resume [SESSION_ID] [PROMPT]`** (stable) — *"Resume a previous session by id … Prompt to send after resuming the session."* Run `codex exec "<turn0>"` (persist the session — no `--ephemeral`), read the session id from the `--json` JSONL events, then on each turn boundary run `codex exec resume <id> "<next user msg>"`. The session preserves full context, so the model sees the prior conversation + the new user turn. The harness becomes an exec→resume→resume… driver that fetches each user turn (via `/user`) between turns.
- **`codex app-server` / `exec-server`** (experimental) — a long-running process with a structured stdio protocol (ships `generate-json-schema` / `generate-ts` for the protocol, plus a `daemon` + `proxy`): a true interactive multi-turn channel where successive user messages are pushed into one running Codex process.

So "harness listens" generalizes even to a third-party CLI; each harness just needs an adapter to its native multi-turn mechanism (default/bash → `/user` in the loop; codex → `exec resume` or app-server).

## Secondary fix (verbatim assistant carry-through)

Under "harness listens" the harness re-sends prior assistants each call, so the `reasoning_content`/`provider_state` drift must also be addressed — the harness should carry the assistant message through **verbatim** (the interception already hands it the provider's raw bytes, `response.raw`) rather than via a lossy re-parse. Same "preserve verbatim" spirit; fixes the secondary contributor for both user-sim and pure-tools multi-turn.

## Impact / workaround

Scoring that reads `trace.branches[-1].messages` sees only the tail (earlier turns are on a sibling branch). `bfcl_v3_v1` works around it by reconstructing turns from `trace.nodes` filtered on `node.sampled`. Discovered porting `bfcl_v3_v1` (research-environments PR #568). verifiers @ f3761de3.

mikasenghaas:
## Summary

Makes the **user simulator an explicit harness feature** instead of a transparent interception trick, fixing the message-graph fork that happened whenever a rollout combined **task tools with a user simulator** (verifiers#1871).

The user simulator is already an MCP server — `vf.User` exposes a `respond` tool. So instead of the interception server playing out the exchange and handing the harness only the final assistant message (which forked the graph), the **harness drives the simulator itself over MCP**, exactly the way it already talks to tool servers:

- The `vf.User` is served as a **harness-reachable MCP server** (like a tool, but never shown to the model).
- On a no-tool-call turn, the harness's conversation driver calls the simulator's `respond` tool, appends the returned user message(s) to **its own** conversation, and re-prompts. A no-prompt task is opened by `respond("")` before the first model call. An empty reply means the simulator is done.
- So every user turn arrives in the harness's own next model call and is recorded as a **regular user message on the single branch** — no fork, no transparent injection.
- The **interception server drops out of the user-sim path entirely**: it's now a pure 1:1 model proxy that knows nothing about user simulation. The simulator's end-of-trajectory `@vf.stop` flag still reaches it over the existing `/state` channel like any other state.

This is the **same mechanism for every harness** — owned chat loops and external CLI agents alike (see below). No bespoke HTTP endpoint, no extra dependency: the program reuses the `mcp` client it already uses for tools.

### Changes
- **interception** (`server.py`): `handle_request` is a 1:1 model proxy; removed the internal inject-and-re-prompt loop, the opening-turn seed, and the `RolloutSession.user` field. (No `/user` endpoint, no `Dialect.extend` — both removed.)
- **launch** (`serve_user`): serves the `vf.User` harness-reachable (like a tool) and yields its MCP URL; `connect_user` stays as the host-side driver utility (for CLI-wrapper harnesses).
- **harness**: `Harness.run`/`launch` thread `user_url` (the simulator's MCP server); the rollout **gates** on `SUPPORTS_USER_SIM` (loud failure, not silent corruption).
- **default / bash / bash_edit** programs connect to the simulator's MCP server and call `respond` (a small `connect_user_sim` helper); deps stay `["openai", "mcp"]`.
- `rlm` / `codex` / others get the threaded param but keep `SUPPORTS_USER_SIM = False`.

## Why

Under transparent injection the interception server played the whole user/assistant exchange inside one program request and returned only the final assistant message to the harness. A harness that runs its own tool loop then re-prompted with its **own** view — which omitted the injected user turns — so `graph.prepare_turn` matched only a short prefix and forked the graph (`sampled=False` duplicate branches). Single-turn and tool-less user-sim rollouts were unaffected; **tools + user-sim** (e.g. BFCL v3 multi-turn) forked every turn, and naive `trace.branches[-1]` scoring saw only the tail. An OpenAI chat response can carry back only one assistant message, so there's no way to keep the harness's conversation correct while keeping it sim-unaware — the simulator has to be a capability the harness opts into.

## How CLI agents fit (same mechanism)

The user turn always comes from the simulator's `respond` tool and is injected through the harness's **native multi-turn input**, so it's a real user message and the graph stays linear — only the injection verb differs per agent (and it's each agent's real entry point, not a hack):

- **Codex** headless ([resume](https://developers.openai.com/codex/cli/reference)): `codex exec "<open>"` then, per turn, `respond(...)` → `codex exec resume <session_id> "<user msg>"`. Codex reconstructs context from its transcript and re-prompts with the user turn in history.
- **Claude Code** headless ([stream-json](https://code.claude.com/docs/en/headless)): `claude -p --input-format stream-json --output-format stream-json` is a bidirectional session — `respond(...)` → write a `{"type":"user",...}` line to stdin; Claude appends it and re-prompts.

In both, the host-side wrapper calls the same `respond` tool (via `connect_user`) and feeds the result into the CLI's own resume/stdin channel. (These two harnesses still set `SUPPORTS_USER_SIM = False` here — implementing their drivers is a follow-up; Codex also can't yet use our MCP *tools*, `SUPPORTS_MCP = False`.)

## Verification

New toy taskset `tool_user_sim_v1` (a calculator that must call a `calc_add` MCP tool **and** answer a user simulator across turns; no prompt, so the simulator opens it), plus an e2e regression test asserting the graph stays linear. On the `default` harness against `deepseek/deepseek-v4-flash`:

| | reward | `num_branches` |
|---|---|---|
| **before** (transparent injection) | 0.67 | **4** |
| **after** (harness-driven MCP) | 1.00 | **1** |

After, the graph is a single linear path: `system → user(2+3) → assistant(tool call) → tool(5) → assistant(<answer>5</answer>) → user(10+20) → …`. `bash` and `bash_edit` were smoke-checked on the same env (reward 1.0, `num_branches == 1`).

No regressions on the existing envs (all reward 1.0, `num_branches == 1`):
- `echo_user_sim_v1` (tool-less user sim, previously relied on transparent injection)
- `alphabet_sort_v1` (user sim, opens the conversation)
- `counter_tool_v1` (pure tools, no user sim)

Closes #1871.

### DIFF
--- a/environments/compact/compact/harness.py
+++ b/environments/compact/compact/harness.py
@@ -38,6 +38,7 @@ class CompactingHarness(Harness[CompactingHarnessConfig]):
         endpoint: str,
         secret: str,
         mcp_urls: dict[str, str],
+        user_url: str | None = None,
     ) -> ProgramResult:
         env = {
             "OPENAI_BASE_URL": endpoint,
--- a/tests/v1/fixtures/pyproject.toml
+++ b/tests/v1/fixtures/pyproject.toml
@@ -17,5 +17,6 @@ include = [
     "echo_tool_v1.py",
     "echo_user_sim_v1.py",
     "tool_response_image_v1.py",
+    "tool_user_sim_v1/**/*.py",
     "pyproject.toml",
 ]
--- a/tests/v1/fixtures/tool_user_sim_v1/__init__.py
+++ b/tests/v1/fixtures/tool_user_sim_v1/__init__.py
@@ -1,0 +1,3 @@
+from tool_user_sim_v1.taskset import ToolUserSimTaskset
+
+__all__ = ["ToolUserSimTaskset"]
--- a/tests/v1/fixtures/tool_user_sim_v1/servers/tools.py
+++ b/tests/v1/fixtures/tool_user_sim_v1/servers/tools.py
@@ -1,0 +1,14 @@
+import verifiers.v1 as vf
+
+
+class CalcToolset(vf.Toolset[vf.ToolsetConfig]):
+    TOOL_PREFIX = "calc"  # the model sees `calc_add`
+
+    @vf.tool
+    def add(self, a: int, b: int) -> str:
+        """Add two integers and return their sum."""
+        return str(a + b)
+
+
+if __name__ == "__main__":
+    CalcToolset.run()
--- a/tests/v1/fixtures/tool_user_sim_v1/servers/user.py
+++ b/tests/v1/fixtures/tool_user_sim_v1/servers/user.py
@@ -1,0 +1,27 @@
+import verifiers.v1 as vf
+
+
+class CalcState(vf.State):
+    user_finished: bool = False
+
+
+class CalcUser(vf.User[vf.UserConfig, CalcState]):
+    """Poses the task's addition problems one per turn; after the last, flags `user_finished`
+    (the taskset's `@vf.stop` ends the rollout on it). The task carries no prompt, so the opening
+    `respond("")` delivers the first problem."""
+
+    async def setup_task(self, task) -> None:
+        self.problems = task.problems
+        self.i = 0
+
+    async def respond(self, message: str) -> vf.Messages:
+        if self.i >= len(self.problems):
+            self.state.user_finished = True
+            return []
+        a, b = self.problems[self.i]
+        self.i += 1
+        return [{"role": "user", "content": f"What is {a} + {b}?"}]
+
+
+if __name__ == "__main__":
+    CalcUser.run()
--- a/tests/v1/fixtures/tool_user_sim_v1/taskset.py
+++ b/tests/v1/fixtures/tool_user_sim_v1/taskset.py
@@ -1,0 +1,94 @@
+"""tool_user_sim (v1): a calculator that must call an MCP tool AND answer a simulated user.
+
+The minimal env that combines task tools with a user simulator. Each turn the user simulator
+poses an addition problem; the model calls the `calc_add` tool to compute it, then replies with
+the answer; the simulator poses the next problem, and after the last one flags `user_finished`
+(the `@vf.stop` ends the rollout). The task carries no prompt — the simulator opens the
+conversation.
+
+Tools (the harness runs its own tool loop) + a user simulator is exactly the case that forks the
+message graph under transparent user injection (see verifiers#1871): `num_branches` is 1 only
+when the conversation stays linear, so it is the regression signal for driving the user simulator
+as an explicit harness feature.
+"""
+
+import re
+
+import verifiers.v1 as vf
+
+from tool_user_sim_v1.servers.tools import CalcToolset
+from tool_user_sim_v1.servers.user import CalcState, CalcUser
+
+PROBLEMS = [
+    (2, 3),
+    (10, 20),
+    (7, 8),
+]  # the user poses these in order; their sums are the answers
+SYSTEM = (
+    "You are a calculator. For each problem the user gives you, call the `calc_add` tool to add "
+    "the two numbers, then reply with only the result inside <answer></answer> tags. Always use "
+    "the tool; do not add the numbers yourself."
+)
+
+
+class ToolUserSimTask(vf.Task):
+    problems: list[list[int]]
+    """The `(a, b)` pairs the user poses, one per turn; their sums are the expected answers."""
+
+
+class ToolUserSimConfig(vf.TasksetConfig):
+    tools: vf.ToolsetConfig = vf.ToolsetConfig()
+    user: vf.UserConfig = vf.UserConfig()
+
+
+class ToolUserSimTaskset(vf.Taskset[ToolUserSimTask, ToolUserSimConfig, CalcState]):
+    @vf.stop
+    async def user_finished(self, trace: vf.Trace) -> bool:
+        return trace.state.user_finished
+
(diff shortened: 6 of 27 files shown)


mikasenghaas:
Closing stale PR (older than 2 months). Feel free to reopen if it is still relevant.

PR #1888, opened 2026-06-27
fix(router): clear request bookkeeping on cancel (forward_cancel leak)

joanvelja:
### Problem
`EnvRouter.forward_cancel` (added in #1055) forwards the cancel signal to the owning worker but never clears its tracking entries — `request_to_worker[request_id]` and `worker.active_requests[request_id]`. The only code that clears them, `complete_request`, runs solely on the response path. A request cancelled *before* it produces a response therefore never gets cleaned up, so both dicts leak one entry per such cancel. Routers that routinely cancel in-flight requests (e.g. off-policy / lag-based RL that cancels stale rollouts) accumulate stale entries unboundedly.

### Fix
`forward_cancel` calls `complete_request(request_id)` after forwarding the cancel. The pop is idempotent, so a late in-flight response that also calls `complete_request` is harmless.

### Test
`tests/test_env_router_cancel.py` — constructs the router via `__new__` (no ZMQ), asserts a cancel clears both dicts, and that an unknown request is a no-op. Fails on `main` (entries retained), passes with the fix.

> [!NOTE]
> ### Fix request bookkeeping leak in `EnvRouter.forward_cancel`
> - `forward_cancel` in [env_router.py](https://github.com/PrimeIntellect-ai/verifiers/pull/1888/files#diff-5f42e5c44b82aeddf6020de01b59c5480cac9144a3f6ffc79c3ee01683ad92ec) now calls `complete_request` after forwarding the cancel frame, removing the request from `request_to_worker` and the worker's `active_requests`.
> - Previously, cancelled requests were never cleaned up, causing a permanent bookkeeping leak since no response arrives to trigger normal cleanup.
> - Cleanup runs unconditionally (outside the send conditional) and uses idempotent `pop` calls, so late responses that arrive after cancellation still clean up safely.
> - Adds regression tests in [test_env_router_cancel.py](https://github.com/PrimeIntellect-ai/verifiers/pull/1888/files#diff-2b3766ba6ca4585aba67e9493c2c4c291343f84dc51f11842b2b716335b7b025) covering both the cleanup path and the no-op behavior for unknown request IDs.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 051b399.</sup>
> 
>

### DIFF
--- a/tests/test_env_router_cancel.py
+++ b/tests/test_env_router_cancel.py
@@ -1,0 +1,55 @@
+"""Regression: EnvRouter.forward_cancel must clear request bookkeeping (no leak).
+
+A cancel that yields no terminal response would otherwise leave
+``request_to_worker`` + the worker's ``active_requests`` populated forever — the
+``complete_request`` cleaner runs only on the response path, which never fires
+for a request cancelled before it responds. (Upstream #1055; surfaced in the
+prime-rl #76 PR1 review as F3.)
+
+Builds the router via ``__new__`` to skip the ZMQ-IPC ``__init__`` —
+``forward_cancel`` only touches ``self.workers`` + ``self.request_to_worker``.
+"""
+
+import asyncio
+
+from verifiers.serve.server.env_router import ActiveRequestInfo, EnvRouter, WorkerHandle
+
+
+class FakeSock:
+    def __init__(self) -> None:
+        self.sent: list = []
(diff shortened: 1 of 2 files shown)


joanvelja:
Closing — opened in error; we'll track this in our fork.

PR #1890, opened 2026-06-28
fix: download sandbox logs in PrimeRuntime.run to avoid read-file 10MB limit

mikasenghaas:
## Summary

- `PrimeRuntime.run` polled `get_background_job`, which reads the job's stdout/stderr logs **inline** via the SDK's `read_file` → gateway `/read-file` endpoint. That endpoint caps reads at 10MB and returns **HTTP 413** (`File size (...) exceeds max read size (10485760 bytes). Use /download instead`). Any command emitting >10MB of output (e.g. a verbose test runner) therefore killed the whole rollout with `SandboxError: prime exec failed: Read file failed: HTTP 413 ...`.
- Fix: poll the tiny exit file directly with `read_file`, then fetch stdout/stderr via `download_file` (no size limit) once the job completes. This mirrors `read()`, which already downloads to dodge this exact limit. The download is factored into a shared `_download_bytes` helper that `read()` now reuses.

This is a latent bug for **any** env whose program emits >10MB on stdout/stderr; it surfaced first in `scaleswe_v1` (its `pytest -vv` scorer produced ~28MB of stdout).

## Verification

Before: a scaleswe-v1 rollout whose scorer emitted 28MB of stdout failed with
```
SandboxError: prime exec failed: Read file failed: HTTP 413 GET .../read-file?path=%2Ftmp%2Fjob_d370cedb.stdout.log: {"error":"File size (28135205 bytes) exceeds max read size (10485760 bytes). Use /download instead."}
```
After: stdout/stderr are downloaded rather than read inline, so the 10MB read-file cap no longer applies. (No behavior change for small output; the background-job wrapper always creates both log files via redirection, so the download cannot 404.)

> [!NOTE]
> ### Download sandbox logs in `PrimeRuntime.run` to avoid the 10MB read-file limit
> - Replaces the `get_background_job` polling approach in [`prime.py`](https://github.com/PrimeIntellect-ai/verifiers/pull/1890/files#diff-d22d61279ff7587a2b14c72ab1a0fc9d3f805a735338a70f8efc7fe19aaa8502) with direct polling of the job's exit file via `client.read_file`, parsing the exit code from its contents.
> - After detecting job completion, stdout and stderr are fetched via a new `_download_bytes` helper that uses `client.download_file` to a temp location, bypassing the gateway's inline read-file size limit.
> - Refactors `PrimeRuntime.read` to delegate to the same `_download_bytes` helper for consistency.
> - Behavioral Change: `run` now returns stdout/stderr decoded from downloaded files rather than inline API responses; large log outputs that previously failed or were truncated will now be returned in full.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8d5cc55.</sup>
> 
>

### DIFF
--- a/verifiers/v1/runtimes/prime.py
+++ b/verifiers/v1/runtimes/prime.py
@@ -120,6 +120,8 @@ class PrimeRuntime(Runtime):
             raise SandboxError(f"prime sandbox provisioning failed: {e}") from e
 
     async def run(self, argv: list[str], env: dict[str, str]) -> ProgramResult:
+        from prime_sandboxes import SandboxFileNotFoundError
+
         try:
             # Poll directly so the rollout stage owns the execution timeout; the SDK helper
             # otherwise imposes its own 15-minute limit.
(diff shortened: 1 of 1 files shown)


mikasenghaas:
Closing stale PR (older than 2 months). Feel free to reopen if it is still relevant.

PR #1898, opened 2026-06-30
feat: agentic group judge for v1 tasksets (AgenticGroupJudge)

mikasenghaas:
## What

Adds `vf.AgenticGroupJudge` — a reusable, **agentic, group-level** LLM judge for v1 tasksets.

Where `Judge` (`judge.py`) is one chat call that scores **one** rollout from text in its prompt, `AgenticGroupJudge` scores a whole **group** of a task's rollouts at once: it gives the judge a sandbox containing every candidate, lets it **investigate with tools** (run the pristine tests, read diffs/transcripts, look for regressions or test-tampering), and emit **one comparative score per rollout**. It is the runtime behind a taskset's `@group_reward`.

## Why

For RL we want to run a group of rollouts and have a judge return a list of scores (one per rollout) that become the group's rewards (GRPO baselines them downstream). A prompt-stuffing judge can't run code or inspect a filesystem; an agentic judge in a sandbox can — e.g. run the *hidden* tests itself instead of us wiring a bespoke verifier, and catch a rollout that passed the shown tests by editing them.

This reuses the existing seams rather than adding new ones: `@group_reward` already exists and already flips `requires_group_rollouts`, and `Runtime`/`make_runtime` already provision sandboxes. The judge collapses into the taskset's group scoring as a reusable component (no new env/router machinery).

## How it's used

Two wiring points on the taskset (see `environments/swe_group_judge_v1/`):

```python
class SweJudge(vf.AgenticGroupJudge):
    repo_path = "/repo"
    def tests_tar(self, task): return make_tar(task.tests_dir)

class SweGroupJudgeTaskset(vf.Taskset[SweTask, JudgedConfig]):
    @functools.cached_property
    def judge(self): return SweJudge(self.config.judge, self.config.judge_runtime)

    async def finalize(self, task, trace, runtime):     # runtime still live → snapshot
        diff = await runtime.run(["sh","-c","cd /repo && git add -A && git diff --cached --binary"], {})
        trace.info["judge"] = {"patch": diff.stdout, "transcript": vf.render_transcript(trace)}

    @vf.group_reward(weight=1.0)                         # → requires_group_rollouts
    async def panel(self, traces, task) -> list[float]:
        return await self.judge.score(traces, task)
```

Lifecycle: each rollout tears down its own runtime before group scoring (episode.py), so `finalize` snapshots the diff + transcript onto the trace while the runtime is live; `score()` then provisions **one** fresh runtime from `task.image` (`resolve_runtime_config` injects it), rebuilds each candidate (concurrently) as `base + that rollout's patch` under `/judge/rollouts/<i>` (shuffled to defeat position bias), stages the pristine tests, runs the judge agent, and returns one score per trace.

Failure handling: a group is N full agentic rollouts, so a single judge fumble doesn't drop it — recoverable tool/format errors (bad JSON args, wrong score count, non-numeric scores, `read_file` on a bad path, unknown tool) are fed back to the judge to retry within `max_turns`. Only a hard failure — never submitting valid scores within `max_turns`, an infra/clone failure, or a missing snapshot — propagates and drops the group. The judge's per-rollout rationale is recorded onto `trace.info["judge_rationale"]` for auditability.

The harness scaffolding (`_cand.patch`, `_apply.log`, `_transcript.txt`, `_tests/`) is git-excluded inside each candidate clone, so the judge's `git diff`/`git status` reflect only that candidate's own change — keeping the diff/tamper inspection clean.

## Changes

- `verifiers/v1/group_judge.py` — `AgenticGroupJudge` + `render_transcript`.
- `verifiers/v1/__init__.py` — export both.
- `environments/swe_group_judge_v1/` — example taskset, modeled on `code_golf_v1`.

## Validation

- ruff + repo pre-commit clean.
- Imports and all dependent seams resolve against the package; the example taskset loads and its `@group_reward` is discovered.
- **Runtime smoke** (`SubprocessRuntime`, stubbed judge LLM): builds a base git repo + N candidate patches where each candidate prints its true id, runs `score()` end to end (clone → apply → stage tests → tool-loop → `submit_scores`), and asserts: (a) the scores unshuffle back to `[0, 1, 2]` — exercising staging, dispatch, and the shuffle→unshuffle mapping; (b) the candidate's `git status` shows only its own change, confirming the scaffolding is git-excluded; (c) a deliberately wrong-count `submit_scores` is fed back and retried (not dropped) and the rationale is recorded on the traces.

Not yet exercised: a live judge model doing tool-calls and a real container image (`task.image` with the repo at `repo_path`) — i.e. a full `prime eval run`. Image-dependent knobs (`repo_path`, the `finalize` diff path, where the pristine tests live) are documented and overridable.

## Notes / open questions

- Score scale: the judge returns raw comparative floats (GRPO uses only within-group spread); we could normalize for log readability.
- Test-run hygiene (per-test timeouts, flaky reruns) and transcript anonymization depth are left to follow-ups.
- The example env uses a placeholder image so it won't `prime eval run` as-is — it's a wiring template.

### DIFF
--- a/verifiers/v1/group_judge.py
+++ b/verifiers/v1/group_judge.py
@@ -39,9 +39,11 @@ task's pristine test suite — then call `score` from `@group_reward`. The shape
                 return await self.judge.score(traces, task)
 
 The scores returned by ``@group_reward`` are summed (× weight) into each ``trace.reward``; in
-prime-rl those become the group's rewards and GRPO baselines them. Any failure — the judge
-times out, returns the wrong count, a snapshot is missing — propagates, which marks the group
-errored and (in prime-rl) drops it. The judge model/endpoint is a ``JudgeConfig`` (its own
+prime-rl those become the group's rewards and GRPO baselines them. A hard failure — the judge
+never submits valid scores within ``max_turns``, the sandbox/clone fails, or a rollout's snapshot
+is missing — propagates, marking the group errored (dropped in prime-rl); recoverable tool/format
+mistakes (bad JSON, wrong count, a missing file) are fed back for the judge to retry. The judge
+model/endpoint is a ``JudgeConfig`` (its own
 endpoint, not the policy); the judge sandbox is a ``RuntimeConfig`` resolved per task so it
 carries the task's toolchain (the candidates' tests actually run).
 """
@@ -195,10 +197,11 @@ class AgenticGroupJudge:
             await asyncio.gather(
                 *(self._stage_rollout(runtime, displayed, traces[real], tests) for displayed, real in enumerate(order))
             )
-            displayed_scores = await self._run_agent(runtime, n)
+            displayed_scores, displayed_rationale = await self._run_agent(runtime, n)
             scores = [0.0] * n
             for displayed, real in enumerate(order):
                 scores[real] = displayed_scores[displayed]
+                traces[real].info["judge_rationale"] = displayed_rationale[displayed]  # audit trail
             return scores
         finally:
             await runtime.stop()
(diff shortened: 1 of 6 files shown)


verifiers/v1/group_judge.py
```diff
@@ -0,0 +1,293 @@
...
+    {
+        "type": "function",
+        "function": {
+            "name": "submit_scores",
+            "description": "Submit one comparative score and rationale per rollout, in displayed order.",
+            "parameters": {
+                "type": "object",
+                "properties": {
+                    "scores": {"type": "array", "items": {"type": "number"}},
+                    "rationale": {"type": "array", "items": {"type": "string"}},
```
rasdani, line 126: would put the rationale before the scores

PR #1900, opened 2026-06-30
feat(default): add an optional search tool

hallerite:
## What

**PR 2 of 2** — this is #1900 rebased on top of the harness rename (#1904 → this).

Adds an opt-in `search` tool to the **`default`** harness (the bash + edit agent, formerly `bash_edit`) so it can drive "bring your own search" tasksets (`openseeker_v1`, `redsearcher_v1`) that ship no search tool of their own. **Off by default.**

### Re-aligned with the rlm `search` skill
Per review, the search implementation is now ported from the canonical `rlm/skills/search.py` (the same place the `edit` tool came from) instead of the original ad-hoc version:
- Serper Google search via **`httpx`** (added to the uv-script deps), organic results formatted as `Result N: title / URL / - snippet`, `num_results=5`.
- Two harness-local adaptations kept from the original #1900 design (justified by the harness context, where rlm's skill runs in an IPython kernel):
  - the `SERPER_API_KEY` is handed to the program over **argv** (`--serper-key`), not the env, so the agent's `bash` subprocesses don't inherit it (same treatment as the interception secret);
  - failures **return an error string** (like `run_bash`/`run_edit`) rather than raising, so a transient search failure can't kill the rollout.

### `search` tool
With `--harness.search true`:
- `DefaultHarnessConfig.search: bool = False` gates it.
- The harness requires `SERPER_API_KEY` in the eval env, passes it via `--serper-key` on argv, and appends a tool note to the injected system prompt.
- The program registers a `search` function tool dispatched on the same `asyncio.to_thread` path as `bash`/`edit`. Function tools are forwarded by the v1 dialects, so any model works.

### Client: no retries + generous timeout
The program's OpenAI client is pinned to `max_retries=0` with a long `timeout`. A client-side retry that re-sends a request whose turn the interception already committed re-samples it and **forks the trace into an extra dead-end branch** — and every branch becomes a training sample, silently polluting the data. No retries keeps each rollout's trace linear; the long timeout lets a slow turn finish. Transient model failures surface as error rows that `--resume`/`--prune` recover.

## Validation
`ruff check` clean; program compiles and `--help` shows the new flags; `run_search` unit-exercised against a mock Serper response (format, empty, no-key, and error paths). Live validation from the original #1900 (glm against a self-hosted deployment) still applies to the search/trace behavior.

## Usage
```bash
uv run eval <search-taskset> --harness.id default --harness.search true --harness.runtime.type prime
```

---

> [!NOTE]
> **Medium Risk**
> Touches rollout tracing behavior (no client retries) for all default-harness runs, and handles a third-party API secret via argv stripping; misconfiguration fails at launch rather than silently leaking the key into bash.
> 
> **Overview**
> Adds an **opt-in** `search` tool to the **`default`** harness so search-oriented tasksets can run without shipping their own search MCP server. Enable with `--harness.search true` and `SERPER_API_KEY` in the host or `--harness.env`.
> 
> **Harness:** `DefaultHarnessConfig.search` defaults to `false`. When enabled, the harness appends `SEARCH_PROMPT` to the injected system text, passes `--search` and `--serper-key` on argv, and **pops** `SERPER_API_KEY` from the program env so `bash` subprocesses do not inherit the key (empty env value is treated as intentional masking, not a host fallback). Missing key raises `ValueError` at launch.
> 
> **Program:** Adds `httpx`, a `search` function tool, and `run_search` (Serper organic results formatted as title/URL/snippet; failures return tool error strings). The OpenAI client is set to **`max_retries=0`** and a **1800s timeout** so retries do not re-sample turns already committed by interception and fork training traces.
> 
> **Docs:** `GUIDE.md` documents `--harness.search true` on the built-in `default` harness row.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 43ac0eac4a83216ef1ca087733cfc02c349950a0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add optional `web_search` tool to the `default` harness backed by serper.dev
> - Adds a `web_search` boolean field to `DefaultHarnessConfig` (default `false`) that, when enabled, exposes a serper.dev-powered search tool to the model.
> - The harness reads `SERPER_API_KEY` from the environment, strips it from the subprocess environment to prevent bash subprocess leakage, and passes it to the program via `--web-search` and `--serper-key` argv flags.
> - `program.py` gains a `run_web_search` function that POSTs to `https://google.serper.dev/search` using `httpx`, parses organic results, and returns formatted title/URL/snippet strings or an error message.
> - The chat loop in `program.py` now uses a fixed 1800s timeout with `max_retries=0` instead of retrying client requests.
> - Risk: if `web_search` is enabled but `SERPER_API_KEY` is missing, `DefaultHarness.launch` raises a `ValueError` at startup.
>
> 
> #### Changes since #1900 opened
>
> - Renamed the exposed tool from `web_search` to `search` in the tool schema definition `SEARCH_TOOL`, updated the function name field in the schema, and modified the tool description text to reference Serper and Google [e056d66]
> - Renamed the CLI flag from `--web-search` to `--search` in the argument parser, and updated the main async entrypoint to conditionally wire and dispatch the `search` tool based on `args.search` instead of `args.web_search` [e056d66]
> - Renamed the configuration field from `web_search` to `search` in the `DefaultHarnessConfig` dataclass, updated the harness launch handler to use `config.search`, pass `--search` and `--serper-key` flags to the program instead of `--web-search`, and modified error messages and comments to reference 'search' [e056d66]
> - Replaced `run_web_search` with `run_search` function, extracted result formatting logic into a new `format_results` helper function, and updated error messages to use 'search failed' wording [e056d66]
> - Renamed the constant from `WEB_SEARCH_PROMPT` to `SEARCH_PROMPT` in the harness module and updated module docstrings across harness and program modules to refer to 'search' instead of 'web_search' [e056d66]
> - Updated documentation in `GUIDE.md` to refer to the tool as 'search' instead of 'web_search', changing the harness description, CLI flag from `--harness.web-search true` to `--harness.search true`, and the tool name from 'web_search' to 'search' [e056d66]
> - Removed references to the rlm `search` skill from docstrings [1ae7e4d]
> - Modified `SERPER_API_KEY` resolution logic in `DefaultHarness.launch` handler to pop the key from harness environment and fall back to host environment only when the key is absent (None), treating empty strings as present values that trigger `ValueError` instead of falling back to host environment [43ac0ea]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized fb6daf9.</sup>
> 
>
>

### DIFF
--- a/verifiers/v1/harnesses/default/harness.py
+++ b/verifiers/v1/harnesses/default/harness.py
@@ -87,14 +87,17 @@ class DefaultHarness(Harness[DefaultHarnessConfig]):
         if self.config.edit:
             args.append("--edit")
         if self.config.search:
-            # Resolve the key from the harness env (--harness.env / forward_env) or the host, and
-            # keep it OUT of the program env: it's handed to the program over argv (--serper-key), so
-            # popping it here stops the agent's `bash` subprocesses from also inheriting it via
-            # $SERPER_API_KEY / /proc/self/environ. Only when search is enabled — otherwise env is
-            # left untouched, so a key forwarded for the agent's own bash-side use survives.
-            serper_key = env.pop("SERPER_API_KEY", None) or os.environ.get(
-                "SERPER_API_KEY"
-            )
+            # Resolve the key and keep it OUT of the program env: it's handed to the program over
+            # argv (--serper-key), so popping it here stops the agent's `bash` subprocesses from
+            # inheriting it via $SERPER_API_KEY / /proc/self/environ. Prefer a key set in the harness
+            # env (--harness.env / forward_env); fall back to the host env only when the key is
+            # *absent* (None), not present-but-empty — a rollout setting SERPER_API_KEY="" is
+            # deliberately masking the host secret, so honor that (the check below then fails loudly
+            # rather than leaking the host key). The pop is scoped to search=true, so an unrelated
+            # key forwarded for the agent's own bash-side use is left untouched.
+            serper_key = env.pop("SERPER_API_KEY", None)
+            if serper_key is None:
+                serper_key = os.environ.get("SERPER_API_KEY")
             if not serper_key:
                 raise ValueError(
                     "default search=true requires SERPER_API_KEY in the eval environment "
--- a/verifiers/v1/harnesses/default/harness.py
+++ b/verifiers/v1/harnesses/default/harness.py
@@ -87,12 +87,20 @@ class DefaultHarness(Harness[DefaultHarnessConfig]):
         if self.config.edit:
             args.append("--edit")
         if self.config.web_search:
-            key = os.environ.get("SERPER_API_KEY")
-            if not key:
+            # Resolve the key from the harness env (--harness.env / forward_env) or the host, and
+            # keep it OUT of the program env: it's handed to the program over argv (--serper-key), so
+            # popping it here stops the agent's `bash` subprocesses from also inheriting it via
+            # $SERPER_API_KEY / /proc/self/environ. Only when web_search is enabled — otherwise env
+            # is left untouched, so a key forwarded for the agent's own bash-side use survives.
+            serper_key = env.pop("SERPER_API_KEY", None) or os.environ.get(
+                "SERPER_API_KEY"
+            )
+            if not serper_key:
                 raise ValueError(
-                    "default web_search=true requires SERPER_API_KEY in the eval environment"
+                    "default web_search=true requires SERPER_API_KEY in the eval environment "
+                    "(the host env or --harness.env)"
                 )
-            args += ["--web-search", f"--serper-key={key}"]
+            args += ["--web-search", f"--serper-key={serper_key}"]
         if mcp_urls:
             # The program connects to the tool servers over HTTP; hand it a standard
             # `mcpServers` URL config (the `mcp` client itself comes from the uv deps).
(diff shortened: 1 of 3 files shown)


verifiers/v1/harnesses/default/harness.py
```diff
@@ -37,6 +44,11 @@ class DefaultHarnessConfig(HarnessConfig):
     """Offer the local `edit` tool (single-occurrence string replacement in a file) alongside
     `bash`. On by default; set `--harness.edit false` for a bash-only agent."""
 
+    web_search: bool = False
```
mikasenghaas: just `search`?

verifiers/v1/harnesses/default/program.py
```diff
@@ -66,6 +70,74 @@
 }
 
 
+WEB_SEARCH_TOOL = {
+    "type": "function",
+    "function": {
+        "name": "web_search",
```
mikasenghaas: search?

mikasenghaas (commented):
is it +- the same search tool as we have in rlm?

verifiers/v1/harnesses/default/program.py
hallerite: sure

samsja (approved):

PR #1986, opened 2026-07-13
fix(v1): tolerate an unknown service_tier in the chat dialect

Ziems:
## What

Make the OpenAI chat dialect tolerate a `service_tier` value outside OpenAI's fixed set when parsing a completion response, instead of raising a pydantic `literal_error`.

## Why

The OpenAI SDK types `ChatCompletion.service_tier` as `Literal["auto", "default", "flex", "scale", "priority"]`, and `ChatDialect.validate_response` parses responses with a strict `response_type.model_validate`. Providers return tiers outside that set — Prime's **provisioned** deployments report `service_tier: "provisioned"` — so an otherwise valid completion is rejected:

```
Input should be 'auto', 'default', 'flex', 'scale' or 'priority'
[type=literal_error, input_value='provisioned', input_type=str]
```

This breaks any `eval` / `gepa` run whose model is served from a provisioned deployment.

The OpenAI SDK's *own* response parsing uses `model_construct` (no validation), so it never hit this — only our explicit `model_validate` in the dialect did. The anthropic dialect already tolerates a non-standard `service_tier`; this brings the OpenAI/chat path in line.

## How

Subclass the SDK's `ChatCompletion` in the chat dialect, widening `service_tier` to `str | None`, and use it as the dialect's `response_type`. Parsing stays lenient about the label and keeps it (rather than dropping the field); `service_tier` isn't consumed downstream, so the change is transparent.

## Testing

- New regression test (`tests/v1/test_dialects.py`): `ChatDialect.validate_response` parses a completion carrying `service_tier: "provisioned"` and preserves the value.
- Verified end-to-end: `uv run gepa reverse-text-v1 --model google/gemini-3-flash-preview --num-train 20 --num-val 10` (served from a provisioned deployment) now completes — it failed with the `literal_error` before this change.
- `ruff` clean; `ty` diagnostics unchanged (the file's pre-existing ones are unrelated to this field).

---

> [!NOTE]
> **Low Risk**
> Narrow parsing/typing change on an unused response field; no auth, data, or business-logic impact.
> 
> **Overview**
> Fixes chat completion parsing when upstream returns a **`service_tier`** label outside the OpenAI SDK’s fixed literal set (e.g. Prime **`provisioned`**), which previously caused `ChatDialect.validate_response` to raise a pydantic `literal_error` and break eval/gepa runs.
> 
> Adds **`ModdedChatCompletion`**, a thin subclass of the SDK’s `ChatCompletion` with **`service_tier: str | None`**, and wires it in as **`ChatDialect.response_type`** and for streamed reassembly in **`ChatStreamParser.finish`**, so validation stays lenient while preserving the field (it is not used downstream).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dba456c0d59de2f27a561096805fdb408f999b99. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Tolerate unknown `service_tier` values in the chat completion dialect
> The upstream OpenAI SDK enforces a fixed set of allowed values for `service_tier`, but some providers return custom labels like `'provisioned'` that fail validation. [chat.py](https://github.com/PrimeIntellect-ai/verifiers/pull/1986/files#diff-5d2fcbf59a03489b63b0626038a853eee89543b591074597161ee29aab901280) introduces a local `ChatCompletion` subclass that widens the `service_tier` field type to `str | None`, overriding the SDK's strict enum. A new test in [test_dialects.py](https://github.com/PrimeIntellect-ai/verifiers/pull/1986/files#diff-c56f2b682d430fdcc55174c9e38af706ab8105e48f2a997f682c191f9462a668) confirms that `ChatDialect.validate_response` accepts and preserves unknown tier labels without error.
>
> 
> #### Changes since #1986 opened
>
> - Renamed the local `ChatCompletion` subclass to `ModdedChatCompletion` within the `verifiers.v1.dialects.chat` module to widen the `service_tier` field type from a restricted set to `str | None`, and updated all references in `ChatStreamParser` response assembly and `ChatDialect.response_type` to use `ModdedChatCompletion` while maintaining `openai.types.chat.ChatCompletion` in class generics [dba456c]
> - Deleted the `tests.v1.test_dialects` test file [dba456c]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1d73aeb.</sup>
> 
>
>


tests/v1/test_dialects.py
```diff
@@ -0,0 +1,31 @@
+"""Chat dialect response parsing."""
```
Ziems: Maybe I should delete these tests, LMK

samsja (dismissed):

tests/v1/test_dialects.py
mikasenghaas: ya remove

verifiers/v1/dialects/chat.py
```diff
@@ -11,7 +11,7 @@
 from dataclasses import dataclass, field as dataclass_field
 from typing import Any
 
-from openai.types.chat import ChatCompletion
+from openai.types.chat import ChatCompletion as OpenAIChatCompletion
```
mikasenghaas: lets import the og

verifiers/v1/dialects/chat.py
```diff
@@ -30,6 +30,16 @@
     content_to_parts,
 )
 
+
+class ChatCompletion(OpenAIChatCompletion):
```
mikasenghaas: and call this `ModdedChatCompletion`

PR #2001, opened 2026-07-14
Reuse uv across runtime provisioning

xeophon:
## Overview

Reuse a successful `uv` installation across provisioning consumers within the same runtime and `HOME`.

## Change

The runtime records which `HOME` values have completed a uv-backed command successfully. The first caller for each value performs the existing bootstrap under a lock; later PEP 723 scripts and colocated sandbox server installs using the same `HOME` only restore `PATH` and `UV_INSTALL_DIR`.

Per-script dependency synchronization, content-addressed script paths, and sandbox server virtual environments remain unchanged. A different `HOME` receives its own bootstrap, and a failed first command remains unready so the next caller retries.

## Impact

- Avoids one complete uv bootstrap for GSM8K's later verifier script, reducing its scoring path by approximately 1.3 seconds.
- Avoids approximately 0.49 seconds of redundant uv installation checks for each later colocated MCP or user server using the same `HOME`.
- Prevents concurrent first users of the same install directory from installing uv more than once.
- Requires no custom image and does not share dependency environments between scripts or servers.

> [!NOTE]
> ### Reuse uv across runtime provisioning by caching and serializing setup per HOME
> - Adds `_uv_lock` (an `asyncio.Lock`) and `_uv_ready_homes` (a set) to `Runtime.__init__` in [base.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2001/files#diff-0811c040a163eb5a3212208f1e3a6c6388d7690fcc4780c081cc7754036232b9) to serialize uv installation and skip re-running it for already-provisioned HOME directories.
> - Extracts a new `_UV_ENV` shell constant containing the `PATH` and `UV_INSTALL_DIR` exports, so that already-ready environments only export env vars instead of re-running the full installer.
> - Applies the same locking and readiness-cache logic to `_install_in_sandbox` in [launch.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2001/files#diff-989d87123d40d48511f9e1df74594d2e30776e9a99c11eca49d9cc20bf9e2230).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 357c5b1.</sup>
> 
>

---

> [!NOTE]
> **Low Risk**
> Provisioning-only optimization with serialized first-time uv install per HOME; no auth or data-path changes, though concurrent cold starts now queue on the lock.
> 
> **Overview**
> **Reuses a successful `uv` install** within a single runtime instance so later PEP 723 script prep and colocated sandbox server installs do not repeat pip/curl bootstrap when they share the same `HOME`.
> 
> `Runtime` now tracks **`_uv_ready_homes`** (keyed by `env["HOME"]`, with `None` for the default in sandbox installs) and serializes the **first** bootstrap per key with **`_uv_lock`**. That first run still uses **`_ENSURE_UV`**; follow-ups only apply the extracted **`_UV_ENV`** (`PATH` / `UV_INSTALL_DIR`). A failed bootstrap is not marked ready, so the next caller retries. Per-script `uv sync`, content-addressed script paths, and separate server venvs are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 357c5b14e45e84167d8b07e29f1fef3dbfa151c9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/verifiers/v1/mcp/launch.py
+++ b/verifiers/v1/mcp/launch.py
@@ -125,8 +125,12 @@ async def _install_in_sandbox(server: ServerBase, runtime: Runtime) -> str:
     # build, silently running the server against a released (older) API. Pretend the
     # local version so the floor is satisfied by the build we uploaded.
     vf_version = importlib.metadata.version("verifiers")
-    async with runtime._uv_lock if not runtime._uv_ready else contextlib.nullcontext():
-        ensure_uv = _UV_ENV if runtime._uv_ready else _ENSURE_UV
+    async with (
+        runtime._uv_lock
+        if None not in runtime._uv_ready_homes
+        else contextlib.nullcontext()
+    ):
+        ensure_uv = _UV_ENV if None in runtime._uv_ready_homes else _ENSURE_UV
         setup = (
             f'{ensure_uv}; set -e; for t in {root}/*.tar.gz; do tar -xzf "$t" -C {root}; done && '
             f"uv venv {venv} && "
@@ -136,7 +140,7 @@ async def _install_in_sandbox(server: ServerBase, runtime: Runtime) -> str:
         )
         result = await runtime.run(["sh", "-c", setup], {})
         if result.exit_code == 0:
-            runtime._uv_ready = True
+            runtime._uv_ready_homes.add(None)
     if result.exit_code != 0:
         raise ToolsetError(
             f"server {server.server_name!r} install failed in runtime: "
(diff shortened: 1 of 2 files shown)


mikasenghaas (commented):
this only applies to repeated local installs right? i think we might want to skip this optimization

PR #2002, opened 2026-07-14
Warm elastic interception by default

xeophon:
## Overview

Warm the first elastic interception server automatically when the pool starts.

## Change

`ElasticInterceptionPool.start()` now schedules its first server immediately. The first acquire awaits the shared task, while later servers are still created only when the configured multiplex capacity is exhausted.

Warm-up is best-effort: if the speculative task fails, acquisition clears it and falls back to the normal lazy creation path. Individual waiter cancellation remains shielded, and pool teardown cancels and awaits unfinished warm work.

No runner, environment, factory, or configuration flag is required. This default applies to every elastic pool; non-elastic interception types are unchanged.

## Impact

- Hides approximately 1.59 seconds of cold shared-tunnel startup behind concurrent Prime VM setup.
- Removes the first-acquire server startup delay for local elastic pools as well.
- Keeps one initial server and the existing multiplex behavior; an eight-rollout eval still uses one tunnel with the default capacity of 32.
- Makes the optimization consistent across in-process evals, EnvServer, GEPA, and direct elastic-pool users.
- Eagerly allocates one server whenever an elastic-pool serving context is entered, even if no rollout acquires it.

> [!NOTE]
> ### Warm the first interception server on `ElasticInterceptionPool` start by default
> - `ElasticInterceptionPool.start` now launches a background task that pre-creates the first interception server, so the server is ready before the first `acquire` call.
> - `acquire` awaits the warm task (via `asyncio.shield`) on first call, ignoring any errors, then clears the task reference so subsequent acquires proceed without waiting.
> - `stop` cancels the warm task if still running and awaits its completion before delegating to the superclass shutdown.
> - Behavioral Change: pools now eagerly spin up one server at start rather than waiting for the first acquire.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e59aacf.</sup>
> 
>

### DIFF
--- a/verifiers/v1/interception/pool.py
+++ b/verifiers/v1/interception/pool.py
@@ -13,6 +13,7 @@ least-loaded.
 """
 
 import asyncio
+import contextlib
 import logging
 from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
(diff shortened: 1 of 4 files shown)


verifiers/v1/interception/pool.py
```diff
@@ -91,15 +92,26 @@ def __init__(
         self,
         config: ElasticInterceptionPoolConfig | None = None,
         requires_tunnel: bool = False,
+        warm: bool = False,
```
mikasenghaas: lets not make this configurable. and just warm one server by default

mikasenghaas (approved):

PR #1977, opened 2026-07-13
fix(v1): bound null harness MCP connect at 60s

eligotts:
## Summary
- Wrap the null harness's `connect_mcp` in `asyncio.wait_for(..., timeout=60)` so a wedged streamable-HTTP handshake fails loudly instead of eating the whole rollout budget as a 0-turn `harness_timeout`.
- Incidental: this file was already out of format vs current ruff on `main`; touching it applies that format (no behavior change beyond the timeout).

## Test plan
- [ ] Smoke a null-harness eval with MCP tools (happy path still connects)
- [ ] Optional: force a stuck connect and confirm it surfaces as a program error within ~60s

---

> [!NOTE]
> **Low Risk**
> Isolated to the null eval harness startup path; improves failure visibility without touching auth, data, or shared production APIs.
> 
> **Overview**
> **MCP connect in the null harness is capped at 60 seconds** so a stuck streamable-HTTP handshake fails with `asyncio.TimeoutError` instead of hanging until the rollout hits a 0-turn `harness_timeout`.
> 
> `connect_mcp` runs inside `asyncio.timeout(60)` when `mcpServers` is configured; the no-MCP path is unchanged. The implementation uses **`asyncio.timeout` rather than `wait_for`** so MCP/httpx cancel scopes registered on the shared `AsyncExitStack` are torn down in the same task.
> 
> The inline script metadata bumps **`requires-python` to `>=3.11`**, which is required for `asyncio.timeout`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 250a5c8c4f549c88248d7b8414188941a601deaa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Bound null harness MCP connect with a 60s timeout
> Wraps the `connect_mcp` call in [program.py](https://github.com/PrimeIntellect-ai/verifiers/pull/1977/files#diff-d4e93ef3c6afb0d7441cceeef2d60bd8618168a7d72b8d1df78f16dbe0289304) with `asyncio.timeout(60)` so MCP initialization fails with `asyncio.TimeoutError` after 60 seconds instead of waiting indefinitely. Also bumps the script's `requires-python` from `>=3.10` to `>=3.11`, which is required for `asyncio.timeout`.
>
> 
> #### Changes since #1977 opened
>
> - Replaced multi-line comment with shorter comment in main function [250a5c8]
> 
>
> 
>
[...]

PR #2020, opened 2026-07-15
fix(v1): disable openai SDK retries in the null harness program

dumko2001:
## Description

`verifiers/v1/harnesses/null/program.py` creates an `AsyncOpenAI` client without `max_retries=0`. The SDK default retries can re-sample an already-committed model turn, which forks the trace and breaks determinism.

`verifiers/v1/harnesses/default/program.py` already sets `max_retries=0` deliberately for the same reason. This change applies the same single-line setting to the null harness.

This is not the same as #1977 (which bounds MCP connect at 60s) or #1919 (which adds lazy task hooks). Both touch the file but do not disable SDK retries. The rebase preserves every upstream change, including the `>=3.11` pin and the `asyncio.timeout(60)` MCP connect guard.

## Type of Change

- [ ] Documentation update
- [ ] Test improvement

## Testing

`uv run pytest tests/v1/test_envs.py -q` still passes (key-gated tests are skipped without `PRIME_API_KEY`).

## Checklist

## Additional Notes

- Rebased onto the current `PrimeIntellect-ai/verifiers:main` (commit `9b01a8172`), so #1977 is not reverted.
- No documentation, recipe, config, or hardware-support changes were required: the harness docstring/comment already explains the client behavior, and the change is a one-line retry-policy fix.
- Per `AGENTS.md`, no unit tests were added.

### DIFF
--- a/verifiers/v1/harnesses/null/program.py
+++ b/verifiers/v1/harnesses/null/program.py
@@ -101,7 +101,7 @@ async def main() -> None:
         payload = path.read_bytes()
         path.unlink()
         initial = json.loads(payload)
-    client = AsyncOpenAI(base_url=args.base_url, api_key=args.api_key)
+    client = AsyncOpenAI(base_url=args.base_url, api_key=args.api_key, max_retries=0)
     config = json.loads(args.mcp_config or "{}")
     async with AsyncExitStack() as stack:
         # asyncio.timeout, not wait_for: the MCP/httpx cancel scopes entered onto


mikasenghaas:
Thanks a lot for the contribution and for pinpointing the trace fork — that diagnosis was spot on and is exactly what motivated the follow-up.

We went with a slightly different fix in #2036 (merged): rather than disabling SDK retries, the interception server now makes the message graph atomic under retries — it replays a completed turn from a per-request cache, and coalesces a retry that arrives while the first attempt is still in flight onto that same attempt. That keeps retries enabled to ride out transient blips while guaranteeing a retry can never fork the graph.

Closing this in favor of #2036, but the root-cause analysis here was the key insight. Thanks again! 🙏

PR #2033, opened 2026-07-15
fix(serve): run env server on Windows (tcp sockets + tornado for zmq async)

### LINKED ISSUE #2030: Env server fails to start on Windows: ZMQ ipc:// transport is unsupported
### What happens

Running an evaluation on native Windows (`prime eval run ...`, which spins up the
verifiers env server) fails before a single rollout runs. The env-server subprocess
can't bind its internal ZMQ sockets, and the client receive loop errors out too.

Two errors fire:

1. Client receive loop:
   > Proactor event loop does not implement add_reader family of methods required for
   > zmq. zmq will work with proactor if tornado >= 6.1 can be found. Use
   > `asyncio.set_event_loop_policy(WindowsSelectorEventLoopPolicy())` or install
   > 'tornado>=6.1' to avoid this error.

2. Env server (fatal):
   > zmq.error.ZMQError: Protocol not supported (addr='ipc:///tmp/vf-...-responses')

### Root cause

`verifiers/utils/serve_utils.py`:

```python
def make_ipc_address(session_id, name):
    return f"ipc:///tmp/vf-{session_id}-{name.replace('/', '--')}"
```

The router↔worker sockets use ZMQ `ipc://` with a hardcoded `/tmp` path. `ipc://` is
Unix-domain-socket only and isn't supported on Windows, so `EnvRouter.__init__` fails at
`bind()`. There's no tcp fallback or override. Installing `tornado` only silences the
first Proactor error; the `ipc://` bind still fails regardless of the event loop.

### Impact

Anything that starts the env server — `prime eval run`, `prime gepa`, local `prime rl` —
cannot run on native Windows. Env authoring, `env install`, and `env push` are unaffected.

### Environment

- OS: Windows 11
- Python 3.12
- verifiers 0.2.0 (same code on `main`)

### Steps to reproduce

```
prime env init demo        # any trivial SingleTurnEnv
prime env install demo
prime eval run demo -m openai/gpt-4o-mini -p openrouter -n 1 -c 1 --disable-tui
```

Traceback (paths trimmed):

```
verifiers.serve.client.env_client.ZMQEnvClient - ERROR - Unexpected error in receive loop
for env server demo (tcp://127.0.0.1:PORT) (Proactor event loop does not implement
add_reader family of methods required for zmq. zmq will work with proactor if tornado
>= 6.1 can be found. Use `asyncio.set_event_loop_policy(WindowsSelectorEventLoopPolicy())`
or install 'tornado>=6.1' to avoid this error.)
Traceback (most recent call last):
  File ".../zmq/asyncio.py", line 51, in _get_selector_windows
    from tornado.platform.asyncio import AddThreadSelectorEventLoop
ModuleNotFoundError: No module named 'tornado'

During handling of the above exception, another exception occurred:

  File ".../verifiers/serve/client/zmq_env_client.py", line 198, in receive_loop
    msg = await self.socket.recv_multipart()
  ...
RuntimeError: Proactor event loop does not implement add_reader family of methods required
for zmq. ...

Process SpawnProcess-1:
Traceback (most recent call last):
  File ".../verifiers/serve/server/env_server.py", line 145, in run_server
    server = cls(*args, **kwargs)
  File ".../verifiers/serve/server/env_router.py", line 134, in __init__
    self.response_pull.bind(self.response_address)
  File ".../zmq/sugar/socket.py", line 320, in bind
    super().bind(addr)
  ...
zmq.error.ZMQError: Protocol not supported (addr='ipc:///tmp/vf-...-responses')
```

### Workaround

Run under WSL/Linux, where `ipc:///tmp` works.

martian56:
Fixes #2030.

On Windows, `prime eval run` (anything that starts the env server) fails before any
rollout: the server can't bind its internal ZMQ sockets, and the client's receive loop
errors out.

Two causes:

- `serve_utils.make_ipc_address` returns an `ipc:///tmp/...` address. ZMQ `ipc://` is a
  Unix-domain-socket transport and isn't supported on Windows, so `EnvRouter` fails at
  `bind()` with `Protocol not supported`.
- ZMQ's async sockets need `add_reader`, which Windows' default Proactor event loop
  doesn't implement. (On other platforms this never comes up because `uvloop` is used.)

### Changes

- `make_ipc_address` returns a loopback `tcp://127.0.0.1:<free_port>` on Windows and keeps
  `ipc://` on POSIX. Each address is generated once and reused, so a per-call free port is
  fine.
- Broadened the router's socket-file cleanup to `except OSError` (a tcp address has no file
  to unlink).
- Added `tornado>=6.1` as a Windows-only dependency; pyzmq uses it to drive its async
  sockets on the Proactor loop. This mirrors the existing Windows exclusion for `uvloop`.

### Validation (Windows 11, Python 3.12)

Ran a local `SingleTurnEnv` eval end to end. The env server now binds on tcp, becomes
healthy, and rollouts complete:

```
ZMQEnvServer started on tcp://127.0.0.1:49486
Env server ... is healthy
Processing 8 groups (8 total rollouts): 100% ... reward=0.783
```

Before this change it stopped at:

```
zmq.error.ZMQError: Protocol not supported (addr='ipc:///tmp/vf-...-responses')
```

### Note on scope

`v1/serve/pool.py` builds its own `ipc://` address the same way and would need the same
tcp fallback for v1 environments. I scoped this PR to the `serve/server` path that classic
environments use and that I could reproduce and validate; happy to extend it to v1 if you'd
prefer it in one change.

> [!NOTE]
> ### Fix env server on Windows using TCP sockets and tornado for zmq async
> - Updates [`make_ipc_address`](https://github.com/PrimeIntellect-ai/verifiers/pull/2033/files#diff-573ce1050a71bef580f9e082113a30474324c2d0e92c785579dce9d67eb1a4ca) to return `tcp://127.0.0.1:<free_port>` on Windows instead of an `ipc://` path, since Windows does not support Unix domain sockets.
> - Adds a conditional dependency on `tornado>=6.1` for Windows, required for zmq async event loop integration.
> - Broadens the exception caught in [`EnvRouter.teardown`](https://github.com/PrimeIntellect-ai/verifiers/pull/2033/files#diff-5f42e5c44b82aeddf6020de01b59c5480cac9144a3f6ffc79c3ee01683ad92ec) from `FileNotFoundError` to `OSError` when removing IPC paths, preventing shutdown errors on Windows.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2f2965b.</sup>
> 
>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,6 +49,7 @@ dependencies = [
     "aiohttp>=3.9.0",
     "prime-pydantic-config[toml]>=0.3.0.dev86",
     "uvloop>=0.21.0; sys_platform != 'win32' and sys_platform != 'cygwin' and platform_python_implementation != 'PyPy'",
+    "tornado>=6.1; sys_platform == 'win32'",
     "loguru>=0.7.0",
     "tomli-w>=1.0.0",
     "renderers>=0.1.8.dev40",
--- a/verifiers/serve/server/env_router.py
+++ b/verifiers/serve/server/env_router.py
@@ -422,7 +422,7 @@ class EnvRouter:
         for path in self.ipc_paths:
             try:
                 os.unlink(path)
-            except FileNotFoundError:
+            except OSError:
                 pass
 
         self.logger.info("Router shut down")
--- a/verifiers/utils/serve_utils.py
+++ b/verifiers/utils/serve_utils.py
@@ -118,7 +118,13 @@ def walk_decode_tensors(obj: Any, *, to_torch: bool = True):
 
 
 def make_ipc_address(session_id: str, name: str) -> str:
-    """Build an IPC address for inter-process communication."""
+    """Build an inter-process address.
+
+    POSIX uses a ZMQ ipc:// socket. Windows has no ipc:// transport, so fall back to a
+    loopback tcp:// socket on a free port.
+    """
+    if sys.platform == "win32":
+        return f"tcp://127.0.0.1:{get_free_port()}"
     return f"ipc:///tmp/vf-{session_id}-{name.replace('/', '--')}"
 
 


martian56:
@moondev How does this look?

PR #2036, opened 2026-07-15
fix: make the message graph atomic w.r.t. harness retries

mikasenghaas:
## Summary

Fixes the root cause behind #2020: an SDK-level retry re-sends the byte-identical request, and if the interception server re-samples it, a second assistant node is committed under the same parent — forking the trace into a dead-end branch. #2020 avoided this by disabling SDK retries (`max_retries=0`), which also throws away retrying *genuine* transient failures. Instead, this makes the graph atomic under retries so retries stay enabled.

Two guards, keyed on the request body digest (per `RolloutSession`), cover the two ways a retry can duplicate a turn:

- **Replay** — the first attempt already finished: `last_request`/`last_response` cache the served response and replay it verbatim, no re-sampling.
- **Coalesce** — the first attempt is still computing (a slow turn): `inflight` maps the digest to the in-flight attempt's future, and the retry awaits it instead of starting a second inference. aiohttp does not cancel the first handler on client disconnect (`handler_cancellation=False`), so it keeps computing while the retry waits.

A growing conversation never repeats a body, so both guards only ever match a real retry; a failed attempt caches nothing and re-runs normally.

With a slow turn now coalesced rather than re-sampled, retries are safe regardless of the client timeout — so both the `default` and `null` harness programs **restore** SDK retries (drop `max_retries=0`) and use a plain `AsyncOpenAI` client (no inflated `timeout=1800`).

The invariant: harness retries ride out transport/provider blips, and retrying can never fork the message graph — only a fully served attempt populates it.

## Implementation notes

- The body digest is hashed **off the event loop** for large bodies: blake2b runs ~1.7 GB/s and `_MAX_REQUEST_BODY` is 1 GiB (to allow big tool results), so hashing a large body inline would stall every multiplexed rollout; small bodies (≤1 MiB) hash inline.
- `serve()` now takes the vf `Response` and extracts `.raw` at the boundary, so every call site reads `serve(response)`; the cache field is `last_response`.

## Verification

Drove the real `InterceptionServer` with a fake model client that re-samples a *different* message per call (so any duplicate commit surfaces as an extra graph leaf):

- **replay** — a sequential identical (retried) request → **1 model call**, same completion id, 1 leaf, 2 nodes (no fork).
- **coalesce** — two identical requests fired concurrently against a 0.5s inference (retry mid-flight) → **1 model call**, both get the same result, 1 leaf, 2 nodes (no fork).
- **new turn** — a genuinely different request → re-samples (2 calls), graph stays linear.

`uv run pytest tests/v1/` passes (key-gated e2e/env tests skipped without `PRIME_API_KEY`).

## Notes

- Supersedes #2020 — closing that in favor of this.
- Per `AGENTS.md`, no unit tests added (verified with a temporary script).

---

> [!NOTE]
> **Medium Risk**
> Changes core interception request handling and trace commit semantics for all non-streaming harness model calls; incorrect replay/coalesce logic could mask failures or return stale completions, but the design limits matches to byte-identical retries.
> 
> **Overview**
> **Harness SDK retries no longer fork the trace message graph** when the interception server would have re-sampled a byte-identical retried request.
> 
> The interception server **digests each request body** (blake2b, inline for small bodies, off the event loop for large ones) and applies two guards on `RolloutSession`: **replay** completed attempts via `last_request`/`last_response`, and **coalesce** in-flight duplicates by awaiting a shared `inflight` future instead of running a second inference. Successful turns go through a **`serve()`** helper that records the cache and completes waiters; a **`finally`** block clears the in-flight slot and signals `None` on error paths so retries can try again.
> 
> The default harness **drops** the custom `AsyncOpenAI` settings (`max_retries=0`, long timeout) and uses library defaults, since coalescing makes slow turns safe without disabling retries.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit deabb6c7cdf9cb10aea08737107d4f600801a9cc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Make the message graph atomic with respect to harness retries by coalescing identical concurrent requests
> - Adds request-body hashing (blake2b) in [server.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2036/files#diff-84e2f8d027d609e8760ef6d533535ec9d99dfc6384d5ddb6111b4001b2010f35) so the interception server can detect identical retried or concurrent requests.
> - Identical requests that have already completed replay the recorded response immediately; identical in-flight requests coalesce on a shared `asyncio.Future` instead of triggering a redundant model call.
> - Adds `last_request`, `last_response`, and `inflight` fields to `RolloutSession` in [session.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2036/files#diff-0eeafa9506eb0d53c459cf476e308ce72a99d147d2f84979affa04baaee79a66) to support replay and coalescing state.
> - Removes the explicit `timeout=1800.0` and `max_retries=0` from the `AsyncOpenAI` client in [program.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2036/files#diff-4cb67cb50bf655658944c055a7e88bcf71e08b1112043d946c684ac2d51a2d80), reverting to library defaults.
> - Adds a retry loop in [test_e2e.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2036/files#diff-ebbbd919487889487f891b359a756e6c4630e106b16ddaef8ccddc6b900a96e1) for `test_tool_state` that reruns once when a transient `invalid_request` provider error is detected.
>
> 
> #### Changes since #2036 opened
>
> - Removed retry logic for transient `invalid_request` errors in `tests.v1.test_e2e.test_tool_state` [deabb6c]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9c5730c.</sup>
> 
>
>

### DIFF
--- a/tests/v1/test_e2e.py
+++ b/tests/v1/test_e2e.py
@@ -3,13 +3,6 @@
 import pytest
 
 
-def _is_transient_provider_invalid_request(trace) -> bool:
-    return bool(trace.errors) and all(
-        error.type == "ProviderError" and '"code":"invalid_request"' in error.message
-        for error in trace.errors
-    )
-
-
 @pytest.mark.e2e
 async def test_single_turn(run_v1, harness, harness_runtime, tmp_path):
     """Single-turn (echo a short phrase back)."""
@@ -101,19 +94,14 @@ async def test_tool_state(run_v1, harness_runtime, tool_runtime, tmp_path):
         pytest.skip(
             "tool server in a prime sandbox needs prime port exposure (unreachable from host here)"
         )
-    # Prime can intermittently return a transient generic invalid_request on tool calls;
-    # one rerun keeps this smoke test deterministic while still surfacing real regressions.
-    for _ in range(2):
-        (trace,) = await run_v1(
-            "counter-tool-v1",
-            harness="null",
-            harness_overrides={"runtime": {"type": harness_runtime}},
-            output_dir=tmp_path,
-            max_turns=8,
-            taskset_overrides={"task": {"tools": tool_runtime}},
-        )
-        if not _is_transient_provider_invalid_request(trace):
-            break
+    (trace,) = await run_v1(
+        "counter-tool-v1",
+        harness="null",
+        harness_overrides={"runtime": {"type": harness_runtime}},
+        output_dir=tmp_path,
+        max_turns=8,
+        taskset_overrides={"task": {"tools": tool_runtime}},
+    )
     assert trace.errors == []
     assert trace.num_turns >= 2  # at least two tool calls accumulated
     assert trace.reward == 1.0
--- a/tests/v1/test_e2e.py
+++ b/tests/v1/test_e2e.py
@@ -3,6 +3,13 @@
 import pytest
 
 
+def _is_transient_provider_invalid_request(trace) -> bool:
+    return bool(trace.errors) and all(
+        error.type == "ProviderError" and '"code":"invalid_request"' in error.message
+        for error in trace.errors
+    )
+
+
 @pytest.mark.e2e
 async def test_single_turn(run_v1, harness, harness_runtime, tmp_path):
     """Single-turn (echo a short phrase back)."""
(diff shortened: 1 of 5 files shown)


tests/v1/test_e2e.py
```diff
@@ -94,14 +101,19 @@ async def test_tool_state(run_v1, harness_runtime, tool_runtime, tmp_path):
...
-        "counter-tool-v1",
-        harness="null",
-        harness_overrides={"runtime": {"type": harness_runtime}},
-        output_dir=tmp_path,
-        max_turns=8,
-        taskset_overrides={"task": {"tools": tool_runtime}},
-    )
+    # Prime can intermittently return a transient generic invalid_request on tool calls;
+    # one rerun keeps this smoke test deterministic while still surfacing real regressions.
+    for _ in range(2):
```
samsja: should we put the 2 in and CONSTANT

samsja (approved):

PR #2038, opened 2026-07-15
feat(v1): taskset system_prompt handoff for GEPA → eval/train

Ziems:
## Summary
- Add Judge-style `TasksetConfig.system_prompt` / `system_prompt_file` (mutually exclusive) plus `resolve_system_prompt()`, and wire reverse-text / lean / textarena / color-codeword `load()` to honor them (override taskset defaults when set).
- Drop `GEPAConfig.initial_prompt`; seed from loaded tasks instead, and write `best_system_prompt.txt` for reuse.
- Document the eval handoff (`--taskset.system-prompt-file`) and update the evaluate-environments REFERENCE.

## Test plan
- [ ] `uv run gepa reverse-text-v1 --dry-run` (no `initial_prompt` in dumped config)
- [ ] Short GEPA run writes `outputs/.../best_system_prompt.txt`
- [ ] `uv run eval reverse-text-v1 -n 2 --taskset.system-prompt-file <that file>` uses the override
- [ ] `uv run eval reverse-text-v1 -n 2` still uses the shipped default prompt
- [ ] Reject setting both `--taskset.system-prompt` and `--taskset.system-prompt-file`
- [ ] Lean still loads with default prompt when neither override is set

Made with [Cursor](https://cursor.com)

> [!NOTE]
> ### Add system prompt handoff from GEPA to eval/train via `TasksetConfig`
> - Adds `system_prompt` and `system_prompt_file` fields to `TasksetConfig` (mutually exclusive), with a `resolve_system_prompt` helper now exported from the `verifiers.v1` public API.
> - Updates `color_codeword_v1`, `reverse_text_v1`, `lean`, and `textarena` tasksets to honor these config overrides, falling back to their baked-in defaults.
> - GEPA runs now write the best system prompt to `best_system_prompt.txt` in the run output directory, making it easy to reuse via `--taskset.system-prompt-file`.
> - Removes `GEPAConfig.initial_prompt` and the `initial_prompt` parameter from `resolve_gepa_seed_prompt`; users must now supply prompts via `--taskset.system-prompt` or `--taskset.system-prompt-file`.
> - Risk: callers that passed `initial_prompt` to `GEPAConfig` or `resolve_gepa_seed_prompt` will break.
>
> 
> #### Changes since #2038 opened
>
> - Added `apply_system_prompt` method to `Taskset` class to compute and apply effective system prompts, introduced `iter_tasks` method to yield tasks with config-layer system prompt overlays applied, and extended `select` method with `apply_config` parameter to control whether config-layer system prompts are applied during task selection [902c959]
> - Modified `resolve_gepa_seed_prompt` function to accept `TasksetConfig` and prioritize config-layer system prompt overrides by calling `resolve_system_prompt`, falling back to the first task with non-None `TaskData.system_prompt`, and updated `run_gepa` function to call `env.taskset.select` with `apply_config=False` before computing the seed prompt [902c959]
> - Added `system_prompt` method to `Harness` base class returning harness-owned framing with default of None, added `compose_system_prompt` method to compose harness framing with task-side system prompt joined by double newlines, and implemented `DefaultHarness.system_prompt` to construct harness-level framing from bash, edit, and search prompts [902c959]
> - Removed per-load resolution of config-layer system prompt overrides from `ColorCodewordTaskset.load`, `ReverseTextTaskset.load`, `LeanTaskset.load`, and `TextArenaTaskset.load` methods, with each now setting `TaskData.system_prompt` to their respective baked-in constants when constructing tasks [902c959]
> - Changed `EnvServer.__init__` to source tasks from `self.env.taskset.iter_tasks()` instead of `self.env.taskset.load()` to ensure config-layer system prompt overlays are applied when serving tasks [902c959]
> - Updated documentation in `docs/v1/evaluation.md`, `docs/v1/gepa.md`, `skills/evaluate-environments/references/REFERENCE.md`, `verifiers/v1/cli/gepa.py`, and `verifiers/v1/gepa/config.py` to clarify the distinction between config-layer system prompts and baked-in `TaskData.system_prompt`, describing GEPA as optimizing the config-layer system prompt and explaining harness handling and config replacement semantics [902c959]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized fb61777. (Automatic summaries will resume when PR exits draft mode or review begins).</sup>
> 
>
>

### DIFF
--- a/environments/reverse_text_v1/reverse_text_v1/taskset.py
+++ b/environments/reverse_text_v1/reverse_text_v1/taskset.py
@@ -46,13 +46,12 @@ class ReverseTextTaskset(vf.Taskset[ReverseTextTask, ReverseTextConfig]):
         from datasets import load_dataset
 
         rows = load_dataset(self.config.dataset_name, split=self.config.dataset_split)
-        system_prompt = vf.resolve_system_prompt(self.config) or SYSTEM
         return [
             ReverseTextTask(
                 ReverseTextData(
                     idx=i,
                     prompt=row["prompt"],
-                    system_prompt=system_prompt,
+                    system_prompt=SYSTEM,
                     answer=row["prompt"][::-1],
                 ),
                 self.config.task,
--- a/environments/reverse_text_v1/reverse_text_v1/taskset.py
+++ b/environments/reverse_text_v1/reverse_text_v1/taskset.py
@@ -46,12 +46,13 @@ class ReverseTextTaskset(vf.Taskset[ReverseTextTask, ReverseTextConfig]):
         from datasets import load_dataset
 
         rows = load_dataset(self.config.dataset_name, split=self.config.dataset_split)
+        system_prompt = vf.resolve_system_prompt(self.config) or SYSTEM
         return [
             ReverseTextTask(
                 ReverseTextData(
                     idx=i,
                     prompt=row["prompt"],
-                    system_prompt=SYSTEM,
+                    system_prompt=system_prompt,
                     answer=row["prompt"][::-1],
                 ),
                 self.config.task,
--- a/docs/v1/evaluation.md
+++ b/docs/v1/evaluation.md
@@ -33,6 +33,9 @@ The output from evaluations are written into `outputs/<taskset>--<model>--<harne
 - `model` — the model id to evaluate, e.g. `nvidia/NVIDIA-Nemotron-3-Ultra-550B-A55B`
 - `sampling` — generation params passed to the model, e.g. `sampling.temperature`
 - `taskset.id` / `harness.id` — pick the taskset and harness
+- `taskset.system_prompt` / `taskset.system_prompt_file` — override the task
+  system prompt (mutually exclusive; honored by tasksets that use them in `load()`,
+  e.g. after a GEPA run's `best_system_prompt.txt`)
 - `num_tasks` — how many tasks to evaluate. Not setting a value means all tasks; an
   infinite taskset (a procedural generator, e.g. `wordle-v1`) requires it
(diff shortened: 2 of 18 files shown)


environments/reverse_text_v1/reverse_text_v1/taskset.py
```diff
@@ -46,12 +46,13 @@ def load(self) -> list[ReverseTextTask]:
         from datasets import load_dataset
 
         rows = load_dataset(self.config.dataset_name, split=self.config.dataset_split)
+        system_prompt = vf.resolve_system_prompt(self.config) or SYSTEM
```
mikasenghaas: i dont think resolution should happen in user space

mikasenghaas: it seems to be like we have diff places that might have
- configurable taskset sys prompt (file, e.g. from gepa)
- built-in taskset sys prompt
- built-in task sys prompt
- built-in harness sys prompt

i think for task <> harness resolution we will want to append, harness, then task.

for the task sys prompts, we should have clear ordering. im guessing 

config sys > task sys > taskset sys

wdyt?

Ziems: image

@mikasenghaas Is this roughly what you have in mind? if so, yeah I directionally agree but have no strong opinions on this

Ziems:
Superseded by https://github.com/PrimeIntellect-ai/verifiers/pull/2171 — re-implements this cleanly on current main (multi-agent API / client-owned env server / Episode adapter), dropping the obsolete iter_tasks/server plumbing that made this branch unmergeable.

PR #1939, opened 2026-07-07
feat(v1): multi-agent api

hallerite:
## Summary

Multi-agent becomes a first-class shape of v1, built from two pieces: an **`Agent`** you run like a function, and an **`Env`** that writes the control flow between agents as plain Python. An episode — one run of the env on one task — is a flat `list[Trace]` linked by a per-trace episode stamp, persisted and served whole. One `[env]` block composes three separately-chosen concerns — the **taskset** (what to solve), each agent's **harness** (how that LLM meets the world), the **env** (how agents interact):

```
uv run eval gsm8k-v1 --env.id best-of-n --env.agent.harness.id codex
```

`--env.id` empty keeps the taskset's own story (its recipe env, else `SingleAgentEnv`) — a plain eval is fully backward compatible.

### The Agent

An `Agent` is (harness × model × runtime policy), built from its `AgentConfig` alone and runnable on any task:

```python
import verifiers.v1 as vf

solver = vf.make_agent(vf.AgentConfig(model="z-ai/glm-5.2"))
trace = await solver.run(vf.Task(vf.TaskData(prompt="What is 2+2?")))
```

- The config carries everything declarative: harness (None = the taskset's default, else `bash`), model, endpoint, sampling, per-stage timeouts, turn/token caps, and whole-run `retries`. Live resources are injected and borrowed, never configured: `make_agent(config, client=..., interception=...)`.
- Task × agent fit validates per run, on the task the agent actually receives (`utils/compile.py`): a task needing MCP on a harness without it, or a `NEEDS_CONTAINER` task on subprocess, refuses before any work. An env-minted task carries its own needs — which is why a bare verdict task pairs a judge with *any* taskset.
- Retries are per-agent and the default retry unit: `Agent.run` reruns its rollout (with backoff) while the trace ends with a retryable error — `--env.judge.retries.max_retries 2` retries a flaky judge without re-burning the solver.

### The Env

`Env` is abstract; every run gets a concrete subclass — plain tasksets resolve to the bundled `SingleAgentEnv`, a taskset package can export its own recipe env, and `--env.id` pairs a reusable env with any taskset. An env declares its config as an `EnvConfig` subclass — **each `AgentConfig` field is an agent, the field name is its name** — and writes three hooks:

```python
class DebateConfig(vf.EnvConfig):
    pro: vf.AgentConfig = vf.AgentConfig()
    con: vf.AgentConfig = vf.AgentConfig()
[...]

PR #2049, opened 2026-07-16
feat(v1): interleaving agents

hallerite:
## Summary

This PR completes the multi-agent API from #1939 with the exchange half: `Agent.interaction(task)` holds one rollout open turn-by-turn, while `Env.run()` writes the control flow between participants as plain Python. There is one mechanism for scripted users, modeled users, game engines, self-play, and human-in-the-loop conversations; the caller of `turn()` is the run's user.

~~~python
async with agents.player.interaction(task) as interaction:
    segment = await interaction.turn()  # a prompted task speaks first
    while not segment.terminated:
        game.step(segment.last_reply)
        if game.done:
            break
        segment = await interaction.turn(game.feedback())
~~~

The old `vf.User` server path is removed. User behavior now lives where the rest of the episode does: in env control flow.

### The interaction

`Agent.interaction()` is a real rollout held open across harness segments:

- `turn(message)` runs exactly one segment and returns a `Segment`. `Segment.messages` contains the assistant/tool messages produced by that segment; `last_reply` is sugar for its final assistant text. `terminated` means the run ended instead of producing another segment.
- A prompted task opens with a bare `turn()`. A prompt-less task is opened by `turn(message)`. `mask_prompt=True` keeps a scenario on the task for scoring while hiding it from the assistant wire, so the caller opens the exchange.
- The live trace, limits, stops, hooks, and scoring are unchanged. Leaving the context closes the rollout as `user_closed` and joins it to the episode.
- Eval concurrency is acquired around active lifecycle work—open, each segment, and close—not while the env is deciding the next turn. Two interactions can therefore interleave at `-c 1` without bypassing the concurrency bound.
- Whole-run retries stay on `Agent.run()`. A caller-driven interaction is not replayable, so retries do not apply to `interaction()`.

### The harness seam

A harness still owns the agent program; it now runs one segment at a time:

- `launch(..., data)` runs the opening segment from that segment's wire view of `TaskData`.
- `resume(..., data, messages)` runs the next segment. The default rebuilds the accumulated transcript and relaunches a Messages-capable harness; `SUPPORTS_RESUME` advertises that contract. Harnesses with native state can override it—Codex uses `codex exec resume --last` in a per-trace home.
- Kimi Code, Pi, and Pool run through the public `vf.ACP` adapter. Their ACP session is persisted across process launches, including MCP access after resume.
- Optional harness `cleanup(trace, runtime)` removes per-rollout state after scoring.

The interception server remains a pure model boundary: it records model turns. The exchange itself lives one layer up, between harness segments.

### Environments

The bundled `user-sim` env is the complete modeled-user mechanism: two ordinary agent interactions relayed by `Env.run()`. The user receives the task prompt as its scenario; the assistant receives the same task with the prompt masked, so existing rewards and judges still score the real row.

`kuhn-poker-v1` demonstrates two-agent self-play with host-side rules and zero-sum rewards. TextArena, OpenEnv, alphabet-sort, and color-codeword use the same interaction loop for scripted/game-driven users. Tools remain inside a harness segment, so a tool-using assistant composes with any of these user loops without a second protocol.

## Removed / breaking

- Removes `vf.User`, `UserConfig`, `Task.user`, `serve_user` / `user_respond`, `UserError`, user-server placement and lifecycle code, and the `vf init -U` scaffold.
- Removes `SUPPORTS_USER_SIM` and the interception server's user-consult machinery.
- Adds the public `Interaction` and `Segment` types plus `Agent.interaction()`.
- Custom multi-turn envs should move user behavior into `Env.run()` and drive `agents.<name>.interaction(task)` directly.

## Verification

- Full repository suite, Ruff, ty, and all pre-commit hooks pass on the merged tree.
- E2E coverage exercises direct interactions, modeled-user relay with tools, game/self-play loops, context preservation across resumed segments, and MCP calls after resume for Kimi Code, Pi, and Pool.
- The ACP matrix runs container-required harnesses in Docker and includes a Pool-on-Prime placement.

## Credits

Built on the run half in #1939. ACP harness support is integrated from #2046.

> [!NOTE]
> ### Replace user simulator architecture with interleaving agent interactions driven by `Env.run()`
> - Removes the `vf.User` simulator concept throughout the framework; multi-turn conversations are now driven by `Agent.interaction()`, an async context manager that yields an `Interaction` object and advances episodes turn-by-turn via `Interaction.turn(message)`.
> - Adds `Harness.SUPPORTS_RESUME` flag and a default `resume()` implementation that rebuilds conversation history and relaunches; `BashHarness`, `NullHarness`, `CodexHarness`, `KimiCodeHarness`, `PoolHarness`, `RLMHarness`, and `PiHarness` all declare resume support.
> - Introduces `UserSimEnv`, an env that models the user as a second `Agent` with its own trace, alternating turns between assistant and simulated user until the user emits a done marker.
> - Adds a two-player `KuhnPokerEnv` environment coordinating two agent interactions with legal-action validation and zero-sum payoff recording.
> - Converts `TextArenaEnv`, `OpenEnvEnv`, `AlphabetSortEnv`, and `ColorCodewordEnv` to env-driven interaction loops, removing their colocated user simulator servers.
> - `PiHarness`, `KimiCodeHarness`, and `PoolHarness` now run via ACP (`agent-client-protocol`) with session persistence for resume support.
> - Risk: `vf.User`, `UserConfig`, `UserError`, `serve_user`, `Dialect.extend`, and the `-U`/`--add-user` CLI flag are removed and will break any existing code or configs that reference them.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 35911a2.</sup>
> 
>


docs/v1/harnesses.md
```diff
@@ -20,8 +21,8 @@ class MyHarness(Harness[MyHarnessConfig]):
     APPENDS_SYSTEM_PROMPT = True
     # When the taskset exports a toolset, they are added as MCP. To show that your harness is able to install MCPs, you have to set this flag to true.
     SUPPORTS_MCP = True
-    # Allow the task to simulate a user and thus drive the execution of the harness
-    SUPPORTS_USER_SIM = True
+    # Allow stateless chat continuation by relaunching on a Messages prompt.
+    SUPPORTS_MESSAGE_PROMPT = True
```
mikasenghaas: will this become `SUPPORTS_ACP` -- the whole message prompt thingy is janky iirc

hallerite: no, that would be the wrong distinction. ACP is just a standardized way of implementing this, but we could implement this for harnesses that don't support ACP without ACP as well. this capability that we are setting here is about resuming the conversation from a list of messages basically

verifiers/v1/agent.py
```diff
@@ -294,10 +401,78 @@ async def _run_once(
...
+    async def chat(
+        self,
+        task: Task,
+        *,
+        runtime: Runtime | None = None,
+        tools: Mapping[str, SharedToolServer] | None = None,
+        mask_prompt: bool = False,
+        on_trace: Callable[[Trace], None] | None = None,
+    ) -> AsyncIterator[ChatSession]:
+        """Converse with this agent turn-by-turn: a full rollout of `task` where
```
eligotts: shorten docstring?

skills/create-environments/SKILL.md
```diff
@@ -187,13 +186,16 @@ Choose placement from the tool's lifetime and filesystem needs:
...
...ent>.*`) — the field name is the agent's name, the only naming site, and per-run caps (turns, tokens, stage timeouts, retries) are agent fields. A declared pin is its author default; an unpinned agent runs the taskset's default harness, and its model context defaults to the run's own. Task x agent fit validates per run, on the task each agent actually receives (an env-minted task carries its own `tools`/`NEEDS_CONTAINER`, so a bare verdict task pairs with any taskset). Then write `run(task, agents)` (imperative control flow, returning nothing — every finished run joins the episode automatically, stamped with its standing; a multi-turn exchange is an interaction, `agents.<name>.interaction(task)`, one `turn()` per harness segment), and optionally `setup(agents)` (env-hardcoded standing, e.g. `agents.judge.trainable = False`) and `finalize(task, episode)` (sibling-dependent judgement over `episode.traces`, via `record_reward`/`record_metric`; `trace.agent_name` names each agent). Before writing one, check the bundled envs (`--env.id best-of-n | agentic-judge | user-sim`) and the reference implementations (`environments/code_golf_v1`, `environments/kuhn_poker_v1`). See docs/v1/env.md.
```
eligotts: this is unrelated to the pr and just reverted to what we had prior, idk why

mikasenghaas (approved):

PR #2109, opened 2026-07-23
feat(v1): configurable agentic-judge grading policy, uploads, and reward weights

mikasenghaas:
## Summary

Makes the built-in `agentic-judge` env a complete out-of-the-box judge, configurable end-to-end from TOML/CLI so domain-specific setups (SWE, search, ...) are pure `.md` + `.toml` artifacts:

- **One box, shared** — the box is provisioned once from the solver's runtime policy (`--env.solver.runtime.type docker|prime`, checked at construction); the solver plays the task in it and the judge then inspects the work exactly as the agent left it. The judge pins only its model; its runtime policy is unused. (Borrowed boxes are never retried into, so per-agent rollout retries don't apply; episode retries still do.)
- **`[env.task]` (`JudgeTaskConfig`)** — the judge's minted task:
  - `prompt`: grading-policy override, one field — a value ending in `.md`/`.txt` is read from disk, anything else is inline policy text (a missing file fails at env construction). A policy replaces only the judging instructions: the verdict contract and workspace note are always appended, so it can never break verdict parsing. `{prompt}` interpolates the solver task's prompt; if absent, the task statement is appended.
  - `hint`: optional hints injected as their own section (inline text or a `.md`/`.txt` file) — task-family pointers into the trace or box, e.g. where the reference answer lives for math, or "diff the repo / read `info.patch`" for SWE.
  - `rubric`: criteria the judge grades against — a `.toml`/`.json` `criteria` file in the plugged rubric judge's format (`{name, text, weight=1, choices worst→best}`; logic mirrored, not imported), defaulting to a single built-in `solved` criterion. The judge writes `{"verdicts": [{"name", "reason", "verdict"}, ...]}` to `/tmp/verdict.json`; `finalize` validates strictly (name set must match, only on-menu answers), records `judge/<name>` metrics on the solver trace, and rewards the weighted mean.
  - The solver's COMPLETE raw trace record always lands at `/tmp/trace.json` (not configurable) — messages, tool calls, `trace.info` artifacts (e.g. a captured patch), and the task row with whatever scores/reference material it carries; nothing is stripped. Independence is prompted, not enforced: the workspace note tells the judge recorded scores and references are context, never the standard, and warns that the record can be very large — peek selectively, never dump it whole. A policy that needs a record field as a file instructs the judge to extract it.
[...]

PR #2144, opened 2026-07-27
feat(v1): grade in an isolated box, with Harbor-native artifacts

rasdani:
Grades in a second sandbox instead of the one the agent worked in, so a policy under RL pressure can't reach the grading machinery. Only what the task declares crosses over.

Supersedes #2067 and unblocks the judge-env port @mikasenghaas asked for there, now that multi-agent has landed (#1939).

**Opt-in.** `agentic-judge` still defaults to `shared`. Isolation is worth having only once a taskset publishes its evidence somewhere that travels — see *Adoption* below.

## Usage

A SWE task produces its delta in `finalize` and declares nothing — anything written to `/logs/artifacts/` is collected by convention:

```python
class SweTask(vf.Task[SweData]):
    async def finalize(self, trace: vf.Trace, runtime: vf.Runtime) -> None:
        await vf.capture_patch(
            trace, runtime, self.data.base_commit,
            write_path=f"{vf.CONVENTION_DIR}/patch.diff",
        )
```

Anything outside the convention dir is declared on the task row:

```python
class SweData(vf.TaskData):
    artifacts: list[vf.Artifact] = [vf.Artifact(source="/work/report", exclude=[".git"])]
```

A Harbor `task.toml` is read as-is:

```toml
artifacts = ["/work/report.json"]

[[verifier.collect]]
command = "pg_dump -U postgres app > /logs/artifacts/db.sql"
```

Run it:

```bash
uv run eval <taskset> --env.id agentic-judge --env.topology isolated \
  --env.solver.runtime.type docker
```

An env composes the sandboxes itself with the two primitives:

```python
async with agents.solver.provision(task) as box:
    solution = await agents.solver.run(task, runtime=box)
    collected = await vf.collect(box, task.data.artifacts)      # barrier

async with agents.judge.provision(task) as judge_box:           # same image
    await vf.restore(judge_box, collected)
    await agents.judge.run(JudgeTask.from_trace(solution, cfg), runtime=judge_box)
```

## Design

`Trace.info` is the durable record (`patch`, `verdict`). It is not a transport channel — though an agentic judge does receive the whole serialized trace, `info` included, at `/tmp/trace.json`. `/logs/artifacts/` is transport: box → host → box, discarded after restore.

`Task.finalize` is the producer hook — it already means "runtime live, agent done, before scoring mutates anything", which is Harbor's collect-hook moment, so `[[verifier.collect]]` maps onto it rather than needing a new stage. `collect()` returning is the barrier; the env owns sandbox topology.

The grading sandbox boots from the task's own image, so only the delta travels. No `Task.scoring_runtime` hook. `runtimes/`, `rollout.py`, `agent.py` and `env.py` are untouched.

## Harbor

`artifacts = [...]` (string and object form, relative sources resolved against the runtime workdir, `exclude` honored) and `[[verifier.collect]]`. Two deliberate divergences from `harbor run`:

- a failing collect hook fails the rollout — here the output is a grading input, not observability, and a silently absent file makes the verifier score a stale state
- `destination` is inert; it places files in Harbor's host trial directory, which verifiers has no equivalent for

Rejected at load: sidecar `service`, `[verifier].user`, explicit `[verifier.environment]` image.

## agentic-judge

`--env.topology shared|isolated`, defaulting to `shared` — what #2109 measured. Under `isolated` the judge gets its own sandbox and the workspace note changes with it: a judge told it stands in the agent's workspace when it stands in a fresh one reads an unmodified tree as failure.

An unpinned judge runtime inherits the solver's, matching how `harness`/`model`/`sampling` already resolve for unpinned seats.

`capture_patch` now distinguishes its two failure modes: the sandbox answering and git refusing (agent's own environment — records `patch_error`, still scores) from the sandbox not answering (raises, since a zero there says only that our infrastructure failed). Telling them apart needs a liveness probe, because `DockerRuntime.run` returns `docker exec`'s non-zero result rather than raising.

## Adoption

Nothing in `research-environments` publishes to the sandbox yet — `write_path` is new here. Under isolation today:

- the seven `capture_patch` tasksets work, because their judge hints point at `info.patch` in the trace record, which travels
- the three plain-Harbor tasksets would not: their hints open with *"Reconstruct the agent's change from the box: `git status`, `git diff`, `git log` in `/testbed`"*, which under isolation is a pristine checkout

So adopt per taskset, once its evidence travels.

## Verification

`ruff`, `ty check verifiers`, 909 non-e2e tests. Against real docker sandboxes, via scratch scripts (not committed, per `AGENTS.md`):

- `collect`/`restore` — 11 cases: convention sweep, declared paths, `exclude`, symlink clobber, strict missing-source, relative-source resolution, over-cap, subprocess refusal
- the isolated `run()` composition — 8 cases with the two agent *runs* stubbed and everything else real: two distinct sandboxes, solver's torn down before the judge's is provisioned, artifact restored at its original path, trace record uploaded, judge sandbox built from the solver's image
- `capture_patch` attribution — healthy capture, broken `.git` records and continues, dead sandbox raises

**No eval has been run.** The isolated path has never executed with a real model or on Prime. On Prime I'd expect judge-sandbox tunnel reachability to break first: it is remote, and nothing has previously provisioned a second remote sandbox needing the interception server within one episode.

## Notes

- BusyBox `tar` has no `-r`, so collection builds one archive per source.
- `restore()` refuses the subprocess runtime — extraction writes to absolute paths, which there is the host filesystem.
- Archives are not validated. An agent that replaces `tar` in its own sandbox can place arbitrary files in the grading sandbox; our own `tar -c` cannot produce an escaping archive, since it does not recurse into symlinked directories.
- Open: `MAX_ARTIFACT_BYTES` is 32 MB on the same-image/delta-only assumption. #2067 budgeted 256 MB for full-tree transfer.

`artifacts.py` is 166 lines.

> [!NOTE]
> ### Add isolated grading topology to `AgenticJudgeEnv` with artifact transport
> - Adds a `topology` field (`'shared'`|`'isolated'`) to `AgenticJudgeEnvConfig`; in `'isolated'` mode the solver runs in one container, artifacts are collected, and a fresh judge container is provisioned to grade using only those artifacts.
> - Introduces [artifacts.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2144/files#diff-694678820f26307a4dc522f8ed8efb10a7aa2182e346fe8f60bb8db9fe1bc42d) with `collect` and `restore` functions that tar declared artifact sources (plus a convention dir sweep) out of the solver box and extract them into the judge box, enforcing a per-rollout size budget.
> - Extends `capture_patch` in [git.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2144/files#diff-88c90ce8263f9107efe7f9e696d8039d388f4d236655604bf00c925b2ec50fa7) to write the patch to an in-sandbox file and unstage specified paths before diffing; adds `snapshot_untracked` helper.
> - Harbor tasks now parse `[[verifier.artifact]]` and `[[verifier.collect]]` entries from `task.toml` into `TaskData.artifacts` and run collect hooks during `finalize`, failing the rollout on hook error.
> - Risk: a missing declared artifact or total artifact size exceeding `MAX_ARTIFACT_BYTES` fails the entire rollout; `restore` raises `ArtifactError` if called on a subprocess runtime.
>
> 
> #### Changes since #2144 opened
>
> - Replaced convention-based artifact collection with Harbor-native artifact restoration between isolated runtimes [9ac6daf]
> - Replaced `topology` configuration with `share_runtime` boolean to control runtime isolation in agentic judge environment [9ac6daf]
> - Removed `ArtifactError` exception class and converted all artifact-related error handling to raise `RuntimeError` [9ac6daf]
> - Increased default timeout for collect hooks and removed exception wrapping in git patch capture [9ac6daf]
> - Removed documentation sections describing legacy artifact collection patterns and topology differences [9ac6daf]
> - Replaced module-level workspace note constants in `JudgeTask` with a unified note referencing `vf.ARTIFACTS_DIR` and `TRACE_FILE`, removing conditional variants based on sandbox type [960d45b]
> - Modified `HarborTask.finalize` method to execute collect hook commands with an empty environment mapping instead of the verifier-specific environment [132d4b3]
> - Fixed `AgenticJudgeEnv.__init__` to preserve the judge agent's configured runtime when `share_runtime` is disabled, removing the previous logic that forcibly overrode the judge runtime to the solver's runtime when the judge was configured as a `SubprocessConfig` [4228c44]
> - Added runtime-aware workspace instructions to `JudgeTask.from_trace` by introducing three module-level constants (`_RECORD_NOTE`, `SHARED_WORKSPACE_NOTE`, `ISOLATED_WORKSPACE_NOTE`) and selecting between `SHARED_WORKSPACE_NOTE` and `ISOLATED_WORKSPACE_NOTE` based on the `share_runtime` argument during prompt assembly [4228c44]
> - Clarified documentation for artifact handling in `TaskData.artifacts` field docstring and reworded comments in the restore function to emphasize that artifacts represent paths collected from one runtime and restored in another, with explicit refusal to restore into subprocess runtime [4228c44]
> - Restructured `Trace` schema to version 1 with agent-scoped fields, auto-populated metadata, and renamed error/run recording APIs [fef21cd]
> - Replaced `EnvServerConfig` with separate `env`, `serve`, and `legacy` configuration blocks in `EvalConfig` and `ServeConfig` [fef21cd]
> - Renamed all harness timeout references to agent timeout and changed rollout deadline expiry to fail with `HarnessError` instead of stopping with timeout condition [fef21cd]
> - Renamed `EnvConfig.max_concurrent` to `max_concurrent_agents` with default 1 and moved episode agent concurrency gating into `Env` implementation [fef21cd]
> - Migrated Harbor taskset parsing to use Harbor's native `Task` model with structured environment fields and MB-unit resource extraction [fef21cd]
> - Changed `JudgeConfig` to accept only file-based prompts via `prompt` field, removing inline string and `prompt_file` options [fef21cd]
> - Added `TasksetConfig.system_prompt` field and `Task.with_system_prompt` method to enable taskset-level prompt overrides with GEPA best prompt export [fef21cd]
> - Reimplemented `is_installed` function to use `importlib.metadata.version` instead of subprocess-based package detection [fef21cd]
> - Moved pool and serving configuration types from `verifiers.v1.configs.cli.env` to `verifiers.v1.configs.serve` and introduced `narrow_taskset_config` helper for CLI resolution [fef21cd]
> - Created `WikiSearchJudge` class with environment-specific prompt file and config for wiki-search taskset [fef21cd]
> - Replaced manual background job polling in `PrimeRuntime.run` with single call to `AsyncSandboxClient.run_background_job` and preserved additional `server_kwargs` in `serve_env` conversion [fef21cd]
> - Added `validate_pairing` call in `AgenticJudgeEnv` initialization to enforce judge, task, and runtime compatibility [fef21cd]
> - Updated dependency versions for `renderers` to >=0.1.9.dev9, `harbor` to 0.20.0, and default Harbor version in `Terminus2HarnessConfig` to '0.20.0' [fef21cd]
> - Changed `JudgeTaskConfig` to accept `prompt` and `hint` fields as string, `Path`, or explicit `TextFile` objects, and modified `JudgeTaskConfig._resolve` method to treat bare strings as inline text rather than using file extension heuristics to determine whether to read from disk [96f2b59]
> - Added `TextFile` to the public exports of the `verifiers.v1.envs.agentic_judge` module [96f2b59]
> - Updated test configuration for the judge task to pass `task.prompt` as an object with a `path` field instead of a string path, and added a `task.hint` string field [96f2b59]
> - Added token budget limit to end-to-end test execution [858ac72]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 615efbb.</sup>
> 
>
> 

---

> [!NOTE]
> **High Risk**
> Changes grading topology, cross-runtime file transport, Harbor finalize strictness, and patch-capture failure attribution—areas that directly affect scores and rollout success.
> 
> **Overview**
> Adds **`vf.collect` / `vf.restore`** and **`TaskData.artifacts`** so declared paths (plus optional `/logs/artifacts/`) can leave the agent sandbox and be replayed at the same absolute paths in another runtime, with a per-rollout size cap and strict errors for missing declared sources.
> 
> **`agentic-judge`** gains **`share_runtime`** (default **true**). When **false**, the solver runs alone, artifacts on **`trace.state`** feed a **fresh judge container** via **`JudgeTask.setup`**, and the judge prompt switches to an **isolated-workspace** note instead of “same box as the agent.” Judge **`prompt` / `hint`** config can be inline strings, paths, or **`TextFile`**.
> 
> **Harbor** loads **`artifacts`** and **`[[verifier.collect]]`** from `task.toml`, runs collect hooks in **`HarborTask.finalize`** (failures fail the rollout), then stores collected bytes on trace state. Unsupported verifier sidecar settings are rejected at parse time.
> 
> **`capture_patch`** can **`ignore`** pre-agent untracked paths (**`snapshot_untracked`**), optionally **`write_path`** into the artifact convention dir, and treats **dead sandbox** vs **git failure** differently (**`SandboxError`** vs **`patch_error`**).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 858ac7292f268d23780924a21410ff8323f4c7e9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/v1/harbor.md
```diff
@@ -87,10 +87,21 @@ added automatically; evaluator-provided `allow` entries add exceptions and `bloc
...
+
+Two deliberate differences from `harbor run`:
+
+- **A failing collect hook fails the rollout.** Harbor logs it and carries on, because there the output is observability; here it is a grading input, and a silently absent file makes the verifier score a stale state.
+- **`destination` has no effect.** It positions a file in Harbor's host trial directory; verifiers has no trial directory (the trace is the record), and Harbor never lets `destination` affect verifier-side placement.
+
+Sidecar `service` entries, `[verifier].user`, and an explicit `[verifier.environment]` image are rejected at load. The grading box is built from the task's own image, so only the agent's delta has to travel; a different verifier image would need the whole working tree copied across.
```
rasdani: delete

docs/v1/tasksets.md
```diff
@@ -224,6 +224,37 @@ class JudgeTraceTaskset(vf.Taskset[JudgedTask, SetConfig]):
 
 To override the judge model, set `env.taskset.task.judge.model` in your config (it is a string).
 
+## Grading artifacts
+
+An environment can grade in a second box rather than the one the agent worked in, so nothing the agent did to its environment can reach the grader. Only what the task declares crosses over.
```
rasdani:
```suggestion
An environment can grade in a second sandbox rather than the one the agent worked in, so nothing the agent did to its environment can reach the grader. Only what the task declares crosses over.
```

verifiers/v1/envs/agentic_judge/env.py
```diff
@@ -288,20 +333,47 @@ def _check_agents(self) -> None:
             )
         if isinstance(self.config.solver.runtime, vf.SubprocessConfig):
             raise ValueError(
-                "agentic-judge plays its judge in the solver's box, but the solver "
-                "(which provisions it) resolves to the subprocess runtime; use "
+                "agentic-judge runs a code-executing solver in a container, but the "
+                "solver resolves to the subprocess runtime; use "
                 "--env.solver.runtime.type docker or prime"
             )
+        if self.config.topology == "isolated" and isinstance(
```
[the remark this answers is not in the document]
rasdani: Confirmed and fixed in e787bc9 — thank you, this was a real break, not just the bundled config.

`AgentConfig.runtime` defaults to `SubprocessConfig` with no `None` to distinguish unset from chosen, so with `isolated` as the default the guard rejected every config that pinned only the solver. That included `test_env_id_agentic_judge`, which pins the solver to docker and leaves the judge unset — it failed at construction before any episode ran. My test runs had it deselected under `-m 'not e2e'`, so I never saw it.

Took the inheritance option rather than editing the config: the judge now falls back to the solver's runtime policy when it has not pinned a container. That matches how `harness`, `model` and `sampling` already resolve for unpinned seats in `Env._episode_agents`, and it is the correct default independently — the grading box is meant to boot from the solver's image, which is what makes 'only the delta travels' true. A judge that pins its own runtime keeps it under `isolated`; under `shared` the solver's still wins.

The explicit subprocess guard is gone, unreachable once the fallback runs first, with the solver-side check still covering both-are-subprocess. Verified across four shapes: unset judge runtime inherits, pinned judge runtime honored, shared overrides, solver-subprocess still raises.

verifiers/v1/tasksets/harbor/taskset.py
```diff
@@ -326,8 +368,69 @@ def parse_task(task_dir: Path, idx: int, harbor_config: HarborConfig) -> HarborD
...
+            raise ValueError(
+                f"{task_dir.name}: artifact {entry.source!r} targets service "
+                f"{entry.service!r}; sidecars need a compose-capable runtime"
+            )
+        # `destination` positions a file in Harbor's host trial directory. Verifiers has
+        # no such directory (the trace is the record) and Harbor never lets destination
+        # affect verifier-side placement, so it cannot change any grading outcome.
+        artifacts.append(
+            Artifact(source=entry.source, exclude=list(entry.exclude or []))
+        )
```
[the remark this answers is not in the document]
rasdani: Real, and worse than described — fixed in 4492b14e7.

It does not fail during collection; it silently collects the wrong file. The existence probe runs with the runtime's workdir set, so `test -e solution.txt` passes against `$workdir/solution.txt`, but the tar runs `-C /` and archives `/solution.txt`. I reproduced it with a decoy at the root:

```
declared relative source -> collected 'DECOY' (wanted INTENDED)
```

So a correct Harbor task — relative main-service sources are permitted, I checked `validate_artifact_entries` — grades against different content with no error raised.

Resolved at collection rather than at parse, and against the runtime's workdir rather than a literal: the workdir is a runtime property and `TaskData.workdir` can override it, so it is not known when `task.toml` is read. `/workspace` in the suggested patch would be wrong on any runtime configured otherwise. Verified the repro now yields `key=/work/solution.txt content='INTENDED'`, with absolute sources unchanged.

docs/v1/harbor.md
rasdani: Done in 4492b14e7 — it repeated the Shortcomings list two lines below.

docs/v1/tasksets.md
```diff
@@ -231,6 +231,41 @@ class JudgeTraceTaskset(vf.Taskset[JudgedTask, SetConfig]):
 
 To override the judge model, set `env.taskset.task.judge.model` in your config (it is a string).
 
+## Grading artifacts
+
+An environment can grade in a second sandbox rather than the one the agent worked in, so nothing the agent did to its environment can reach the grader. Only what the task declares crosses over.
```
hallerite: we def need to de-slop this section imo

docs/v1/tasksets.md
```diff
@@ -231,6 +231,41 @@ class JudgeTraceTaskset(vf.Taskset[JudgedTask, SetConfig]):
...
+```python
+class MyData(vf.TaskData):
+    artifacts: list[vf.Artifact] = [
+        vf.Artifact(source="/work/report", exclude=[".git"])
+    ]
+```
+
+A declared path that is missing at collection time fails the rollout: it was declared because grading needs it, and grading a partial state scores the rollout wrong rather than loudly failing it. The convention dir is exempt — it is collected for every task, and most never write to it.
+
+The grading sandbox boots from the same image as the agent's, so the repo and its dependencies are already present. Only the agent's delta has to travel, which is why the collection cap (`vf.artifacts.MAX_ARTIFACT_BYTES`) is sized for a patch rather than a tree.
```
hallerite: these are specific to the agentic judging env and should not be in tasksets.md imo

verifiers/v1/envs/agentic_judge/env.py
```diff
@@ -1,25 +1,32 @@
-"""agentic-judge: a solver plays the task, a judge verifies it in the same box.
+"""agentic-judge: a solver plays the task, a code-executing judge verifies the work.
```
hallerite: code-executing sounds weird

verifiers/v1/envs/agentic_judge/env.py
```diff
@@ -1,25 +1,32 @@
...
-verdicts to `/tmp/verdict.json`; `finalize()` validates them strictly onto the
-solver's trace — `judge/<name>` metrics plus a weighted-mean `judge` reward,
-composed with the taskset's own rewards via `[env.score]` (judge-only by
-default).
+verdicts to `/tmp/verdict.json`, with the solver's full trace record uploaded at
+`/tmp/trace.json`. `finalize()` validates them strictly onto the solver's trace —
+`judge/<name>` metrics plus a weighted-mean `judge` reward, composed with the
+taskset's own rewards via `[env.score]` (judge-only by default).
+
+`--env.topology` decides where the judge stands. Under `shared` (the default) it
```
hallerite: need a different name here than topology. topology to me means which agents exist and what is their relationship, whereas this is only a subset of that

verifiers/v1/envs/agentic_judge/env.py
```diff
@@ -130,16 +153,28 @@ def __init__(self, data: vf.TaskData, files: dict[str, bytes]) -> None:
         self.files = files
 
     @classmethod
-    def from_trace(cls, solution: vf.Trace, config: "JudgeTaskConfig") -> "JudgeTask":
-        """Mint the judge's task from the solver's finished trace."""
+    def from_trace(
+        cls,
+        solution: vf.Trace,
+        config: "JudgeTaskConfig",
+        topology: str = "shared",
```
hallerite: change name

hallerite: also can we type as a union of literals and define that as a type?

verifiers/v1/envs/agentic_judge/env.py
```diff
@@ -251,23 +286,46 @@ class ScoreConfig(vf.BaseConfig):
...
-    """The solver agent. It owns the shared box, so its runtime must be a
-    container: `--env.solver.runtime.type docker|prime`."""
+    """The solver agent. Its runtime must be a container:
+    `--env.solver.runtime.type docker|prime`."""
     judge: vf.AgentConfig = vf.AgentConfig()
-    """The judge agent. It plays in the solver's box; its own runtime policy is
-    ignored (overwritten with the solver's)."""
+    """The judge agent. Under `isolated` it provisions its own box from this policy;
+    under `shared` it plays in the solver's box and this policy is ignored."""
+    topology: Literal["shared", "isolated"] = "shared"
```
hallerite: lol it's a union of literals here

verifiers/v1/envs/agentic_judge/env.py
```diff
@@ -251,23 +286,46 @@ class ScoreConfig(vf.BaseConfig):
...
 
 class AgenticJudgeEnv(vf.Env[AgenticJudgeEnvConfig]):
     def __init__(self, config: AgenticJudgeEnvConfig) -> None:
-        # The judge plays in the solver's box, so its effective runtime IS the
-        # solver's policy — aligning the config keeps the base env's subprocess
-        # warning and the runtime stamped on the judge's trace truthful.
-        config.judge = config.judge.model_copy(
-            update={"runtime": config.solver.runtime}
-        )
+        # The judge inherits the solver's runtime policy unless it pins a container of
```
hallerite: slop comment

verifiers/v1/envs/agentic_judge/env.py
```diff
@@ -298,13 +356,36 @@ async def setup(self, agents: vf.Agents) -> None:
...
 
     async def run(self, task: vf.Task, agents: vf.Agents) -> None:
+        if self.config.topology == "shared":
+            async with agents.solver.provision(task) as box:
+                solution = await agents.solver.run(task, runtime=box)
+                judge_task = JudgeTask.from_trace(solution, self.config.task, "shared")
+                await agents.judge.run(judge_task, runtime=box)
+            return
+
+        # Collection is the barrier: once it returns, nothing downstream needs the
```
hallerite: slop comment

verifiers/v1/tasksets/harbor/taskset.py
```diff
@@ -327,8 +369,69 @@ def parse_task(task_dir: Path, idx: int, harbor_config: HarborConfig) -> HarborD
...
+    )
+
+
+def parse_verifier_extras(
+    task_dir: Path, parsed
+) -> tuple[list[Artifact], list[CollectHook]]:
+    """Harbor's `artifacts` and `[[verifier.collect]]` blocks, narrowed to what a
+    single-container runtime can honor.
+
+    The convention dir is deliberately not prepended here (Harbor's
```
hallerite: slop comment

verifiers/v1/utils/git.py
```diff
@@ -74,32 +83,86 @@ async def resolve_head(runtime: Runtime, env: dict | None = None) -> str:
...
     full, capped = f"{_FULL}_{nonce}", f"{_CAPPED}_{nonce}"
     cmd = _DIFF.format(full=full, capped=capped, cap=PATCH_CAP_BYTES + 1)
     try:
         result = await runtime.run(
-            ["sh", "-c", cmd],
+            ["sh", "-c", cmd, "vf-capture-patch", *(ignore or [])],
             {**(env or {}), "VF_DIFF_BASE": base_commit or "HEAD"},
         )
         if result.exit_code != 0:
+            # Not every runtime raises when the box is gone — Docker returns `docker
```
hallerite, line 151: unrelated to this PR, but this seems like a smell @mikasenghaas

docs/v1/harbor.md
```diff
@@ -98,10 +98,19 @@ added automatically; evaluator-provided `allow` entries add exceptions and `bloc
...
 accepts host-level entries and rejects combinations that need both policy modes.
 
+## Artifacts and collect hooks
+
+`artifacts = [...]` and `[[verifier.collect]]` are read from `task.toml` ([Harbor Docs](https://www.harborframework.com/docs/run-jobs/results-and-artifacts)). Collect hooks run in the agent's box from the task's `finalize`, which is Harbor's own ordering — after the agent phase, before collection — and declared paths plus the `/logs/artifacts/` convention dir are then carried into the grading box and restored at their original paths ("no translation", as in Harbor).
+
+Two deliberate differences from `harbor run`:
+
+- **A failing collect hook fails the rollout.** Harbor logs it and carries on, because there the output is observability; here it is a grading input, and a silently absent file makes the verifier score a stale state.
+- **`destination` has no effect.** It positions a file in Harbor's host trial directory; verifiers has no trial directory (the trace is the record), and Harbor never lets `destination` affect verifier-side placement.
```
xeophon: wait what?

docs/v1/harbor.md
```diff
@@ -98,10 +98,19 @@ added automatically; evaluator-provided `allow` entries add exceptions and `bloc
...
+
 ## Shortcomings
 
 verifiers does not have parity with Harbor yet, so some features are missing and currently being worked on. The most notable missing features right now are:
 
 - Switching to a different verifier-phase network policy ([Harbor Docs](https://www.harborframework.com/docs/tasks/network-policy))
-- Shared & separate verifiers ([Harbor Docs](https://www.harborframework.com/docs/tasks#verifier-environment-shared-vs-separate))
+- Sidecar services, and the sidecar artifacts and collect hooks that go with them ([Harbor Docs](https://www.harborframework.com/docs/tasks#sidecar-artifacts-and-collect-hooks))
```
xeophon: remove

verifiers/v1/artifacts.py
```diff
@@ -0,0 +1,166 @@
+"""Carry a task's declared files out of the agent's box and into a grading box.
```
hallerite: imo this should not be tied to agentic judging, as it is presented here. this is way more general than that and the comment should thus be more general

docs/v1/tasksets.md
xeophon: before 0.3.0, i will go over all the docs changes and de-slop but yeah, this is a lot

verifiers/v1/artifacts.py
```diff
@@ -0,0 +1,166 @@
...
+
+Grading in the box the agent worked in leaves a seam a policy under RL pressure will
```
hallerite: what in the slop

tests/v1/test_e2e.py
```diff
@@ -392,21 +392,22 @@ async def test_env_id_best_of_n(run_v1, tmp_path):
...
-    """The agentic judge over the echo taskset (needs docker): the box is
-    provisioned once from the solver's runtime policy, the solver plays in it,
-    the judge lands in the SAME box with the graded trace uploaded,
-    investigates with real execution, and its parsed verdict
-    lands on the solver's trace under the spec's reward key. Wiring, not taste:
-    the judge followed the verdict-file contract — the grade itself is the
-    model's call. Exercises the config surface too: a policy-only prompt
-    override (the verdict contract is appended regardless) and
-    reward-composition weights."""
+    """The agentic judge over the echo taskset (needs docker), on the default
```
xeophon: just drop all comments tbf

hallerite (commented):
left some comments, overall like the direction!

verifiers/v1/envs/agentic_judge/env.py
```diff
@@ -111,6 +115,25 @@ def _render(template: str, **fields: str) -> str:
 wrong and references can be narrower than the task; do not over-index on how a
 reference solves it. Your verdict is what YOU verified by execution."""
 
+SHARED_SANDBOX_NOTE = f"""\
```
xeophon: why are those notes needed?

verifiers/v1/envs/agentic_judge/env.py
```diff
@@ -111,6 +115,25 @@ def _render(template: str, **fields: str) -> str:
...
+## Your workspace
+
+Your sandbox is the SAME box the graded agent worked in, in the state the agent
+left it — its edits (and any scoring side effects) are applied. {_RECORD_NOTE}"""
+
+ISOLATED_SANDBOX_NOTE = f"""\
```
xeophon: why are those notes needed?

verifiers/v1/tasksets/harbor/taskset.py
```diff
@@ -76,6 +79,13 @@ class Author(StrictBaseModel):
     email: str | None = None
 
 
+class CollectHook(StrictBaseModel):
+    """One `[[verifier.collect]]` command, run in the agent's box by `finalize`."""
+
+    command: str
+    timeout_sec: float = 60.0
```
xeophon: i would bump considerably, the other timeouts in v1 are really high on purpose as well. 60s could be too short for huge files / compilation scripts

verifiers/v1/tasksets/harbor/taskset.py
```diff
@@ -327,8 +369,69 @@ def parse_task(task_dir: Path, idx: int, harbor_config: HarborConfig) -> HarborD
...
+        artifacts.append(
+            Artifact(source=entry.source, exclude=list(entry.exclude or []))
+        )
+
+    hooks: list[CollectHook] = []
+    for hook in verifier.collect:
+        if hook.service != MAIN_SERVICE_NAME:
+            raise ValueError(
+                f"{task_dir.name}: collect hook targets service {hook.service!r}; "
+                "sidecars need a compose-capable runtime"
```
xeophon: no one knows what sidecars are in this context

verifiers/v1/artifacts.py
```diff
@@ -0,0 +1,166 @@
...
+    """Tar the convention dir and every declared path out of `runtime`.
+
+    Keyed by source path; the values are tar archives. Insertion order is the order
+    they were declared, and a path cannot be collected twice.
+
+    A declared source that is missing raises: it was declared because grading needs it,
+    and grading a partial state scores the rollout wrong rather than failing it. The
+    implicit convention sweep is exempt — most tasks never write there.
+
+    One archive per source: BusyBox `tar` (every alpine-based image) implements only
```
xeophon: what about other images?

verifiers/v1/artifacts.py
```diff
@@ -0,0 +1,166 @@
...
+
+from verifiers.v1.errors import ArtifactError
+from verifiers.v1.types import StrictBaseModel
+
+if TYPE_CHECKING:
+    from verifiers.v1.runtimes import Runtime
+
+logger = logging.getLogger(__name__)
+
+CONVENTION_DIR = "/logs/artifacts"
```
mikasenghaas: `ARTIFACTS_DIR`

verifiers/v1/artifacts.py
```diff
@@ -0,0 +1,166 @@
...
+    collected: dict[str, bytes] = {}
+    budget = MAX_ARTIFACT_BYTES
+    for artifact in entries:
+        source = artifact.source
+        if (await runtime.run(["test", "-e", source], {})).exit_code != 0:
+            if sweep and source == CONVENTION_DIR:
+                continue
+            raise ArtifactError(
+                f"declared artifact {source!r} does not exist in the box; the task must "
+                "produce it in finalize() (or a [[verifier.collect]] hook)"
```
mikasenghaas: would not mention harbor-concepts in core vf code

verifiers/v1/errors.py
```diff
@@ -74,6 +74,13 @@ class TaskError(RolloutError):
     """Task-authored code raised — `setup`, `finalize`, or a `@reward`/`@metric`."""
 
 
+class ArtifactError(RolloutError):
```
mikasenghaas: dont think we should have a error-class for this. general convention is that we only map "where" the error occurred (env, task, harness, etc.) so this would fall under task error. we then wrap this error so you would get details on that the artifact download/upload failed

verifiers/v1/artifacts.py
```diff
@@ -0,0 +1,166 @@
...
+            await runtime.run(["rm", "-f", path], {})
+        except Exception:
+            logger.debug("failed to remove %s", path, exc_info=True)
+
+
+async def _run(runtime: Runtime, command: str, action: str) -> None:
+    result = await runtime.run(["sh", "-c", command], {})
+    if result.exit_code:
+        detail = (result.stderr or result.stdout).strip()[-500:]
+        raise ArtifactError(f"failed to {action}: {detail}")
```
mikasenghaas: all these can then become standard `RuntimeError`

verifiers/v1/utils/git.py
mikasenghaas: yes agree, made a ticket for it

verifiers/v1/utils/git.py
```diff
@@ -74,32 +83,86 @@ async def resolve_head(runtime: Runtime, env: dict | None = None) -> str:
...
             trace.info["patch_error"] = (
                 f"exit={result.exit_code} {(result.stderr or '').strip()[-500:]}"
             )
             return
         raw = await runtime.read(capped)
-    except Exception as exc:  # noqa: BLE001 - capture must never fail the rollout
-        trace.info["patch_error"] = f"{type(exc).__name__}: {exc}"
-        return
+    except Exception as exc:
+        raise SandboxError(f"patch capture could not reach the box: {exc}") from exc
```
mikasenghaas: dont think we need the cutom error here. can just let the natural error propagate

mikasenghaas: agree tho that we want this to bubble up

verifiers/v1/utils/git.py
```diff
@@ -111,3 +174,13 @@ async def capture_patch(
         raw = raw[:PATCH_CAP_BYTES]
         trace.info["patch_truncated"] = True
     trace.info["patch"] = raw.decode("utf-8", errors="replace")
+    if write_path is not None:
+        try:
+            parent = str(PurePosixPath(write_path).parent)
+            await runtime.run(["mkdir", "-p", parent], env or {})
+            await runtime.write(write_path, raw)
+        except Exception as exc:
```
mikasenghaas: same here

verifiers/v1/__init__.py
```diff
@@ -155,6 +162,9 @@
 from verifiers.v1.utils.git import (
     resolve_head as resolve_head,
 )
+from verifiers.v1.utils.git import (
+    snapshot_untracked as snapshot_untracked,
```
mikasenghaas: huh?

verifiers/v1/envs/agentic_judge/env.py
mikasenghaas: would just do `share_runtime: bool = True` here

verifiers/v1/envs/agentic_judge/env.py
```diff
@@ -298,13 +356,36 @@ async def setup(self, agents: vf.Agents) -> None:
...
-            judge_task = JudgeTask.from_trace(solution, self.config.task)
-            await agents.judge.run(judge_task, runtime=box)
+            if not solution.ok:
+                # The rollout errored, so its own scoring never ran either; grading it
+                # would spend a second sandbox to reproduce a failure the trace already
+                # records. Return rather than raise: `episode.ok` follows the failed
+                # trace, and `episode_should_retry` classifies off that trace's own
+                # error — an exception raised here would bury the real type.
+                return
+            collected = await vf.collect(box, task.data.artifacts)
```
mikasenghaas: would do the collection in the solver task finalize() method, and then we restore in judge task setup method. this pattern naturally fits our api. then the else block becomes a 2-liner
```
solution = await agents.solver.run(task)
await agents.judge.run(JudgeTask.from_trace(solution, self.config.task))
```

mikasenghaas: we dont need the the provision api here

verifiers/v1/tasksets/harbor/taskset.py
```diff
@@ -94,11 +104,42 @@ class HarborData(TaskData):
...
+        makes the verifier score a stale state instead of failing loudly.
+        """
+        for hook in self.data.collect:
+            try:
+                result = await asyncio.wait_for(
+                    runtime.run(["sh", "-c", hook.command], verifier_env(self.data)),
+                    hook.timeout_sec,
+                )
+            except TimeoutError as exc:
+                raise TaskError(
```
mikasenghaas: we never need to raise vf errors. those are entirely framework-handled (e.g. each user task hook is try..except'ed and wraps `TaskError`. this would lead to double task error)

verifiers/v1/envs/agentic_judge/env.py
rasdani: Mostly they aren't — dug into this and the mode-specific pair is gone in 960d45b64, replaced by one mode-neutral note.

What made the pair look necessary: a judge in a fresh box sees a clean tree, and "clean tree" is ambiguous between "agent did nothing" (grade 0) and "the work travelled as an artifact" (read/apply it). But the evidence channels turn out to be placement-independent — `/tmp/trace.json` is uploaded in both modes, and `/logs/artifacts` is populated in both too (finalize writes it in the solver's box, so the shared judge sees the same files the isolated judge gets restored). So one note covers both placements, and the note-must-match-placement invariant on `from_trace` disappears.

What survives is one sentence of the old content: "an unmodified tree does not mean the agent did nothing." That one earned its keep empirically — the isolated note's old wording ("artifacts were restored at their original paths") produced exactly the misreading it existed to prevent, in a reviewer who knows this design well. Kept, reworded, and the note now actually names `/logs/artifacts` instead of gesturing at "original paths".

verifiers/v1/envs/agentic_judge/env.py
rasdani: Mostly they aren't — dug into this and the mode-specific pair is gone in 960d45b64, replaced by one mode-neutral note.

What made the pair look necessary: a judge in a fresh box sees a clean tree, and "clean tree" is ambiguous between "agent did nothing" (grade 0) and "the work travelled as an artifact" (read/apply it). But the evidence channels turn out to be placement-independent — `/tmp/trace.json` is uploaded in both modes, and `/logs/artifacts` is populated in both too (finalize writes it in the solver's box, so the shared judge sees the same files the isolated judge gets restored). So one note covers both placements, and the note-must-match-placement invariant on `from_trace` disappears.

What survives is one sentence of the old content: "an unmodified tree does not mean the agent did nothing." That one earned its keep empirically — the isolated note's old wording ("artifacts were restored at their original paths") produced exactly the misreading it existed to prevent, in a reviewer who knows this design well. Kept, reworded, and the note now actually names `/logs/artifacts` instead of gesturing at "original paths".

xeophon (dismissed):

verifiers/v1/envs/agentic_judge/env.py
```diff
@@ -97,19 +99,26 @@ def _render(template: str, **fields: str) -> str:
...
-left it — its edits (and any scoring side effects) are applied. The agent's raw
-trace record (JSON: messages, tool calls, and its `info` artifacts) is uploaded
-at `{TRACE_FILE}`. The record can be very large — never dump it whole; peek
-selectively (list its keys, then slice out specific fields with python or jq)
-and pull only what you need. It is complete — it may also carry the task's own
-scores/metrics and reference material (a gold answer, a reference solution,
-held-out tests). Those are context, not your standard: recorded scores can be
-wrong and references can be narrower than the task; do not over-index on how a
-reference solves it. Your verdict is what YOU verified by execution."""
+Your sandbox is either the box the graded agent worked in or a fresh one built
```
hallerite: this is strange. we should have different prompts for both cases imo

verifiers/v1/envs/agentic_judge/env.py
```diff
@@ -251,23 +278,24 @@ class ScoreConfig(vf.BaseConfig):
...
 class AgenticJudgeEnv(vf.Env[AgenticJudgeEnvConfig]):
     def __init__(self, config: AgenticJudgeEnvConfig) -> None:
-        # The judge plays in the solver's box, so its effective runtime IS the
-        # solver's policy — aligning the config keeps the base env's subprocess
-        # warning and the runtime stamped on the judge's trace truthful.
-        config.judge = config.judge.model_copy(
-            update={"runtime": config.solver.runtime}
-        )
+        if config.share_runtime or isinstance(
+            config.judge.runtime, vf.SubprocessConfig
```
hallerite: I don't like that we have this fail-safe here for the SubprocessConfig. The judge task needs a container, which should be enough

verifiers/v1/envs/agentic_judge/env.py
```diff
@@ -298,13 +326,24 @@ async def setup(self, agents: vf.Agents) -> None:
...
+        solution = await agents.solver.run(task)
+        if not solution.ok:
+            return
+        await agents.judge.run(
+            JudgeTask.from_trace(solution, self.config.task, share_runtime=False)
+        )
 
     async def finalize(self, task: vf.Task, episode: vf.Episode) -> None:
         by_agent = {t.agent_name: t for t in episode.traces}
+        if "judge" not in by_agent:
```
hallerite, line 364: what is this for?

verifiers/v1/artifacts.py
```diff
@@ -0,0 +1,144 @@
...
+
+    logger.debug("collected artifact roots: %s", list(collected))
+    return collected
+
+
+async def restore(runtime: Runtime, collected: dict[str, bytes]) -> None:
+    """Extract `collected` in `runtime` at the original absolute paths."""
+    if not collected:
+        return
+    # Extraction writes to absolute paths. In a container that is the point; under the
```
hallerite: wondering whether we shouldn't force non-subprocess runtimes for this. otherwise we go into weird behavior where concurrent rollouts might clobber each other

verifiers/v1/task.py
```diff
@@ -122,6 +123,12 @@ class TaskData(StrictBaseModel):
     """Execution-time destinations denied by this task and combined with runtime
     blocks. Non-empty concrete allowlists cannot be combined with blocklists. Docker
     framework routes take precedence; ordinary Prime deny rules pass through unchanged."""
+    artifacts: list[Artifact] = Field(default_factory=list)
+    """Paths carried out of the agent's box and into a grading box, on top of the
```
hallerite: this is more general than what the comment suggests, so let's not constrain ourselves to that in the comments

hallerite (commented):
left some more comments

verifiers/v1/envs/agentic_judge/env.py
mikasenghaas: i think the solution not ok early return?

verifiers/v1/utils/git.py
mikasenghaas: this is fixed in #2156

verifiers/v1/envs/agentic_judge/env.py
rasdani: yes early return and don't spin up grading sandbox

verifiers/v1/envs/agentic_judge/env.py
rasdani: ok @xeophon flagged the sandbox notes before

will revert then to the two versions

verifiers/v1/artifacts.py
rasdani: should or shouldn't?

i force non-subprocess runtimes on line 91

hallerite: then the comment is misleading (if you don't read the code below that is)

verifiers/v1/envs/agentic_judge/env.py
```diff
@@ -180,31 +217,40 @@ async def finalize(self, trace: vf.Trace, runtime: vf.Runtime) -> None:
...
     """Criteria the judge grades against: a `.toml`/`.json` file with a
     `criteria` list — the plugged rubric judge's format, so the same rubric
     files work for both. None grades the single built-in `solved` criterion."""
 
+    @staticmethod
+    def _resolve(value: Path | str) -> str:
+        path = Path(value)
+        if isinstance(value, Path) or path.suffix in (".md", ".txt"):
+            return path.read_text(encoding="utf-8")
+        return str(value)
```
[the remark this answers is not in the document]
hallerite: Fixed in 96f2b5937. Plain strings are now always inline text, including values ending in `.md` or `.txt`. File-backed text is explicit: pass a `Path` in Python or `{ path = "policy.md" }` in config. The existing E2E config now exercises both the structured file form and an inline hint ending in `README.md`.

verifiers/v1/envs/agentic_judge/env.py
hallerite: Addressed in 4228c44 and fef21cd34. The isolated judge no longer gets the solver runtime as a fallback; its configured runtime is preserved. `JudgeTask.NEEDS_CONTAINER` remains the source of truth, and the existing generic `validate_pairing` check now runs during environment construction as an early preflight rather than after the solver rollout.

verifiers/v1/envs/agentic_judge/env.py
hallerite: Addressed in 4228c44. `JudgeTask.from_trace` now selects distinct `SHARED_WORKSPACE_NOTE` and `ISOLATED_WORKSPACE_NOTE` text from `share_runtime`, so the judge is told whether it is in the solver workspace or a fresh restored sandbox.

verifiers/v1/envs/agentic_judge/env.py
```diff
@@ -180,31 +217,40 @@ async def finalize(self, trace: vf.Trace, runtime: vf.Runtime) -> None:
...
+    """Optional inline hints or a `.md`/`.txt` hints file, injected as their own
+    section: task-family pointers into the trace or box — e.g. for math, where
+    the reference answer lives in the record; for SWE, to diff the repo or read
+    `info.patch`."""
     rubric: Path | None = None
     """Criteria the judge grades against: a `.toml`/`.json` file with a
     `criteria` list — the plugged rubric judge's format, so the same rubric
     files work for both. None grades the single built-in `solved` criterion."""
 
+    @staticmethod
```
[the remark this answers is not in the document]
hallerite: Fixed in 96f2b5937. Plain strings are now always inline text, including values ending in `.md` or `.txt`. File-backed text is explicit: pass a `Path` in Python or `{ path = "policy.md" }` in config. The existing E2E config now exercises both the structured file form and an inline hint ending in `README.md`.

hallerite (approved):

PR #2170, opened 2026-07-29
Add an interception example environment

xeophon:
## Overview

This adds a small runnable environment showing the minimal custom interception interface.

The task defines one deterministic guard and one judge-backed guard directly with `@vf.intercept`. Both receive one `vf.Trace`; the deterministic guard reads `trace.last_message`, the judge sees `trace.messages`, and either can return `trace.replace(content)` to rewrite the candidate.

## What this PR adds

- An `environments/interception_v1` package with two focused examples.
- A deterministic guard based on `trace.last_message`.
- A judge-backed guard based on `trace.messages`.
- Explicit priority ordering between the two guards.
- A reward that reads interception records from the trace.
- A short README with the complete handler shape.

## Stack

This is PR 5 of 5. Each PR is based on the one before it and is intended to merge in order.

1. **Stream replay safety** — keep an explicitly retried streamed request on the same model response.
2. **Request and response interception** — run `@vf.intercept` before model requests and before responses reach the harness, and make rewrites canonical.
3. **Terminal interception** — let a guard end a rollout with a final reason and reward.
4. **Reusable guards** — add common tool guards and make custom handlers operate directly on `vf.Trace`.
5. **Example environment** — show deterministic and judge-backed trace guards directly with `@vf.intercept` in a runnable taskset.

### DIFF
--- a/environments/interception_v1/README.md
+++ b/environments/interception_v1/README.md
@@ -3,22 +3,22 @@
 An example v1 environment that defines deterministic and judge-backed guards directly
 with `@vf.intercept`.
 
-The task class installs two response handlers in priority order:
+The task class installs two handlers in priority order:
 
-1. `deterministic_guard` checks `exchange.message` with an exact rule.
-2. `judge_guard` calls `exchange.judge()` only when the deterministic rule did not
-   rewrite the candidate.
+1. `deterministic_guard` checks `trace.last_message` with an exact rule.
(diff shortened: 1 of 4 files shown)


xeophon:
Superseded by #2178, which replaces the guard layer and example environment with the reworked interception surface (typed rules, prebuilt guards).

PR #2176, opened 2026-07-30
fix(serve): fall back to loopback TCP where zmq has no ipc transport

arthi-arumugam-git:
Partially addresses #2030, reported by @martian56, who identified the root cause correctly.

## What is wrong

`prime eval run` cannot start on native Windows. `make_ipc_address` hardcodes the transport:

```python
def make_ipc_address(session_id: str, name: str) -> str:
    return f"ipc:///tmp/vf-{session_id}-{name.replace('/', '--')}"
```

`ipc://` is a Unix domain socket transport. On Windows libzmq is built without it, so `EnvRouter.__init__` dies on the very first `bind()`, before a single rollout runs.

Measured on Windows 11, pyzmq 27.1.0, libzmq 4.3.5:

```
zmq.has("ipc") = False

main         binds response_pull to  ipc:///tmp/vf-abc123-responses
                                     -> ZMQError: Protocol not supported

this branch  binds response_pull to  tcp://127.0.0.1:50465
                                     -> binds
```

## The fix, and why it gates on a capability rather than the platform

`zmq.has("ipc")`, not `os.name == "nt"`. The transport choice depends on whether libzmq has ipc compiled in, which is exactly what that call reports, and it also covers a libzmq built without ipc on any OS. Where ipc is unavailable it falls back to `tcp://127.0.0.1` using the `get_free_port` helper that already sits directly below in the same module, so no new dependency and no new idiom.

## Two more defects in the same three lines

- **`/tmp` was hardcoded**, so the socket path ignored `tempfile.gettempdir()`. That breaks anywhere `/tmp` is not writable, independent of Windows. It now resolves.
- **Cleanup would have tried to unlink a TCP address.** `EnvRouter` builds `ipc_paths` by stripping the `"ipc://"` prefix, and shutdown calls `os.unlink` on every entry. A TCP address has no file behind it. Added `ipc_path_of`, which returns the path for `ipc://` and `None` for anything else, and only real paths are tracked. Without this the fix would have swapped one Windows failure for a different one on shutdown.

## One thing I checked before changing it

A random port makes `make_ipc_address` non-deterministic, and `get_worker_address` recomputes the address rather than reading a stored one, so I checked whether anything relies on calling it twice for the same worker. It does not: `get_worker_address` has exactly one caller, in `start_worker`, and the result is handed to the spawned process as the `request_address` kwarg rather than recomputed worker-side. So non-determinism is safe here. If you would rather it stayed deterministic, the alternative is to bind with `bind_to_random_port` and read `zmq.LAST_ENDPOINT` back, which is more robust against the small TOCTOU window in `get_free_port` but touches all three bind sites. Happy to do it that way instead.

## Tests

`tests/test_serve_utils_addresses.py`, 11 cases, all passing. They cover both branches by patching `zmq.has`, so they exercise the Windows path on Linux CI too and the ipc path on Windows.

- ipc branch uses the resolved temp dir, not `/tmp`, and flattens a slash in the env id
- fallback returns loopback TCP, and **the returned address is actually bound** to prove it works rather than just asserting on the string
- distinct names get distinct ports, since router and each worker bind separately
- `ipc_path_of` returns `None` for tcp, inproc and pgm, so nothing but a real path reaches `os.unlink`
- a regression guard that the literal `ipc:///tmp/` prefix from the issue is never returned again

```
tests/test_serve_utils_addresses.py  11 passed
ruff format, ruff check: clean
```

## Scope, deliberately

The issue reports a second error: the client receive loop needs a selector event loop for zmq, and pyzmq already works under Proactor when `tornado>=6.1` is installed. That reads as a dependency or CLI-entry-point concern rather than a transport bug, and setting a global `asyncio` event loop policy from inside a library affects anyone embedding this. So it is not in here. Say the word and I will add it at the CLI entry point.

I also could not verify end to end: `prime eval run demo -m openai/gpt-4o-mini -p openrouter -n 1 -c 1` needs a provider key I do not have. What is verified is that the router's binds now succeed on Windows where they previously raised, plus the unit tests above. Worth a maintainer running the full command on Windows before merge.

---

> [!NOTE]
> **Medium Risk**
> Changes core serve startup/shutdown transport and port allocation for all platforms; misallocation could still cause bind races or failed env server startup, though behavior is heavily unit-tested.
> 
> **Overview**
> Fixes **#2030** by replacing hardcoded `ipc:///tmp/...` router/worker addresses with transport-aware construction so `EnvRouter` can bind on Windows and in other environments where `/tmp` or Unix sockets are unusable.
> 
> **`make_ipc_address`** now uses **`zmq.has("ipc")`**: when ipc is available it builds paths under the shortest writable temp directory (not fixed `/tmp`), flattens slashes in names, and shortens paths that would exceed **`sun_path`** limits; when ipc is missing it returns **`tcp://127.0.0.1:<port>`** via an enhanced **`get_free_port`** that skips excluded ports and avoids reissuing recently probed ports. **`ipc_path_of`** ensures shutdown only **`unlink`**s real ipc socket files.
> 
> **`EnvRouter`** tracks **`issued_ports`** across response, stats, and worker addresses so TCP ports allocated before child processes bind are not reused, and only ipc paths are added to **`ipc_paths`**.
> 
> Adds **`tests/test_serve_utils_addresses.py`** covering ipc vs TCP branches, **`sun_path`** shortening, port exclusion, and bindability.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 99618300542e0cded08b641f65dd46f4c18d8d77. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fall back to loopback TCP in `make_ipc_address` when zmq has no ipc transport
> - `make_ipc_address` in [`serve_utils.py`](https://github.com/PrimeIntellect-ai/verifiers/pull/2176/files#diff-573ce1050a71bef580f9e082113a30474324c2d0e92c785579dce9d67eb1a4ca) now detects `zmq.has("ipc") == False` and returns a `localhost` TCP address instead of an ipc path, using `get_free_port` with an exclusion set to avoid port reuse.
> - `get_free_port` holds probe sockets open during selection and accepts an `exclude` container, reducing the chance of issuing the same port to multiple sockets.
> - [`EnvRouter`](https://github.com/PrimeIntellect-ai/verifiers/pull/2176/files#diff-5f42e5c44b82aeddf6020de01b59c5480cac9144a3f6ffc79c3ee01683ad92ec) passes a shared `issued_ports` set into every `make_ipc_address` call and uses the new `ipc_path_of` helper to track only real filesystem paths, avoiding bogus unlink attempts on TCP addresses.
> - Risk: `get_free_port` now raises `RuntimeError` if no acceptable port is found within 100 attempts, whereas it previously always succeeded.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9961830.</sup>
> 
>

### DIFF
--- a/tests/test_serve_utils_addresses.py
+++ b/tests/test_serve_utils_addresses.py
@@ -5,6 +5,7 @@ socket transport, so on Windows `EnvRouter.__init__` could not bind at all and t
 server died before a single rollout ran.
 """
 
+import os
 import tempfile
 from pathlib import Path
 from unittest.mock import MagicMock, patch
@@ -12,7 +13,12 @@ from unittest.mock import MagicMock, patch
 import pytest
 import zmq
 
-from verifiers.utils.serve_utils import get_free_port, ipc_path_of, make_ipc_address
+from verifiers.utils.serve_utils import (
+    _SUN_PATH_MAX,
+    get_free_port,
+    ipc_path_of,
+    make_ipc_address,
+)
 
 
 class _ScriptedProbe:
@@ -64,13 +70,23 @@ class TestIpcPathOf:
 
 
 class TestMakeIpcAddressWithIpcSupport:
-    def test_uses_ipc_and_the_resolved_temp_dir(self):
+    def test_uses_ipc_and_a_writable_directory(self):
         with patch.object(zmq, "has", return_value=True):
             address = make_ipc_address("abc123", "responses")
         assert address.startswith("ipc://")
-        # not a hardcoded /tmp: it has to follow tempfile.gettempdir()
-        expected = Path(tempfile.gettempdir()) / "vf-abc123-responses"
-        assert address == f"ipc://{expected}"
+        path = Path(ipc_path_of(address))
+        assert path.name == "vf-abc123-responses"
+        # not a hardcoded /tmp: whatever directory is chosen has to be writable
+        assert os.access(path.parent, os.W_OK)
+
+    def test_falls_back_to_gettempdir_when_tmp_is_not_writable(self):
+        """The reason the hardcoded /tmp went away in the first place."""
+        with (
+            patch.object(zmq, "has", return_value=True),
+            patch("verifiers.utils.serve_utils.os.access", return_value=False),
+        ):
+            address = make_ipc_address("abc123", "responses")
+        assert Path(ipc_path_of(address)).parent == Path(tempfile.gettempdir())
 
     def test_slashes_in_name_are_flattened(self):
         """Worker names embed the env id, which can contain a slash."""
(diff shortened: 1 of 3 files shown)


arthi-arumugam-git:
Good catch from Bugbot on the port race, and it is right. Fixed in b82375d.

`get_free_port` closes its probe socket before returning, so the port is only free
until something binds it. The router's own two sockets bind on the line after the
probe, so their window is negligible. The workers are the real case: `get_worker_address`
runs in the parent, but the child only binds after `load_environment`, and
`start_workers` allocates in a loop. So worker 1's probe really can be handed the port
worker 0 has not claimed yet, and worker 0 then dies on `Address already in use`.
That is the same startup crash this PR exists to remove, so it was worth closing.

`get_free_port` now takes an `exclude` set and `EnvRouter` threads one shared
`issued_ports` through all three allocations. Rejected probes are held open until the
search finishes, so the OS has to offer a distinct port on each attempt rather than
possibly repeating the one just refused; the search is bounded and raises instead of
spinning. Six tests cover it, all of which fail without the change.

One thing I did not do, so it is visible rather than buried. This closes the
in-process race, which is the one this PR creates. It does not close the window
against an unrelated process on the machine grabbing the port between probe and bind,
which is inherent to bind-probing and already applies to the existing
`get_free_port` call in `environment.py`. Eliminating it properly means not
pre-computing the address at all: bind `tcp://127.0.0.1:*` and read back
`zmq.LAST_ENDPOINT`, the pattern already used in `v1/serve/pool.py` and
`v1/serve/server.py`. For the worker channel that needs either a back-channel for the
child to report its port, or inverting bind/connect so the parent's PUSH binds and the
child's PULL connects. Both are more than a startup-crash fix should carry, and I did
not want to guess at your intent on reconnect semantics. Happy to do it in a follow-up
if you would prefer that shape.

arthi-arumugam-git:
Bugbot's second one is also right, and it is a regression this PR introduced rather
than something pre-existing. Fixed in 9961830.

`sun_path` is 104 bytes on macOS. `/tmp` cost 4 of them; `tempfile.gettempdir()` on
macOS returns the per-user `TMPDIR` launchd sets under `/var/folders/`, which is about
48. So swapping the hardcode took roughly 44 bytes off the budget available to the
socket name, and the worker names are the long ones because they carry the env id.
Concretely, a 12-char session id plus `some-org--some-env-name-v2-0` lands at 118 bytes
under a macOS TMPDIR. That fails on `bind` in the child, after the router has already
come up, which is the same shape of late startup failure this PR is trying to remove.

Two parts to the fix. `_ipc_socket_dir` now picks the shorter of `gettempdir()` and
`/tmp` among whichever is actually writable, so the common case is back to spending 4
bytes and the unwritable-`/tmp` case that motivated dropping the hardcode still works.
`_fit_socket_path` then length-checks the assembled path and, if it still does not fit,
truncates and appends an 8-char digest of the full name so two workers cannot collapse
onto the same path. That second part is not strictly a regression fix: a long enough
env id overflowed under plain `/tmp` before this PR too. It raises rather than
returning a path that cannot bind if the directory alone leaves no room.

Five tests, three of which fail without the change. The other two are guards: one
pins that a name which already fits gets no digest suffix, the other that
`gettempdir()` is still used when `/tmp` is not writable.

PR #2210, opened 2026-07-31
fix(eval-client): give each rollout its own HTTP transport

stu-cao:
## Problem

`EvalClient` shares one httpx pool across every rollout, hardcoded to `max_connections=128`. That silently caps concurrent in-flight model calls regardless of `--max-concurrent`: with `-c 512`, rollouts past the 128th block inside httpx (`timeout=None`, so no error and no log) waiting for a pool slot.

Measured against a self-hosted vLLM deployment (per-rank `/metrics`, 10s samples): engine-side in-flight pinned at **exactly 128, never 129** across 1,600+ samples spanning two full SWE-bench runs — including a phase where server queue-time p99 reached 24s and an unbounded client would have piled far higher. A separate load generator with its own pool drove 136+ concurrent through the same router and engines, so the ceiling was client-side.

## Change

`EvalClient` keys its transports by `session_id`, the rollout trace id, in a single map — calls made outside a rollout (aux routes, the only session-less caller) use an empty id, so there is one lookup and one construction path. `InterceptionServer.unregister` releases it where the session is already sealed, so transport lifecycle matches rollout lifecycle. In-flight capacity now scales with rollout count structurally rather than tracking a pool-size knob, one rollout's connection state can't contend with another's, and an aborted rollout can't leak connections into a shared pool. The 128-connection bound carries over from the old shared pool but now applies per rollout, stated explicitly rather than inherited from httpx defaults: a harness may run subagents that call in parallel, so a tight per-rollout cap would reintroduce the same starvation one scope down. The public `self.http` attribute is gone with the shared pool (nothing in verifiers, its tests, or prime-rl referenced it).

`TrainClient` keeps its process-shared transport: its renderer pool costs ~311 MB and ~2s per slot (measured, GLM-5.2 tokenizer), so per-rollout instances would be prohibitive at real concurrency. It inherits `release_session` as a no-op.

## Validation

512-rollout SWE-bench eval on this branch: 512/512 traces, sessions released cleanly with no fd growth over the run, transient upstream-connection error rate unchanged. Per-session reuse and isolation, session-less fallback, idempotent release, and `close()` draining stragglers verified directly; the client, interception, and eval test files pass.

> [!NOTE]
> ## Problem
>
> `EvalClient` shares one httpx pool across every rollout, hardcoded to `max_connections=128`. That silently caps concurrent in-flight model calls regardless of `--max-concurrent`: with `-c 512`, rollouts past the 128th block inside httpx (`timeout=None`, so no error and no log) waiting for a pool slot.
>
> Measured against a self-hosted vLLM deployment (per-rank `/metrics`, 10s samples): engine-side in-flight pinned at **exactly 128, never 129** across 1,600+ samples spanning two full SWE-bench runs — including a phase where server queue-time p99 reached 24s and an unbounded client would have piled far higher. A separate load generator with its own pool drove 136+ concurrent through the same router and engines, so the ceiling was client-side.
>
> ## Change
>
> `EvalClient` keys its transports by `session_id`, the rollout trace id, in a single map — calls made outside a rollout (aux routes, the only session-less caller) use an empty id, so there is one lookup and one construction path. `InterceptionServer.unregister` releases it where the session is already sealed, so transport lifecycle matches rollout lifecycle. In-flight capacity now scales with rollout count structurally rather than tracking a pool-size knob, one rollout's connection state can't contend with another's, and an aborted rollout can't leak connections into a shared pool. The 128-connection bound carries over from the old shared pool but now applies per rollout, stated explicitly rather than inherited from httpx defaults: a harness may run subagents that call in parallel, so a tight per-rollout cap would reintroduce the same starvation one scope down. The public `self.http` attribute is gone with the shared pool (nothing in verifiers, its tests, or prime-rl referenced it).
>
> `TrainClient` keeps its process-shared transport: its renderer pool costs ~311 MB and ~2s per slot (measured, GLM-5.2 tokenizer), so per-rollout instances would be prohibitive at real concurrency. It inherits `release_session` as a no-op.
>
> ## Validation
>
> 512-rollout SWE-bench eval on this branch: 512/512 traces, sessions released cleanly with no fd growth over the run, transient upstream-connection error rate unchanged. Per-session reuse and isolation, session-less fallback, idempotent release, and `close()` draining stragglers verified directly; the client, interception, and eval test files pass.
>
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)
> 
> #### Changes since #2210 opened
>
> - Increased connection limits for per-rollout HTTP clients in `EvalClient._http_for` [4f7cf7f]
> - Replaced dual HTTP client storage model with unified session-keyed map in `EvalClient` [4f7cf7f]
> - Updated `EvalClient.close` to release all clients from unified map [4f7cf7f]
> - Modified `InterceptionServer` to track HTTP transport release tasks in a `release_tasks` set field, changed the `unregister` method to schedule transport releases using a new `_release_transport` helper method that suppresses exceptions, and added cleanup logic to remove completed tasks from the tracking set [96364d5]
> - Added cleanup mechanism to await in-flight transport release tasks during server shutdown [287f271]
> 
>

---

> [!NOTE]
> **Medium Risk**
> Changes eval upstream concurrency and connection lifecycle at high `--max-concurrent`, but scope is limited to EvalClient and interception teardown with TrainClient unchanged.
> 
> **Overview**
> **Eval** no longer funnels every rollout through one shared `httpx` client capped at 128 connections. Each rollout gets its own transport keyed by `session_id` (the trace id), with session-less aux calls using a separate empty-key client, so concurrent in-flight model calls can scale with rollout count instead of blocking silently on a global pool.
> 
> The base `Client` adds `release_session`, implemented on `EvalClient` to close and drop that rollout’s client; `close()` drains any stragglers. **Interception** schedules transport release when a rollout unregisters (tracked tasks, drained on server shutdown) so teardown doesn’t block and connections don’t leak across rollouts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 287f271e3a535b6ed8b620c24fe10a237346f0b0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


verifiers/v1/clients/eval.py
```diff
@@ -136,23 +141,44 @@ def _headers(
         headers.update(dialect.auth_headers(self.api_key))
         return headers
 
+    def _http_for(self, session_id: str | None) -> httpx.AsyncClient:
+        """The rollout's own client, or the shared one for session-less traffic. Rollout
```
mikasenghaas: what is session less traffic? would prefer just having one code path ideally

verifiers/v1/clients/eval.py
```diff
@@ -136,23 +141,44 @@ def _headers(
...
+        turns are sequential (one outstanding request), so a tiny pool suffices — the
+        headroom is for harness-SDK retry overlap."""
+        if session_id is None:
+            return self.http
+        client = self._session_http.get(session_id)
+        if client is None:
+            client = self._session_http[session_id] = httpx.AsyncClient(
+                timeout=None,
+                limits=httpx.Limits(max_connections=4, max_keepalive_connections=2),
```
mikasenghaas: i think even for a single rollout we might want to haver >4 concurrency conns bc a harness might have subagents which have parallel reqs. i think prob default http limits are reasonable

mikasenghaas (commented):
im wondering if we should make the `vf.Client` atomic itself. ie. it owns its own lifecycle and then `session` (which is all state per rollout) creates and tearsdown its own client

iirc, this would mean the client moves off the `ctx: ModelContext` and into `Rollout` or `Session`

verifiers/v1/clients/eval.py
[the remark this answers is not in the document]
stu-cao: Good point. Fixed in newer revision to track them in `self.release_tasks`

stu-cao: Good point. The default limit is 100. I set it to 128 explicitly in the new revision

verifiers/v1/clients/eval.py
stu-cao: Yes, I fixed this in the newer revision. Only one self._http

stu-cao:
Fix is merged in a separate PR: https://github.com/PrimeIntellect-ai/verifiers/pull/2218

PR #2214, opened 2026-07-31
feat(v1): task-authored interception of model exchanges (@vf.intercept)

xeophon:
## Overview

Adds a task-authored policy layer for model exchanges in v1. Tasks can define `@vf.intercept` handlers, or attach ready-made policies, to inspect and control model traffic at the shared interception server before it reaches either the model or the harness.

The API stays typed and provider-neutral while the existing dialect layer owns wire-format changes for OpenAI Chat Completions, the Responses API, and Anthropic Messages.

## API

Handler direction is inferred from the `message` annotation:

- `AssistantMessage` handles a completed model response before the harness sees it.
- `ToolMessage` handles a new tool result before the model sees it in the next request.
- An untyped handler can observe both directions.

Handlers run in priority order and return:

- `None` to leave the exchange unchanged.
- `str` to replace an assistant response or tool result with inert text.
- `vf.Terminate(reason=..., reward=...)` to stop the rollout and record a policy reward.

```python
class RestrictedTask(vf.Task):
    block_git = vf.block_shell_commands(
        "git",
        reply="Git commands are disabled for this task.",
    )

    @vf.intercept(priority=10)
    async def redact_tool_output(
        self, message: vf.ToolMessage
    ) -> str | None:
        if message.name == "read_secret":
            return "The tool result was withheld by policy."
        return None
```

## Design

- **One interception boundary:** policies run in `RolloutSession` at the interception server, so they apply consistently to every harness that sends model traffic through v1.
- **Typed policy, native delivery:** handlers operate on canonical `vf` messages. Each `Dialect` performs the corresponding mutation in the provider's native request, response, or stream format.
- **Original and delivered messages remain distinct:** the trace preserves the model's original sampled message and token sequence for training and audit. A rewritten response is stored separately as `delivered_message`, representing what the harness actually received.
- **Request rewrites precede trace parsing:** rewritten tool results are parsed only after policy execution, keeping the model-visible prompt and recorded prompt aligned.
- **Termination preserves evidence:** terminating on a response commits the original sampled action but never delivers it to the harness. The termination reason, reward, and handler are recorded on the trace. Request-side termination happens before another model call is made.
- **Streaming remains protocol-correct:** streams stay live when no response policy needs classification. Intercepted streams are assembled, classified, and emitted as valid Chat Completions, Responses, or Anthropic events.
- **Retries remain atomic:** identical concurrent requests are coalesced and completed requests are replayed, preventing SDK retries from sampling or committing the same turn twice.
- **Provider-managed state cannot bypass policy:** response interception uses single-response, stateless modes where a provider-side conversation could otherwise retain the original response.

## Ready-made policies

`verifiers.v1.intercepts` is re-exported through `vf` and includes:

- `match_tool` for normalized matching across harness and provider tool names.
- `block_tool_calls`, `block_shell_commands`, `block_web_search`, and `block_code_search` for common deterministic policies.
- `block_with_judge` for rubric-based message classification.
- `disable_provider_tools` for removing matching provider-hosted tools and stale forced choices while preserving client-owned function tools, including same-name tools.

Every rewrite or termination is appended to `trace.interceptions`, including its direction, handler, target, reason, and bounded pre-policy value.

> [!NOTE]
> ### Add task-authored model exchange interception via `@vf.intercept`
> - Introduces a new `@intercept` decorator that task authors use to register handler functions that run at request or response time during rollouts, able to rewrite or terminate model exchanges.
> - Adds `verifiers/v1/intercepts/` with core types (`Interceptor`, `Terminate`, `InterceptRecord`, `InterceptOutcome`) and built-in policies: `block_tool_calls`, `block_shell_commands`, `block_web_search`, `block_code_search`, `disable_provider_tools`, and `block_with_judge` (async LLM-judged blocking).
> - The interception server in `verifiers/v1/interception/server.py` now buffers streamed responses, applies intercept handlers, and can replay buffered payloads for retried identical requests keyed by `(path, body-digest)`.
> - `MessageNode` gains a `delivered_message` field so traces distinguish between the sampled and the delivered (rewritten) assistant message; `Branch.delivered_messages` and `Trace.interceptions` persist the full interception record.
> - Harness resume now builds conversation context from `delivered_messages` instead of original sampled messages.
> - Replay skips re-scoring for traces terminated by an intercept and marks them `skipped: terminal interception`.
> - Risk: response intercepts enforce new constraints — continuing a rewritten response via `previous_response_id` is rejected, stateful `/v1/responses` requests are rejected, and `/v1/chat/completions` must request exactly `n=1`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 30ab8a8.</sup>
> 
>

### DIFF
--- a/verifiers/v1/__init__.py
+++ b/verifiers/v1/__init__.py
@@ -45,9 +45,9 @@ from verifiers.v1.errors import (
 from verifiers.v1.graph import MessageNode
 from verifiers.v1.harness import Harness
 from verifiers.v1.intercepts import (
+    Interceptor,
     InterceptRecord,
     InterceptResult,
-    Interceptor,
     Terminate,
     block_code_search,
     block_shell_commands,
--- a/verifiers/v1/__init__.py
+++ b/verifiers/v1/__init__.py
@@ -44,6 +44,19 @@ from verifiers.v1.errors import (
 )
 from verifiers.v1.graph import MessageNode
 from verifiers.v1.harness import Harness
+from verifiers.v1.intercepts import (
+    InterceptRecord,
+    InterceptResult,
+    Interceptor,
+    Terminate,
+    block_code_search,
+    block_shell_commands,
+    block_tool_calls,
+    block_web_search,
+    block_with_judge,
+    disable_provider_tools,
+    match_tool,
+)
 from verifiers.v1.judge import Judge, JudgeResponse, JudgeView
 from verifiers.v1.judges import (
     Criterion,
@@ -116,7 +129,7 @@ from verifiers.v1.utils.artifacts import (
     collect,
     restore,
 )
-from verifiers.v1.utils.decorators import metric, reward, stop, tool
+from verifiers.v1.utils.decorators import intercept, metric, reward, stop, tool
 from verifiers.v1.utils.git import (
     PATCH_CAP_BYTES as PATCH_CAP_BYTES,
 )
@@ -218,6 +231,19 @@ __all__ = [  # noqa: RUF022 - grouped by public API area
     "tool",
     "metric",
     "reward",
+    "intercept",
+    # interception
+    "InterceptRecord",
+    "InterceptResult",
+    "Interceptor",
+    "Terminate",
+    "block_code_search",
+    "block_shell_commands",
+    "block_tool_calls",
+    "block_web_search",
+    "block_with_judge",
+    "disable_provider_tools",
+    "match_tool",
     # errors
     "RolloutError",
     "EnvError",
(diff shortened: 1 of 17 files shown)


xeophon:
## Migrated review context

This PR carries the restacked interception work from #2211. Outstanding points retained for follow-up:

- Evaluate rollout refusal and stop conditions before request interceptors run.
- Complete retry/coalescing bookkeeping for streamed calls with request-only interceptors.
- Ignore nullable Anthropic wire fields when deriving preserved provider state and message identity.
- Avoid cloning and comparing an entire large request for request-only raw policies.

xeophon:
Superseded by the existing native interception stack #2179 (#2164, #2165, #2166, and #2178), which carries this work as reviewable layers.

PR #2241, opened 2026-08-04
fix(v1): preserve exact filenames during capture

### LINKED ISSUE #2196: v1: harden capture_patch handling of pre-existing untracked filenames
## Context

Post-merge review of #2144 found two filename-handling gaps in `snapshot_untracked()` / `capture_patch()` now on `main`:

1. Pre-existing paths are passed to `git reset` as ordinary pathspecs, so names containing pathspec metacharacters can unstage unrelated agent changes.
2. `git ls-files -z` output is decoded with `errors="replace"` before splitting, so a non-UTF-8 filename is corrupted and the real pre-existing file is not excluded from the captured patch.

These comments appeared on #2152 only while GitHub was retargeting the stack after #2144 was squash-merged; `verifiers/v1/utils/git.py` is not part of #2152's current diff.

## Done when

- Ignore entries are passed to Git as literal pathspecs.
- NUL-delimited filenames survive without lossy text decoding, or unsupported byte sequences fail explicitly before producing a wrong patch.
- Focused coverage includes glob metacharacters and non-UTF-8 filename bytes.

gtnv:
Closes: https://github.com/PrimeIntellect-ai/verifiers/issues/2196

untracked filenames lost identity in 2 spots: runtime output (replaced by invalid utf-8) and git parsed savednames (pathspecs). now, snapshotting now reads the raw bytes and rejects any unsupported filenames before capture. however, reset will treat every saved name literally. added test for coverage.

---

> [!NOTE]
> **Medium Risk**
> Touches finalize-time patch capture used for SWE grading; behavior changes only for edge-case filenames but can affect which paths are ignored or whether setup fails on invalid UTF-8 names.
> 
> **Overview**
> Fixes two ways pre-agent filenames could be corrupted or mis-handled when building the agent diff for grading.
> 
> **`snapshot_untracked`** no longer relies on decoded command stdout. It writes `git ls-files --others` NUL output to a unique temp file, reads it with the byte-preserving runtime API, and decodes paths as UTF-8 so invalid names raise instead of being silently rewritten.
> 
> **`capture_patch`** unstaging of ignored paths now runs `git --literal-pathspecs reset`, so names that look like pathspec magic (e.g. `:(glob)*.py`) stay literal and are not expanded when excluded from the cached diff.
> 
> Adds integration coverage for literal pathspec handling during capture and for `UnicodeDecodeError` on non-UTF-8 untracked names.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c87239c1857a5a53fe111de7082d382667bf4396. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Preserve exact filenames during capture in `verifiers.v1.utils.git`
> - `_DIFF` unstage step now runs `git --literal-pathspecs reset -q -- "$@"` so pathspec arguments are interpreted literally instead of as glob patterns.
> - `snapshot_untracked` now writes `git ls-files --others --exclude-standard -z` output to a temp file, reads raw bytes, and splits on `b"\0"` to avoid lossy stdout decoding from `runtime.run`.
> - Risk: undecodable filenames will now raise a decode error instead of returning a silently-altered replacement path; check `snapshot_untracked` callers for any dependency on lossy behavior.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0315217.</sup>
> 
>


verifiers/v1/utils/git.py
```diff
@@ -95,12 +95,26 @@ async def snapshot_untracked(runtime: Runtime, env: dict | None = None) -> list[
...
+                "vf-snapshot-untracked",
+                snapshot_path,
+            ],
+            env or {},
+        )
+        if result.exit_code != 0:
+            return []
+        raw = await runtime.read(snapshot_path)
+    finally:
+        await runtime.run(["rm", "-f", snapshot_path], env or {})
```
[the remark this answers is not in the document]
gtnv: Invalid: cleanup exceptions mean the runtime transport died before the agent ran, so they must propagate.

tests/v1/test_git.py
```diff
@@ -0,0 +1,53 @@
+# Pre-agent filenames must retain their identity through patch capture.
```
xeophon: lets remove this unit test

xeophon (changes requested):

tests/v1/test_git.py
gtnv: done @xeophon thx

PR #2247, opened 2026-08-04
feat: add sync parameter to prepare_uv_script / run_uv_script for offline/pre-built images

Alex-Shan:
## Problem

`prepare_uv_script` unconditionally runs two network-dependent steps **every time a runtime starts**, even when using a pre-built Docker image:

1. `_ENSURE_UV` — pip/curl/wget bootstrap (`~17s` cold, still touches PyPI when warm)
2. `uv sync --script` — dependency resolution from PyPI (`~40-50s` cold, `~2-5s` index refresh even when cached)

For **pre-built images** (e.g. the [Dockerfile pattern](https://github.com/primeintellect-ai/verifiers/blob/main/environments/gsm8k_v1/image/Dockerfile)) that already contain uv and the script's venv baked into the image layer, this is pure overhead. In **network-restricted Docker allowlists** (common in RL training clusters where the network is cut before scoring), these steps fail outright.

## Solution

Add `sync: bool = True` parameter to `prepare_uv_script` and `run_uv_script`.

| `sync=True` (default, today's behavior) | `sync=False` (for pre-built images) |
|---|---|
| `_ENSURE_UV` + `uv sync` + `uv python find` | `mv` + `uv python find` only |
| Requires network | **Zero network, <0.2s** |

When `sync=False`, the method assumes both `uv` and the script's venv are already present (pre-baked into a Docker image) and only does the atomic write + interpreter discovery.

### Error message when venv is missing

If `sync=False` is used but the venv hasn't been pre-built, the error includes a hint:

> failed to prepare uv script: ... (sync=False assumes a pre-baked venv — did you build the image with `uv sync --script` for this script?)

## Backward compatibility

Fully backward-compatible — `sync=True` is the default, preserving today's behavior exactly.

## Related

- The [gsm8k_v1 Dockerfile](https://github.com/primeintellect-ai/verifiers/blob/main/environments/gsm8k_v1/image/Dockerfile) already demonstrates the pre-bake pattern
- `_ENSURE_UV` is defined at module level in `base.py:38-42` and also requires network

Co-Authored-By: Claude <noreply@anthropic.com>

---

> [!NOTE]
> **Low Risk**
> Default `sync=True` preserves existing behavior; `sync=False` is opt-in and only affects script preparation paths that explicitly use it.
> 
> **Overview**
> Adds a **`sync`** parameter (default **`True`**) to **`prepare_uv_script`** and **`run_uv_script`** on the runtime base so PEP 723 script setup can skip network-heavy work when **`uv`** and the script venv are already in the image.
> 
> With **`sync=True`**, behavior is unchanged: atomic write, **`_ENSURE_UV`**, **`uv sync --script`**, then **`uv python find`**. With **`sync=False`**, only the write and **`uv python find`** run—intended for Docker images that pre-bake dependencies via **`uv sync --script`** at build time and for scoring when egress is blocked.
> 
> Failures under **`sync=False`** append a hint that a pre-baked venv was expected.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8adb248704ffb72f2d9303801ea2a4a021056f4b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add `sync` parameter to `Runtime.prepare_uv_script` and `run_uv_script` for offline/pre-built images
> Adds a `sync` boolean parameter (default `True`) to `prepare_uv_script` and `run_uv_script` in [base.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2247/files#diff-0811c040a163eb5a3212208f1e3a6c6388d7690fcc4780c081cc7754036232b9). When `sync=False`, the methods skip uv installation and dependency syncing, instead relying on uv and the script's venv being pre-installed, and locate the interpreter directly via `uv python find`. Error messages in `sync=False` mode include a hint about the pre-baked venv assumption. Default behavior is unchanged.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b0bcf1e.</sup>
> 
>

### DIFF
--- a/verifiers/v1/runtimes/base.py
+++ b/verifiers/v1/runtimes/base.py
@@ -230,7 +230,18 @@ class Runtime(ABC):
         self,
         script: str | bytes,
         env: dict[str, str] | None = None,
+        sync: bool = True,
     ) -> list[str]:
+        """Write *script* into the runtime and return the argv to execute it under the
+        script's venv interpreter.
+
+        When *sync* is True (the default), ``uv`` itself and the script's PEP 723
+        dependencies are installed / synced before the interpreter is discovered —
+        that requires network access.  When *sync* is False the method assumes that
+        both ``uv`` and the script's venv are **already present** in the runtime (e.g.
+        pre-baked into a Docker image) and only runs ``uv python find`` to locate the
+        interpreter — a fast, offline-only operation.
+        """
         data = script.encode() if isinstance(script, str) else script
         digest = hashlib.sha256(data).hexdigest()
         path = f"/tmp/vf-scripts/{digest}.py"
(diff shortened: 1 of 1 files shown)


Alex-Shan:
Good catch — `sync=False` was missing the `$HOME/.local/bin` PATH export that `_ENSURE_UV` and the run-wrapper both set up. Fixed in [b0bcf1e](https://github.com/Alex-Shan/verifiers/commit/b0bcf1e22):

1. Added `export PATH="$HOME/.local/bin:$PATH"` in the `sync=False` command, matching what `_ENSURE_UV` already does. Images that install uv via `pip install --user` or the standalone installer (both target `~/.local/bin`) now find it.
2. Broadened the error hint to mention both possible causes: uv not on PATH **or** venv not pre-baked.

Your specific Dockerfile (`pip install --no-cache-dir -q -U uv` without `--user`) installs system-wide so it already worked — but the fix covers all installation patterns consistently.

PR #2299, opened 2026-08-08
Block provider network escapes in restricted runtimes

xeophon:
## Summary

- enforce each rollout session’s effective runtime network restriction before a native model request reaches the provider
- remove blocked provider-side URLs, hosted tools, MCP, and provider-held state from restricted requests, then add a value-free runtime policy error to the model prompt so it can recover
- cover Responses, Chat Completions, Anthropic Messages, streaming requests, and Anthropic count_tokens
- keep intrinsically networked search models as a typed HTTP 400 configuration failure because their hosted access cannot be removed from an individual request

## Security model

Any restricted runtime policy disables provider-side fetches and hosted capabilities. Provider DNS and redirects happen outside the sandbox firewall, so a runtime allowlist cannot safely authorize a provider-side URL.

Mediation removes protocol-defined external capability fields before the upstream call. The model sees only structural field paths in the policy notice; URLs, file IDs, MCP authorization, and other request values are never copied into the notice. The sanitized effective request is also what Verifiers parses and records.

## Harness behavior

Recoverable policy violations no longer return an HTTP error to the harness. The provider receives a safe request containing the policy notice, and the harness receives the model’s ordinary response, so all harnesses can continue without dialect-specific error handling. Only a capability intrinsic to the configured model remains a hard error.

## Testing

- full local V1 suite passed; PRIME_API_KEY-gated cases skipped
- full local non-V1 suite passed
- Chat, Responses, and Anthropic mediation passed in streaming and non-streaming modes
- Anthropic count_tokens mediation passed
- Ruff format and check passed
- ty check passed
- commit and push hooks passed

The previous live V1 CI failure was an unrelated transient 429 during rubric scoring; 58 other live tests passed.

---

> [!NOTE]
> **High Risk**
> Touches security-critical request mediation on every intercepted model call and changes client/interception contracts; incorrect stripping or policy composition could break harnesses or leak egress.
> 
> **Overview**
> **Blocks provider-side network escapes** when a rollout’s effective runtime policy is restricted. The interception server applies overrides, then dialect-specific `mediate_external_capabilities` on Chat, Responses, and Anthropic (including streaming and `count_tokens`), removing hosted tools, MCP, remote URLs/file IDs, web search, and similar fields while injecting a value-free user notice so the harness still gets a normal model response.
> 
> **Policy plumbing:** `NetworkPolicyConfig` moves to `verifiers/v1/configs/runtime`; task/runtime allowlists now intersect on exact shared rules (blocks always merge). `RolloutSession` carries the resolved policy; clients take a fully prepared body (`sampling` only—no separate `model` param).
> 
> **Observability:** `ModelCall.policy` (`PolicyEvent`) records mediation paths; Responses parsing raises typed errors for failed terminal statuses; Harbor docs note that allowlists govern runtime egress only, not provider-resolved resources.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3c7fb451cc27baa65b31ee3137264aab33fff599. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Block provider network escapes in restricted runtimes by mediating AI provider requests against rollout network policy
> - Adds `NetworkPolicyConfig.permits` to test egress URLs against allow/block rules, and propagates the resolved policy through `RolloutSession` into the interception pipeline.
> - Implements `mediate_external_capabilities` on all three dialects ([anthropic.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2299/files#diff-abc86b3ad07c2d6ab09cd923f50271c4f6c4cb928a9bb82c12f6e97aac17b04d), [chat.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2299/files#diff-5d2fcbf59a03489b63b0626038a853eee89543b591074597161ee29aab901280), [responses.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2299/files#diff-73f89a1b25536fd7a806555e765120f5189cb934eff813f4c7900838efa56b5a)): removes disallowed tools, content references (images, files, hosted IDs), and provider-side features (web search, MCP servers, audio voices), then injects a policy notice into the earliest user message when removals occur.
> - Rollouts targeting models with intrinsic provider-side network capabilities (e.g. built-in search) now fail early with `InterceptionError` before the request is sent.
> - Adds `PolicyEvent` to `ModelCall` traces so mediation decisions and affected request paths are recorded alongside each call.
> - Refactors `Client.get_response`/`relay` signatures: `model` is removed as a separate parameter and `sampling_args` is renamed to `sampling`; callers must supply a fully mediated body. Risk: any custom `Client` subclass must update its method signatures.
>
> 
> #### Changes since #2299 opened
>
> - Replaced `_HOSTED_TOOL_TYPES` frozenset with `_HOSTED_TOOL_TYPE` compiled regex pattern matching `file_search|mcp|code_interpreter|programmatic_tool_calling|image_generation|web_search_preview(?:_\d{4}_\d{2}_\d{2})?` to support date-suffixed web_search_preview variants, updated `_mediate_tools` function to use regex fullmatch for hosted tool detection with string type guard and special handling for `tool_search` and `shell` kinds, and modified `ResponsesDialect.mediate_external_capabilities` method to reject dict `tool_choice` entries matching hosted-tool or web tool regex patterns and to validate `shell` or `tool_search` tool_choice only when matching tool types exist in the tools array [fde99e0]
> - Added exception handling in `network_rule_matches` function to catch `ValueError` when accessing parsed port and return `False` for rules with invalid port specifications [fde99e0]
> - Renamed module-level constants and helper function from private to public exports [0962376]
> - Moved `NetworkPolicyConfig` class and `network_rule_matches` function from `verifiers.v1.runtimes.base` module to `verifiers.v1.configs.runtime` module [b0cb17b]
> - Redesigned capability mediation across all API dialects to unconditionally block provider-resolved remote resources and hosted tools in restricted runtimes [938a5e9]
> - Changed network policy composition in `NetworkPolicyConfig.with_task_network_policy` to retain only exact shared allow rules and always combine blocklists [938a5e9]
> - Introduced syntax error in `TaskData` model by inserting free-form descriptive text between field docstrings outside string literals [938a5e9]
> - Enhanced ACP session transcript replay to trim prior assistant messages based on replay status rather than session newness [938a5e9]
> - Extended resumed MCP call test to require two codewords in multi-message delta interaction [938a5e9]
> - Added typed exception handling for streaming errors in interception server turn creation and enforced rollout model selection for auxiliary endpoints [938a5e9]
> - Improved error surfacing in `response_from_wire` by raising typed exceptions for non-completed response statuses [938a5e9]
> - Removed preflight check in `Rollout.open` that failed rollouts based on intrinsic provider-side model capabilities under restricted policies [938a5e9]
> - Refactored URL scheme detection in egress proxy to use string prefix checks instead of URL parsing [938a5e9]
> - Fixed `ChatDialect.apply_overrides` merge order to ensure selected model remains authoritative over sampling payload fields [938a5e9]
> - Updated documentation to clarify network policy behavior for allowlist retention, blocklist combination, and provider resource restrictions [938a5e9]
> - Refactored message trimming logic in `verifiers.v1.acp.runner` module to key directly off `is_new` parameter [a248e75]
> - Simplified `tests.v1.fixtures.echo_acp_resume_v1` test fixture to use single codeword instead of two [a248e75]
> - Removed conditional guards in `AnthropicDialect.mediate_external_capabilities`, `ChatDialect.mediate_external_capabilities`, and `ResponsesDialect.mediate_external_capabilities` methods to unconditionally append user notices via `append_user_notice` to mediated request messages regardless of whether restricted capabilities were detected [3c7fb45]
> - Updated docstrings for `append_user_notice` function and `Dialect.mediate_external_capabilities` abstract method to reflect that stable restricted-network context is added on every call rather than conditionally [3c7fb45]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9e94611.</sup>
> 
>


snimu (dismissed):

hallerite (commented):
important PR to get in. I would prefer if we can return a stable structured error to let the different harnesses optionally recover (we should def not put the onus of crafting a response to let the model know why the request was rejected on the proxy)

otherwise, lgtm!

tests/v1/test_e2e.py
mikasenghaas, line 1: most of these tests dont seem to be e2e?

verifiers/v1/dialects/anthropic.py
```diff
@@ -41,6 +47,66 @@
     "stop_sequence": "stop",
 }
 THINKING = ("thinking", "redacted_thinking")
+_CLIENT_TOOL_TYPES = frozenset(
+    {
+        "bash_20241022",
+        "bash_20250124",
```
mikasenghaas: can we regex match so we dont regress when new tool version drops?

verifiers/v1/dialects/anthropic.py
```diff
@@ -274,6 +393,102 @@ class AnthropicDialect(Dialect[dict, AnthropicMessage]):
     upstream_path = "/v1/messages"
     response_type = ModdedAnthropicMessage
 
+    def external_capability(self, body: dict) -> str | None:
```
mikasenghaas: can we use the dialect-types for this kinda stuff?

verifiers/v1/dialects/base.py
```diff
@@ -28,6 +28,24 @@
 logger = logging.getLogger(__name__)
 
 
+def is_remote_url(value: object) -> bool:
```
mikasenghaas: object typing🥲

verifiers/v1/dialects/base.py
```diff
@@ -28,6 +28,24 @@
...
+    """Whether a native URL field makes the provider resolve data outside the request."""
+    return (
+        isinstance(value, str) and bool(value) and not value.lower().startswith("data:")
```
mikasenghaas: this seems very broad. basically only `data:` prefix is accepted as local url?

verifiers/v1/session.py
```diff
@@ -65,6 +65,9 @@ def reached(self, trace: Trace) -> str | None:
 class RolloutSession:
     ctx: ModelContext
     trace: Trace
+    network_restricted: bool
```
mikasenghaas: if this is a bool how can e.g. a whitelist of websites propagate to external resources

mikasenghaas (commented):
hmm i really dont like us being in charge of network restrictions on *any external resource* (any inference deployment, api, mcp server, etc. etc.). in my opinion this is a battle we really cannot win. plus it adds a huge amount of line noise and unmaintainable ai slop into our codebase. i would much rather maintain a list of known ways to "escape" the sandbox and tune our interception machinery to reliably detect these things

verifiers/v1/clients/client.py
```diff
@@ -33,29 +33,26 @@ async def get_response(
         self,
         dialect: Dialect,
         body: dict,
-        model: str,
```
mikasenghaas, line 36: why no model anymore? do we let the harness decide?

verifiers/v1/dialects/anthropic.py
```diff
@@ -41,6 +53,10 @@
     "stop_sequence": "stop",
 }
 THINKING = ("thinking", "redacted_thinking")
+_WEB_TOOL_TYPE = re.compile(r"(?:web_search|web_fetch)_\d{8}").fullmatch
```
mikasenghaas: no _

verifiers/v1/dialects/responses.py
```diff
@@ -45,6 +57,16 @@
 )
 # Sampling knobs the eval owns, in this format's shape (Responses uses `max_output_tokens`).
 _SAMPLING_KEYS = frozenset({"temperature", "top_p", "max_output_tokens", "max_tokens"})
+_WEB_TOOL_TYPE = re.compile(r"web_search(?:_\d{4}_\d{2}_\d{2})?").fullmatch
```
mikasenghaas: same here, can also fix the existing ones

mikasenghaas (commented):
can we run some example traces for each external tool with no block and block where we explicitly ask the model to do use the blocked tool and then check that we a) block the tool and b) so its clear what our mediation strategy acc is

also we choose to only care abt external tools through infernece. we deliberately dont try to block e.g. mcp (this is user respnosibility). should we hav a small note on this in the docs?

verifiers/v1/session.py
```diff
@@ -17,6 +17,7 @@
 from pydantic import TypeAdapter
 
 from verifiers.v1.clients import Client, ModelContext
+from verifiers.v1.runtimes.base import NetworkPolicyConfig
```
mikasenghaas: configs should be in `verifiers.v1.configs`

mikasenghaas (dismissed):

mikasenghaas (approved):

PR #2337, opened 2026-08-12
v1: reject duplicate artifact sources before collection

### LINKED ISSUE #2198: v1: reject or normalize duplicate artifact sources before collection
## Context

`collect()` currently archives every artifact declaration and stores results in a `dict[source, archive]`. Duplicate resolved sources therefore consume the byte budget multiple times while later entries silently overwrite earlier archives; differing `exclude` lists make the result order-dependent.

This was found in post-merge review of #2144 and appeared on #2152 only during its automatic retarget to `main`; `artifacts.py` is not part of #2152's current diff. The broader transport security boundary is tracked in #2194 and the durable archival contract in #2189.

## Done when

- Duplicate resolved sources are rejected with a clear error, or normalized with explicitly documented merge semantics before any archive is created.
- The byte budget corresponds exactly to the returned collection.
- Differing exclusion rules cannot silently overwrite one another.

AyushPatel94:
Fixes #2198.

## What

`collect()` keys its result by resolved source. Two declarations that resolve to the same path therefore:

- archive the same tree twice, charging `budget` for both, so the byte ceiling no longer corresponds to what is returned;
- overwrite each other in `collected`, silently discarding the first archive;
- with differing `exclude` patterns, produce a result that depends on declaration order.

The docstring already promised "a path cannot be collected twice" — nothing enforced it.

## Approach

The issue allows either rejection or documented merge semantics. This takes rejection: merge semantics for two differing `exclude` lists have no obviously correct answer (union changes what each declaration asked for; intersection silently widens the archive), and rejecting is what the existing docstring already claims.

Validation runs **after** workdir resolution and **before** the convention sweep, so `./out`, `out/`, `/work//out` and `/work/out` are all caught, and nothing is tarred before the rejection — which is what keeps the budget equal to the returned collection.

`ValueError` rather than `RuntimeError`: this is an invalid declaration detected before the runtime is touched, and it matches the surrounding Harbor conversion path, which already raises `ValueError` for invalid declarations.

## Verified

Checked with a throwaway script (not committed, per `AGENTS.md` on unit tests):

- duplicates rejected for identical strings, `./out`, absolute vs relative, trailing slash, double slash — each raising `ValueError`, naming the resolved path, **with no archive created**;
- differing `exclude` patterns raise rather than silently picking one;
- distinct sources still collect normally, convention sweep included;
- returned bytes exactly equal archived bytes (budget accounting is now exact);
- single declaration and the no-declarations case unchanged.

`ruff check` and `ruff format --check` clean on the touched file.

I could not run `uv run pytest tests/` end to end: the dev group builds `pycosat` (via `reasoning-gym`), which needs a C toolchain I do not have here. The change is confined to one pure function with no I/O, so I would appreciate CI confirming it.

## Blast radius

A Harbor task that declares the same source twice now fails loudly instead of silently dropping one archive. That is the intended behaviour change, but it is a behaviour change for any such task that exists today.

## Out of scope

`..` segments are deliberately not collapsed — doing so lexically is only sound without symlinks — so two sources differing only by a `..` are not treated as duplicates. Happy to revisit if you would rather resolve those too.

Also happy to switch to normalization, change the exception type, or move the check into a validator on the declaration model if you prefer any of those.

> [!NOTE]
> ### Reject duplicate artifact sources in `artifacts.collect` before archiving begins
> Adds a precondition check in [`artifacts.collect`](https://github.com/PrimeIntellect-ai/verifiers/pull/2337/files#diff-0d082bad54931be61d7ef55ddce0757034be3013f69ae71cb803cd8416df9138) that calls the new helper `_reject_duplicate_sources` to scan resolved artifact paths before any archiving occurs. If two declarations resolve to the same source path, a `ValueError` is raised identifying both the original and conflicting declarations. Risk: existing callers with duplicate artifact sources will now fail with `ValueError` instead of silently overwriting or double-archiving files.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 834e16e.</sup>
> 
>

### DIFF
--- a/verifiers/v1/utils/artifacts.py
+++ b/verifiers/v1/utils/artifacts.py
@@ -54,6 +54,7 @@ async def collect(
         a.model_copy(update={"source": str(workdir / a.source)})
         for a in artifacts or []
     ]
+    _reject_duplicate_sources(artifacts or [], declared)
     convention = PurePosixPath(ARTIFACTS_DIR)
     declared_paths = [PurePosixPath(artifact.source) for artifact in declared]
     if convention in declared_paths:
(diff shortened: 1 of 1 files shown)


AyushPatel94:
Closing this — I opened it without my collaborator's sign-off, and the verification behind it was thinner than the PR implied: the logic was checked against a hand-written fake runtime, not against the repo's own test suite. Apologies for the noise on #2198; leaving the issue unclaimed for whoever picks it up.

PR #2340, opened 2026-08-12
fix(legacy): align RubricGroup.score_group metrics and advantages with score_rollout

crazywriter1:
Supersedes #2009 — same fix, rebased onto current `main` after #2303 moved v0 rubrics to `verifiers/legacy/`.

## Summary
- `RubricGroup.score_group` summed child rewards correctly but left per-child `advantage` on state and often left stale trajectory `advantage`/`reward` from earlier children.
- Metrics were merged into preexisting keys, unlike `score_rollout` and `Rubric.score_group`, which fully replace `state["metrics"]`.
- After aggregation: recompute `advantage` from summed rewards, fill trajectory fields only when still `None`, and replace metrics entirely.
- Restore `advantage` and trajectory fields between child rubrics so partial writes do not leak.

## Risk / notes
- Aggregate **rewards** were already correct; this is contract/consistency cleanup for legacy group scoring.
- Default `Rubric.has_advantages` is still `False`; trainers that recompute advantages from rewards are largely unaffected.

## Test plan
- [x] Manual: children `[[1,0],[0,1]]` → rewards `[1,1]`, advantages `[0,0]`
- [x] Manual: preset trajectory step reward preserved; unset steps still get aggregate fill
- [x] Manual: preexisting stale metrics cleared after group scoring
- [x] Manual: child with empty metrics → `state["metrics"] == {}`

> [!NOTE]
> ### Fix `RubricGroup.score_group` to align metrics and advantages with `score_rollout`
> - Adds an early return with a warning when `score_group` receives zero states.
> - Snapshots each state's reward, metrics, advantage, and per-step trajectory values before invoking child rubrics, then restores them after each child so all rubrics see identical inputs.
> - Accumulates rewards and metrics across child rubrics, then writes the final aggregated reward to each state and sets `state['advantage']` as `aggregated_reward - average_reward`.
> - Per-step trajectory reward and advantage are only written when currently `None`, matching `score_rollout` behavior.
> - Behavioral Change: metrics aggregation now always writes a fresh dict per state rather than conditionally merging into an existing one.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b07b7f1.</sup>
> 
>

### DIFF
--- a/verifiers/legacy/rubrics/rubric_group.py
+++ b/verifiers/legacy/rubrics/rubric_group.py
@@ -114,12 +114,31 @@ class RubricGroup(Rubric):
     async def score_group(self, states: list[State]):
         """
         Evaluate all reward functions in-place for a group of rollouts.
+
+        Mirrors ``score_rollout`` for metrics (full replace) and ``Rubric.score_group``
+        for advantages: child rubrics may write per-rubric advantages/trajectory fields,
+        so the group recomputes state advantages from aggregated rewards and only fills
+        trajectory fields that are still ``None`` (preserving intentional step values).
         """
-        aggregated_rewards = [0.0] * len(states)
+        num_states = len(states)
+        if num_states == 0:
+            self.logger.warning("No states to score")
+            return
+
+        aggregated_rewards = [0.0] * num_states
         aggregated_metrics: dict[str, list[float]] = {}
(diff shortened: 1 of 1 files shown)


xeophon:
we will remove legacy very soon and don't intend to change its behaviour

PR #2342, opened 2026-08-12
fix(v1): collect artifacts for isolated agentic-judge

### LINKED ISSUE #2195: v1: make isolated agentic-judge collect generic task artifacts and honor judge image
## Context

Post-merge review of #2144 found two coupled gaps in `AgenticJudgeEnv(share_runtime=False)` now on `main`:

1. `AgenticJudgeEnv.run()` calls `agents.solver.run(task)` without retaining the solver runtime, then `JudgeTask.from_trace()` restores `solution.state.artifacts`. Only `HarborTask.finalize()` currently populates that field; an ordinary task using `TaskData.artifacts` or `/logs/artifacts/` reaches the fresh judge box with an empty artifact set.
2. `JudgeTask.from_trace()` copies `solution.task.data.image` onto the judge task even in isolated mode. Runtime compilation gives that task image precedence over `judge.runtime.image`, so a purpose-built judge image is silently replaced by the solver image despite the config contract that the judge runtime is honored when sharing is disabled.

These findings landed on #2152 only because its base branch disappeared after #2144 was squash-merged; they are inherited `main` behavior, not part of #2152's Harbor diff.

## Suggested direction

- In isolated mode, provision the solver runtime explicitly, run the solver in it, and collect `task.data.artifacts` (including the convention sweep) after task finalization but before leaving the runtime context.
- Pass that collected set to the isolated judge task rather than assuming a task subclass populated transient trace state.
- Do not set the isolated judge task's `image` from the solver row; let the configured judge runtime image win. Preserve only task-level fields that are intentionally shared.
- Keep shared mode unchanged.

## Done when

- A plain `Task` that writes to `/logs/artifacts/` is restored into an isolated judge sandbox.
- Declared artifact paths work without a Harbor-specific `finalize()` override.
- An explicit `judge.runtime.image` is the image used in isolated mode.
- Shared mode continues to use the live solver workspace and runtime.

crazywriter1:
## Summary
- Fixes #2195 (artifact half): `IsolatedAgenticJudgeEnv` now owns the solver box, collects declared + convention artifacts after finalize, and hands that set to `JudgeTask.from_trace` instead of relying on `solution.state.artifacts`.
- Shared mode is unchanged; solver-policy inheritance for the judge runtime is unchanged (intentional after #2234).
- Skips the issue's "honor judge.runtime.image" ask: #2234 deliberately derives both modes from the solver policy, and dropping the task image would break delta-sized artifact restore (grading box boots from the task image).

## Behavioral note
Owning the solver box via `agents.solver.provision(task)` + `run(..., runtime=box)` disables **agent-level** retries on the solver (borrowed boxes are not retried — same rule shared mode already follows). Episode-level `EnvConfig.retries` still apply. This is required so collection can run after finalize and before teardown.

## Test plan
- [x] `uv run ruff check verifiers/v1/envs/agentic_judge/env.py`
- [x] `uv run pytest tests

> [!NOTE]
> ### Fix artifact collection for isolated agentic judge runs
> - `IsolatedAgenticJudgeEnv.run` now provisions an explicit runtime box for the solver and collects artifacts from it when the solver does not publish any to `solution.state.artifacts`.
> - `JudgeTask.from_trace` accepts an optional `artifacts` parameter; for `share_runtime=False`, it embeds the provided artifacts (or an empty dict) instead of reading from `solution.state.artifacts`.
> - Behavioral Change: isolated judge runs previously received no artifacts or stale artifacts from the solution state; they now receive artifacts collected directly from the solver's runtime.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 60e3277.</sup>
> 
>

### DIFF
--- a/verifiers/v1/envs/agentic_judge/env.py
+++ b/verifiers/v1/envs/agentic_judge/env.py
@@ -137,13 +137,15 @@ class JudgeTask(vf.Task):
         solution: vf.Trace,
         config: "JudgeTaskConfig",
         share_runtime: bool = True,
+        artifacts: dict[str, bytes | None] | None = None,
     ) -> "JudgeTask":
         """Mint the judge's task from the solver's finished trace.
 
         `share_runtime` selects both the workspace note and artifact transport. In
         the solver's box the published artifacts are already on disk, so none
-        travel; a fresh box gets the collected set, restored by `setup` at the
-        paths they had.
+        travel; a fresh box gets `artifacts` — the set the env collected off the
+        solver's box before tearing it down — restored by `setup` at the paths
+        they had.
         """
         solved = solution.task.data
         files = {TRACE_FILE: json.dumps(solution.to_record()).encode()}
@@ -170,7 +172,7 @@ class JudgeTask(vf.Task):
                 network_block=solved.network_block,
             ),
             files=files,
-            artifacts={} if share_runtime else solution.state.artifacts,
+            artifacts={} if share_runtime else dict(artifacts or {}),
         )
 
     async def setup(self, trace: vf.Trace, runtime: vf.Runtime) -> None:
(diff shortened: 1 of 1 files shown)


crazywriter1:
Friendly ping. Whenever you have a moment, would love a look at this fix for #2195 (artifact collection in isolated `agentic-judge`). Happy to address any feedback. @hallerite @mikasenghaas

xeophon:
implemented slightly differently in #2468, thank you for the contribution!!

PR #2363, opened 2026-08-14
Prevent --clean from deleting output roots

xeophon:
## Overview

Prevent eval and validate cleanup from treating the shared output directory as an individual run directory.

## Details

Both CLIs now require the resolved run path to be a strict child of `output_dir` before recursively deleting it. Paths that resolve to the output root, including `run.dir = "."`, exit with a clear error while preserving sibling runs.

The behavior remains unchanged for valid child run directories and for paths that already escape the configured output directory.

### DIFF
--- a/tests/v1/test_e2e.py
+++ b/tests/v1/test_e2e.py
@@ -62,35 +62,6 @@ USER_RUNTIMES = [
 ]
 
 
-@pytest.mark.parametrize("command", ["eval", "validate"])
-def test_clean_refuses_output_root(command, tmp_path, monkeypatch):
-    """Cleaning one run must never delete the directory grouping every run."""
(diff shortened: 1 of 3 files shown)


tests/v1/test_e2e.py
```diff
@@ -61,6 +61,36 @@ def pair(a: str, b: str, id: str, *extra_marks):
     pytest.param("modal", marks=[mark.modal], id="harness-in-modal"),
 ]
 
+
+@pytest.mark.parametrize("command", ["eval", "validate"])
```
hallerite: can remove test imo

xeophon: Removed in a0b2a5565.

PR #2369, opened 2026-08-14
Support training with streaming harnesses

hallerite:
## Summary

- support renderer-backed training for harnesses that require `stream=true`
- canonicalize Chat Completions, Responses, and Anthropic requests for rendering, then let each dialect serialize the completed response in its native JSON and SSE shape
- open buffered streams immediately and emit dialect-valid, content-free progress events while generation runs
- commit the authoritative renderer result before terminal events, preserving exact completion token IDs and logprobs

## Why

Streaming harness calls enter the interception server's relay path. Eval clients can relay provider SSE, but the training client previously only implemented non-streaming `get_response()`, so those harnesses failed before generation.

Renderer generation remains intentionally buffered. The engine's completed result is still the source of truth for tokens and logprobs; Verifiers only adapts its delivery to the streaming protocols expected by harness SDKs.

Ordinary SSE comments are insufficient for clients that reset liveness only after decoding an API event. During generation, Verifiers therefore sends native no-content events: an empty Chat delta, Responses lifecycle events, or an empty Anthropic message delta. No fake model text or tool arguments are introduced.

Grok Build's newer content watchdog intentionally ignores lifecycle and empty-delta events. The companion Grok harness PR, #2366, sets that independent watchdog beyond Verifiers' maximum remote runtime instead of fabricating model tokens.

## Scope

- renderer output is buffered, not token-streamed
- Chat Completions, Responses, and Anthropic are supported through the dialect interface
- eval-provider streaming remains unchanged
- request/response rewrites and trace commits happen before the client receives a terminal event

## Validation

- `uv run pytest tests/` — 914 passed, 75 skipped
- `uv run pytest tests/v1 -m 'not e2e'` — 70 passed
- temporary protocol matrix validated text, reasoning, tool calls, and truncation across all three dialect parsers and the installed OpenAI/Anthropic SDK event models
- exact Grok Build 1.0.3 behavioral smoke confirmed parsed Responses lifecycle events keep its pinned watchdog alive
- changed-file pre-commit hooks passed
- pre-push Ruff, format, and `ty` hooks passed

Live Prime requests reached the inference service, but the currently available model deployments did not provide a complete training smoke: attempts encountered a removed model, an unsupported generate route, missing generate logprobs, and a temporary 502. The deterministic protocol and full repository suites above are green.

The all-files markdown hook reports 321 pre-existing MD033 findings in the untouched `verifiers/legacy/envs/experimental/composable/tasksets/swe/README.md`.

The commit is signed with the contributor's verified SSH signing key.

### DIFF
--- a/verifiers/v1/clients/client.py
+++ b/verifiers/v1/clients/client.py
@@ -20,11 +20,17 @@ router pins a rollout's turns to one engine and its growing prefix stays KV-cach
 
 @dataclass
 class RelayReply:
-    """A relayed upstream response: content type, complete SSE events, and connection cleanup."""
+    """A streaming response: complete SSE events, cleanup, and optional result future.
+
+    Eval clients relay provider bytes and leave ``result`` unset. Renderer-backed training
+    clients synthesize SSE from a completed generation and retain its result future here so the
+    trace keeps the exact token ids and logprobs while the server can send keepalives meanwhile.
+    """
 
     content_type: str
     chunks: AsyncIterator[bytes]
     close: Callable[[], Awaitable[None]]
+    result: Awaitable[Response] | None = None
 
 
 class Client(ABC):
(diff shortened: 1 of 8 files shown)


hallerite:
still a work in progress.

PR #2374, opened 2026-08-16
Revert nano-rlm harness pin, default version back to main

samsja:
## Summary

- revert the default `RLMHarnessConfig.version` from pinned commit `41739cf` back to `main`
- keep the `rlm-harness` -> `nano-rlm` rename from #2373

## Why

#2373 changed the default-install behavior (tracking `main` -> frozen commit) alongside the rename. Per the Slack discussion, the rename is fine but the behavior change shouldn't ride along in that PR — the default should stay latest `main`, and anyone who needs a reproducible eval can pin an explicit `version` override.

Note: #2376 builds on this and additionally locks the resolved commit at run start.

### DIFF
--- a/verifiers/v1/harnesses/rlm/harness.py
+++ b/verifiers/v1/harnesses/rlm/harness.py
@@ -20,7 +20,7 @@ logger = logging.getLogger(__name__)
 
 BuiltinSkill = Literal["edit", "search"]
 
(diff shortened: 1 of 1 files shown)


hallerite (dismissed):

samsja:
Superseded by #2376, which carries the full stack (revert + rename + run-start lock) as one PR.

mikasenghaas (approved):

PR #2384, opened 2026-08-17
feat: preserve harness protocol metadata

hallerite:
## Summary

- forward ACP initialize, capability, new-session, prompt, and close response metadata through the framed runner protocol
- retain the ordered lifecycle as public `HarnessArtifact` values that are available through scoring but excluded from trace serialization
- preserve repeated prompt metadata instead of overwriting it, append prompt artifacts under the session lock, and cap cumulative metadata retained per session
- let ACP harnesses require namespaced agent capabilities and `session/close` support before `session/new` sends private metadata
- make runner shutdown cancellation-resilient so its process stack cannot be abandoned mid-close

## Validation

- `uv run pytest tests/v1/test_acp_metadata.py tests/v1/test_trace.py -q`
- complete deterministic v1 suite passed as part of the stack
- Ruff, format, and Ty passed locally and in CI


hallerite:
Superseded by #2386, which now contains the full RLM ACP contract as one PR against main.

PR #2385, opened 2026-08-17
feat: preserve RLM call lineage across retries

hallerite:
## Summary

- add a generic, harness-bound logical-call resolver for stable identity, private request headers, and namespaced `ModelCall.metadata`
- replay completed non-streaming calls and coalesce concurrent attempts by opaque logical ID while preserving the existing Stainless retry fallback
- bind every logical ID to route, request bytes, and stable metadata so changed bodies or lineage cannot receive a stale replay
- move every `X-RLM-*` constant, parser, schema rule, and stripping decision into the RLM harness protocol module
- keep `Trace.nodes` authoritative for training branches; runtime lineage remains descriptive call metadata

Unrelated harnesses cannot activate RLM behavior merely by sending `X-RLM-*` headers.

Depends on #2384.

## Validation

- `uv run pytest tests/v1/test_rlm_lineage.py tests/v1/test_trace.py -q`
- complete deterministic v1 suite passed as part of the stack
- Ruff, format, and Ty passed locally and in CI

### DIFF
--- a/tests/v1/test_rlm_lineage.py
+++ b/tests/v1/test_rlm_lineage.py
@@ -8,17 +8,18 @@ import verifiers.v1 as vf
 from verifiers.v1.clients import ModelContext
 from verifiers.v1.configs.client import EvalClientConfig
 from verifiers.v1.errors import ProviderError
-from verifiers.v1.interception.server import (
+from verifiers.v1.harnesses.rlm.protocol import (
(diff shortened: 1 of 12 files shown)


hallerite:
Superseded by #2386, which now contains the full RLM ACP contract as one PR against main.

PR #2390, opened 2026-08-17
feat!: remove hub auto-install of plugin packages

mikasenghaas:
## Summary

- Remove auto-install of environments from the Environments Hub. Plugin resolution (`_import_plugin`) now imports the package directly and never installs anything — verifiers assumes every plugin package (taskset, harness, judge, env) is already installed.
- Delete `verifiers/v1/utils/install.py` and `verifiers/v1/utils/install_utils.py` (`ensure_installed`, `install_from_hub`, `install_from_local`, `install_from_repo`, `parse_env_id`, `is_hub_env`, `env_name`, `env_module`, ...).
- Make `ID` a plain `str`. No `org/name[@version]` parsing or validation; an id is the name of an installed package.
- Remove the `skip_plugin_install` contextvar. It existed only to suppress parse-time hub installs; with no installs it has no purpose, and nothing sets it.
- The `name` properties on `TasksetConfig` / `HarnessConfig` now return the id as-is (no org/version stripping).
- Update config docstrings, error messages, and the `evaluate-environments` skill to drop `org/name` id forms. The skill's commands also switch from `prime eval run` to the v1 entrypoints (`uv run eval` / `uv run validate`) — `prime eval` is not yet v1-native — and its resume/output section now documents the real flow (`uv run eval @ <run-dir>/configs/eval.json --resume` and the `configs/` + `logs/` run-dir layout).

- Update `test_reward_name_fallback` to use a plain id instead of `org/my-judge@1.0.0` (the reward-name fallback is now `name` > `id` > snake-cased class name).

The legacy (v0) tree, its `vf-install` entrypoint, and the prime-cli command modules are untouched. Pushing eval results to the hub (`utils/platform.py`) is also untouched — only pulling is removed.

## Breaking

- Hub ids (`org/name`, `org/name@version`) no longer resolve. A plugin id must be the name of an installed package; install hub packages yourself (e.g. `prime env install`) before referencing them.
- Removed from `verifiers.v1.utils`: the `install` and `install_utils` modules and everything they exported.
- Removed `verifiers.v1.utils.loaders.skip_plugin_install`.


samsja:
>So a missing explicit plugin load (via load_taskset / load_harness / load_judge, or an explicit env.id) fails loudly
 with the message above — but a missing taskset used only to probe for a bundled harness/env silently falls back.
 
 is this behavior true ? if yes seems not good that we fallback to another harness if the one we want to use is not installed

mikasenghaas:
Update: reverted the probe change in a2764bf — keeping the previous default behavior (the probes catch `ModuleNotFoundError` too). The analysis above still holds: explicitly specified plugins never fall back, and a missing taskset errors loudly at config narrowing before the probes ever run, so the arm is unreachable in a real run either way.

samsja (approved):

PR #2404, opened 2026-08-19
fix: enable Qwen3-Max prompt caching for hosted evals

stu-cao:
## Why this change is needed

Hosted Evals launches `prime eval run`, whose default inference path builds a non-streaming Chat Completions request in `OpenAIChatCompletionsClient`. Alibaba's `qwen/qwen3-max` endpoint does not create an explicit prompt cache from provider pinning alone: the request must contain block-level `cache_control: {"type": "ephemeral"}` breakpoints. See the [OpenRouter Alibaba Qwen caching contract](https://openrouter.ai/docs/guides/best-practices/prompt-caching#alibaba-qwen).

The inference gateway already preserves caller-authored markers. It should not infer stable prompt boundaries for heterogeneous clients. This client is the narrowest layer where the resolved model, destination API, final structured messages, and ordered tool definitions coexist immediately before the request is sent.

## What changes

For the exact `qwen/qwen3-max` model on the exact `openrouter.ai` or `api.pinference.ai` hosts, the default Chat Completions client now adds explicit cache breakpoints to:

- the final cacheable text block in the leading system/developer prefix;
- the final function tool, preserving tool order.

The transformation is copy-on-write. It does not mutate the rollout prompt or tool definitions, and it never marks the changing user/assistant rollout tail. This limits Alibaba's 1.25x cache-write premium to content expected to be reused across examples, parallel rollouts, and tool turns.

The generated marker is exactly `{"type": "ephemeral"}`. No `ttl` is sent because Alibaba supports a fixed five-minute cache, not Anthropic's optional one-hour retention.

Caller intent always wins: if `cache_control` or `prompt_cache_breakpoint` is already present at the request, message, direct content-block, or tool level, this client leaves the entire request unchanged. Tool parameter schemas containing a property named `cache_control` are not mistaken for cache directives.

## Scope and related changes

- [PrimeIntellect-ai/inference-iac#157](https://github.com/PrimeIntellect-ai/inference-iac/pull/157) controls provider locality and pins Qwen3-Max to Alibaba.
- [PrimeIntellect-ai/platform#4450](https://github.com/PrimeIntellect-ai/platform/pull/4450) locks the gateway passthrough contract for caller-authored Qwen markers.
- [PrimeIntellect-ai/platform#4447](https://github.com/PrimeIntellect-ai/platform/pull/4447) was closed because gateway-side breakpoint inference and `_pi_prompt_cache` route metadata put prompt policy at the wrong layer.

This PR intentionally covers the default Hosted Evals Chat Completions path. Explicit Responses, Anthropic Messages, renderer, native v1 proxy, and direct-SDK harness paths are separate request producers and are unchanged.

## Validation

- focused client tests: 22 passed;
- offline test suite: 926 passed, 75 API/e2e tests deselected;
- Ruff: passed;
- Python 3.13 type check: passed;
- all pre-commit hooks relevant to the changed Python files passed. The all-files run still reports 321 pre-existing MD033 inline-HTML findings in the untouched legacy SWE README.

No paid inference APIs were called.

## Rollout

After merge, publish a `verifiers` release and ensure Hosted Evals resolves that release. The hosted runner currently installs an upgraded version satisfying the environment's `verifiers` requirement; exact older pins are intentionally preserved and will not receive this behavior. Post-deploy, verify two back-to-back requests with an identical marked static prefix and require the second response to report nonzero cached prompt tokens.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 93a9b8a239375f86353248a245c9cf4a59ba7994. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add automatic prompt cache markers for `qwen/qwen3-max` on hosted eval gateways
> - When the model is `qwen/qwen3-max` and the API base URL is `openrouter.ai` or `api.pinference.ai`, `OpenAIChatCompletionsClient.get_native_response` now auto-inserts `cache_control={'type': 'ephemeral'}` on the last text segment of the final leading system/developer message and on the last function tool.
> - Auto-insertion is skipped if the caller has already supplied any `cache_control` or `prompt_cache_breakpoint` markers in the prompt, tools, or `extra_body`.
> - New utility helpers in [openai_chat_completions_client.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2404/files#diff-687c2124e93bc3b7a864e3fa533d29c4ff552e54e00ffb55f7056fbb13d0ae11) handle detection and mutation of cache markers without mutating caller-owned objects.
> - Behavioral Change: requests for `qwen/qwen3-max` on the listed hosts will now include prompt cache markers by default unless the caller opts out by providing their own markers.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 93a9b8a.</sup>
> 
>

### DIFF
--- a/tests/test_client_multimodal_types.py
+++ b/tests/test_client_multimodal_types.py
@@ -1,8 +1,13 @@
+from copy import deepcopy
 from types import SimpleNamespace
+from typing import Any, cast
 
 import pytest
 
 from verifiers.clients.openai_chat_completions_client import OpenAIChatCompletionsClient
+from verifiers.legacy.clients import (
+    openai_chat_completions_client as openai_chat_completions_module,
+)
 from verifiers.types import (
     AssistantMessage,
     ImageUrlContentPart,
@@ -22,6 +27,46 @@ from verifiers.utils.response_utils import parse_response_message
 class _OpenAIMessage(SimpleNamespace):
     def model_dump(self):
         return self.__dict__
+
+
+async def _capture_chat_request_body(
+    monkeypatch: pytest.MonkeyPatch,
+    *,
+    base_url: str,
+    model: str,
+    prompt: Any,
+    tools: Any = None,
+    sampling_args: Any = None,
+) -> dict[str, Any]:
+    captured_body: dict[str, Any] = {}
+    sentinel = cast(Any, object())
+
+    async def fake_post_chat_completion(
+        _client: object,
+        path: str,
+        *,
+        body: dict[str, Any],
+        extra_headers: Any = None,
+    ) -> Any:
+        assert path == "/chat/completions"
+        assert extra_headers is None
+        captured_body.update(body)
+        return sentinel
+
+    monkeypatch.setattr(
+        openai_chat_completions_module,
+        "post_chat_completion_with_routed_experts_sidecar",
+        fake_post_chat_completion,
+    )
+    client = OpenAIChatCompletionsClient(SimpleNamespace(base_url=base_url))
+    response = await client.get_native_response(
+        prompt=prompt,
+        model=model,
+        sampling_args=sampling_args or {},
+        tools=tools,
+    )
+    assert response is sentinel
+    return captured_body
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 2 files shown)


stu-cao:
close it as this is not the right approach. We handle this in server side: PR: https://github.com/PrimeIntellect-ai/platform/pull/4450

PR #2405, opened 2026-08-20
Increase default connect timeout to 30 seconds

rasdani:
## What changed

- Increase the default HTTP client connection timeout from 5 seconds to 30 seconds.

## Why

A 5-second connection timeout is too aggressive when connection establishment is temporarily slow. The longer default allows transient connection delays to recover without changing the existing overall request timeout.

## Impact

HTTP clients now wait up to 30 seconds to establish a connection by default. Explicitly configured connection timeouts continue to override the default.

## Validation

- Ruff and formatting checks passed.
- Type checks passed.
- Verified the runtime default resolves to 30 seconds.

### DIFF
--- a/verifiers/v1/clients/base.py
+++ b/verifiers/v1/clients/base.py
@@ -9,7 +9,7 @@ from verifiers.v1.configs.client import BaseClientConfig, resolve_api_key
 
 # No read timeout: agentic completions are slow and the rollout timeout is the real
 # backstop. The connect bound stays so an unreachable endpoint still fails fast.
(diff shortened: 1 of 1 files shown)


tianjianl:
A 1,024-concurrency PRIME-RL production run reproduced this failure severely (57–60% final episode errors, all `ProviderError: Request timed out.`). I opened #2632 as a superseding draft with the timeout raised to 60s plus the connection-pool/keepalive and proxy hardening required by that run.

PR #2362, opened 2026-08-14
Support Harbor runtime environment and health checks

xeophon:
## Overview

Adds Harbor environment variables and health checks to V1 runtimes.

## Details

- exposes live-only task runtime environment values across subprocess, Docker, Modal, and Prime
- resolves Harbor environment templates at rollout time while keeping resolved values out of traces
- runs Harbor-compatible health checks after environment staging and before harness setup
- applies the same behavior to separate verifier runtimes

---

> [!NOTE]
> **Medium Risk**
> Touches every runtime exec path and changes concurrency for borrowed sandboxes; env resolution affects secrets and process behavior but follows existing Harbor template semantics.
> 
> **Overview**
> Harbor-style **environment variables** and **healthchecks** now flow into live sandboxes without landing in traces. Tasks expose a `runtime_env()` hook (default empty); Harbor resolves `[environment.env]` templates at execution time and applies them on every backend.
> 
> **Runtime behavior:** each box carries mutable `runtime.env`, merged with per-command env via `process_env()` (process values win). Docker/Modal/Prime pass env at container/sandbox creation and on exec; subprocess merges into the host-filtered environment. `provision_runtime` accepts an optional `env` dict before `start()`.
> 
> **Rollouts:** owned runtimes set `runtime.env` from the task at open; borrowed boxes take `borrow_lock`, swap in the borrower's env, and restore the owner's env on release/abort so shared sandboxes stay task-isolated.
> 
> **Harbor tasks:** `HarborData` / verifier configs gain `env` and `healthcheck`; setup can poll a healthcheck command after staging `environment/`; separate verifier boxes get their own env/healthcheck and skip agent environment upload when not a fresh copy. Verifier env templating is generalized as `resolve_env()`.
> 
> **Tooling:** debug, validate, and agent `provision()` set `runtime.env` consistently (empty when provisioning without a task).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f6b4d09ea70c19fc3d83d77183cb0fd801f14204. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add per-task runtime environment variables and healthcheck support to Harbor tasks
[...]

PR #2418, opened 2026-08-21
fix(v1): allow concurrent borrowed runtime environments

parkerpettit:
## Summary

- give each borrowed rollout its own environment-scoped runtime view
- allow multiple agents to share one sandbox concurrently
- preserve sandbox-level network policy validation
- preserve owner lifecycle checks against the physical runtime

## Why

#2362 added a rollout-wide lock because borrowed rollouts were applying their process environment directly to the shared runtime. Without serialization, concurrent agents could overwrite each other’s environment variables.

The lock prevented that race, but it also serialized the entire lifetime of every borrowed rollout. This can deadlock nested-agent workflows:

1. Agent 1 opens and retains the sandbox.
2. Agent 1 acquires the borrow lock.
3. Agent 2 is launched in the same sandbox.
4. Agent 1 needs Agent 2’s response to finish.
5. Agent 2 cannot start until Agent 1 releases the lock.
6. Neither agent can make progress.

This PR removes the need for that lock by using `Runtime.with_env()`. Each rollout gets its own process-environment view while continuing to share the same underlying sandbox handles. Concurrent rollouts therefore cannot overwrite each other’s environment, which preserves the isolation the lock was added to provide without serializing agents.

Network policy remains sandbox-scoped. The existing borrowed-placement validation rejects any agent or task whose required policy is incompatible with the sandbox. All valid borrowers therefore share the same network policy and can run concurrently.

Owner lifecycle checks continue to operate on the physical runtime, so the sandbox cannot be closed or replaced while borrowed agents are using it.


tests/v1/test_borrowed_runtime_env.py
mikasenghaas, line 1: remove

verifiers/v1/rollout.py
```diff
@@ -80,8 +80,8 @@ def __init__(
         self._shared_tools = shared_tools or {}
         self._interception = interception
         self.runtime = runtime
+        self._borrowed_runtime = runtime
         self._owns_runtime = runtime is None
```
mikasenghaas: these seem like inverses?

parkerpettit: yeah they were redundant, good catch. got rid of self._owns_runtime

mikasenghaas (approved):

PR #2438, opened 2026-08-25
feat(v1): type ACP prompt results

hallerite:
## Summary

- carry each completed ACP prompt as a typed `ACPTurnResult`: visible reply, stop reason, prompt-response `_meta`, and update metadata
- treat `session/prompt` as the completion boundary and keep `session/close` as best-effort resource cleanup
- validate required ACP extension markers by presence from capability or initialize metadata
- preserve the harness-visible `primary_reply` independently of descendant branches
- unify explicit `Idempotency-Key` and marked SDK retry replay through one bounded non-streaming cache

## Why

This is the shared ACP substrate for nano-RLM and Prime Agent. Agent-specific metadata remains opaque to the runner and is consumed only after a successful prompt response. A close/transport problem cannot invalidate an otherwise completed and scoreable trajectory.

This is PR 1 of the three-PR Verifiers stack. The nano-RLM and Prime Agent integrations are stacked above it.

## Validation

- `uv run ruff check --fix .`
- `uv run pytest tests/` — 919 passed, 77 credential-gated skips
- focused idempotency and primary-reply regressions — 8 passed
- changed-file pre-commit hooks and push-hook CI-parity type check

The repository-wide markdown hook also reports an existing inline-HTML violation in `verifiers/legacy/envs/experimental/composable/tasksets/swe/README.md`; this stack does not touch that file.

---

> [!NOTE]
> **Medium Risk**
> Changes ACP wire format, trace semantics (`last_reply`), and model-interception retry/idempotency behavior—areas that affect scoring and graph atomicity under retries.
> 
> **Overview**
> Introduces **`ACPTurn`** as the structured result of each ACP prompt (visible reply, stop reason, response/prompt metadata, and per-update metadata). The ACP runner and host wire protocol now return a **`result`** object instead of a bare **`reply`**, with **`ACPHarness.acp_turn_result`** for harness-specific consumption after a successful prompt.
> 
> Adds **`Trace.root_reply`** (serialized on **`WireTrace`**) so scoring and segments can treat the **root agent’s answer** separately from the last assistant node when the message graph has branches. **`Segment.last_reply`** and **`Trace.last_reply`** prefer **`root_reply`**; each rollout segment clears it at start and **restores the previous value** if the segment commits no model turns (failed/partial segments).
> 
> The interception server replaces the single “last request” replay slot with a **bounded idempotency cache**: explicit **`Idempotency-Key`** (non-streaming only) and **marked SDK retries** (body-digest keys) coalesce in-flight attempts and replay the same HTTP response without double-committing graph turns.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ba9c80a19e2b74d9067db78ac0e5b147962e9a1e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


verifiers/v1/agent.py
```diff
@@ -210,7 +214,10 @@ async def _turn(self, message: str | Messages | None) -> Segment:
                     saw_assistant = True
                 elif saw_assistant and isinstance(node.message, ToolMessage):
                     segment_messages.append(node.message)
-            return Segment(messages=segment_messages)
+            return Segment(
+                messages=segment_messages,
+                primary_reply=self.trace.primary_reply,
+            )
```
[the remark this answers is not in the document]
hallerite: Confirmed and fixed in 24d45f0e6. Rollout.step now clears the prior primary reply at each segment boundary and restores it only when the attempted segment commits no model turns. A regression test covers a later segment that samples a current reply and then fails; Segment.last_reply and Trace.last_reply both select the current partial output. Focused tests and the push-hook CI-parity checks pass.

tests/v1/test_idempotency.py
mikasenghaas, line 1: u sure we need this? can we not cover this w/ e2e tests?

verifiers/v1/interception/server.py
```diff
@@ -111,6 +115,87 @@ def _completion_response(completion: dict | None) -> web.Response:
...
+    return web.Response(
+        body=response.body,
+        status=response.status,
+        content_type="application/json",
+        charset="utf-8",
+    )
+
+
+@dataclass(frozen=True)
+class _IdempotentAttempt:
```
mikasenghaas, line 141: hmm can you tell me again what this enables compare to main?

verifiers/v1/trace.py
```diff
@@ -406,6 +406,8 @@ class Trace(BaseModel, Generic[DataT, StateT, AgentConfigT]):
     """Unweighted, named metrics; `None` as in `rewards`."""
     info: dict[str, Any] = Field(default_factory=dict)
     """Scratch space for task-specific metadata."""
+    primary_reply: str | None = None
```
mikasenghaas: this seems weird what is it?

tests/v1/test_idempotency.py
hallerite: some things we can't test e2e, because it requires specific agent behavior

verifiers/v1/trace.py
hallerite: "The graph knows all branches, but does not identify which final assistant message belongs to the root agent. ACP’s prompt result does. Without storing that selection, scoring or user-sim can accidentally consume a subagent’s reply."


maybe better to rename this as root_reply?

verifiers/v1/interception/server.py
hallerite: with nano-rlm and prime-agent (and technicallly codex with subagents too), there could be multiple subagents making progress concurrently, so there might be multiple requests in parallel. if request A is committed but its response is lost, request B may complete before A is retried. Main only remembers the latest completed request, so A’s retry is no longer recognized and may be sampled and committed twice. that's the problem we solve here

verifiers/v1/acp/__init__.py
```diff
@@ -31,6 +33,17 @@
 JsonObject: TypeAlias = dict[str, JsonValue]
 
 
+class ACPTurnResult(BaseModel):
```
mikasenghaas: what abt just `ACPTurn`?

verifiers/v1/interception/server.py
mikasenghaas: nice, makes sense

verifiers/v1/trace.py
mikasenghaas: ya i like root more

hallerite: Implemented in signed commit `ba9c80a19`: `Trace.primary_reply` and `Segment.primary_reply` are now `root_reply`, including rollout boundary handling, ACP assignment, serialization, and tests.

verifiers/v1/acp/__init__.py
hallerite: Agreed. Renamed the public and runner turn types to `ACPTurn` in signed commit `ba9c80a19`, and updated both stacked harness consumers.

mikasenghaas (approved):

PR #2449, opened 2026-08-26
feat: add optional ACP semantic edges

hallerite:
## Summary

Adds an optional, versioned ACP extension for typed semantic relationships between model requests while leaving Verifiers' physical training branches unchanged.

- accept cumulative `SemanticEdgeSet` payloads under `ai.prime.acp/semantic-edges-v1`
- record harness-owned correlation IDs as `ModelCall.acp.request_id`
- resolve request-ID edges incrementally into typed `MessageNode.semantic_parents`
- preserve arbitrary validated edge labels so new harness relationships do not require a Verifiers schema change
- pin nano-RLM's merged semantic-edge implementation from [#153](https://github.com/PrimeIntellect-ai/nano-rlm/pull/153) at [`4a63696`](https://github.com/PrimeIntellect-ai/nano-rlm/commit/4a6369611c06d3943ac40681f374a464feb706b9)

This remains one message-node DAG with physical and semantic parent links, not a second lineage graph. The unrelated concurrent physical-graph race is handled separately in [#2477](https://github.com/PrimeIntellect-ai/verifiers/pull/2477).

## Data model

| Concept | Meaning |
| --- | --- |
| `MessageNode.parent` | Exact physical-prefix parent used for training branches. |
| `MessageNode.semantic_parents` | Ordered typed links to other committed message nodes. Multiple semantic parents are valid. |
| `ParentLink` | `{node, type}`, where `node` is an index into `Trace.nodes`. |
| `Branch` | A root-to-leaf physical-parent path, not an agent/session identity. |
| `ACPInfo.request_id` | Harness-owned request identifier used to resolve ACP edges to committed Verifiers nodes. |

For example, a parent request consuming two child results is represented directly on its message node:

```python
MessageNode(
    parent=physical_parent,
    semantic_parents=[
        ParentLink(node=5, type="subagent_return"),
        ParentLink(node=8, type="subagent_return"),
    ],
    ...,
)
```

nano-RLM initially uses `continuation`, `subagent_call`, `subagent_return`, and `compaction`. These are conventions rather than a closed enum; other validated labels and namespaced relationships are preserved.

## Transport

1. The harness assigns a request ID and includes `X-ACP-Model-Request-ID` on the corresponding model request.
2. Interception stores the ID as `ModelCall.acp.request_id` and removes both the private correlation header and its locally owned `Idempotency-Key` before provider forwarding.
3. The harness publishes its cumulative request-ID edge set in ACP response metadata.
4. `Trace.add_semantic_edges` resolves committed source and target requests, validates the additions atomically, and appends `ParentLink`s to the target nodes.
5. Replayed cumulative payloads are idempotent, and `WireTrace` persists the resolved links with the rollout.

Harnesses that omit the header and extension metadata behave exactly as before: `ModelCall.acp=None` and `semantic_parents=[]`.

## Validation

- full deterministic Verifiers suite passes locally; credential-gated cases skip when credentials are absent
- focused trace/interception and cumulative-replay coverage: 11 passed
- Ruff, formatting, Ty, commit hooks, and push hooks pass
- CI is green on Python 3.11, 3.12, and 3.13, including the live V1 E2E

---

> [!NOTE]
> **Medium Risk**
> Touches interception (headers, idempotency/streaming), trace serialization, and ACP shutdown; behavior is backward-compatible when harnesses omit the extension, but edge resolution errors or mis-correlated request IDs could fail rollouts.
> 
> **Overview**
> Adds an **optional ACP extension** so harnesses can declare typed relationships between logical model requests without changing how **physical training branches** are built from `MessageNode.parent`.
> 
> Harnesses can send **`X-ACP-Model-Request-ID`** on intercepted calls; the server stores it on **`ModelCall.acp`** and strips private ACP headers before the provider. Cumulative edge payloads under **`ai.prime.acp/semantic-edges-v1`** are applied on ACP turns and on **`session/close`** via **`Trace.add_semantic_edges`**, which maps request IDs to sampled nodes and appends **`MessageNode.semantic_parents`** (`ParentLink` with extensible edge types). Validation rejects duplicates, self-edges, and cycles; SDK retries sharing one request ID resolve to the last committed node.
> 
> **RLM** is bumped to a nano-rlm commit that emits these edges, with **version-scoped install caches** and session metrics also consumable from close metadata. E2E checks for RLM now assert every call has ACP info and semantic parent links point at sampled nodes. **`aiohttp`** is pinned to **>=3.14.1**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a0dc9c8f0b2c4179eb01fecf84154e2ed4be3035. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


verifiers/v1/interception/server.py
```diff
@@ -492,7 +495,10 @@ async def handle_request(
         del raw
         body = dialect.apply_overrides(body, session.ctx.model, session.ctx.sampling)
         streaming = dialect.streaming(body)
-        upstream_headers = dict(request.headers)
+        try:
+            lineage, upstream_headers = extract_call_lineage(request.headers)
```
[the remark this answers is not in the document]
hallerite: Fixed in [`6c54c5c`](https://github.com/PrimeIntellect-ai/verifiers/commit/6c54c5cbf134b9b64e8c3880320e6e525e4cf26a). A streaming request with validated ACP lineage now uses `Idempotency-Key` only to bind the logical request ID, strips both the key and private lineage headers before provider forwarding, and creates no replay/coalescing state. Streaming requests with an idempotency key but no lineage remain rejected. Verified with a focused handler regression and the full suite: 919 passed, 76 credential-gated skipped.

verifiers/v1/interception/server.py
```diff
@@ -509,14 +515,18 @@ async def handle_request(
...
                     ),
                     status=400,
                 )
-            replay_key = f"explicit:{idempotency_key}"
+            if not streaming:
+                replay_key = f"explicit:{idempotency_key}"
+            # Streaming lineage uses the key only to bind its logical request id;
+            # streaming replay/coalescing remains unsupported. Never forward this
+            # local transport identity (or the private lineage envelope) upstream.
             upstream_headers = {
```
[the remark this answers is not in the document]
hallerite: Fixed in [`6dca9d3`](https://github.com/PrimeIntellect-ai/verifiers/commit/6dca9d3d9f80aff1d3008bdf65e02b8b1d1fa51c). `Idempotency-Key` is now stripped only when a validated lineage envelope makes it local request identity. Ordinary non-lineage non-streaming requests preserve the provider-facing key; local replay/coalescing is unchanged, and ordinary streaming requests with a key remain rejected. The focused handler regression covers both forwarding paths, and the full suite passes: 919 passed, 76 credential-gated skipped.

verifiers/v1/harnesses/rlm/harness.py
```diff
@@ -36,7 +36,7 @@ class _SessionSnapshot(BaseModel):
 
 class RLMHarnessConfig(HarnessConfig):
     version: str = Field(
-        default="d4ce3e10e63b359f4f3d432d58a77471e9e21fe7", min_length=1
+        default="48e2a761d4d5b525d90783e006cc76fd24bd11b8", min_length=1
```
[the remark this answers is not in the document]
hallerite: Fixed in b2a4e4896. The install root is now keyed by a SHA-256 of RLMHarnessConfig.version, and each version owns its checkout, lock, readiness marker, and binary. prepare_acp resolves the binary from the same version-keyed root, so an executable cached for an older nano-RLM ref cannot satisfy setup for a new ref. I also validated branch-like refs, generated-shell syntax, and command/path agreement; the full verifier suite passes (919 passed, 76 credential-gated skipped).

verifiers/v1/trace.py
```diff
@@ -171,6 +172,8 @@ class ModelCall(BaseModel):
     """The failure that ended this call, coupled to the exchange that caused it."""
     policy: PolicyEvent | None = None
     """Policy mediation applied to the request before this call."""
+    lineage_request_id: str | None = Field(default=None, pattern=LINEAGE_ID_PATTERN)
```
mikasenghaas: `lineage_id`?

verifiers/v1/trace.py
```diff
@@ -196,10 +199,49 @@ class Branch(BaseModel):
...
+    @property
+    def compaction_ids(self) -> tuple[str, ...]:
+        """Compactions correlated with calls on this physical message path."""
+        return tuple(
+            dict.fromkeys(
+                request.compaction_id
+                for request in self._requests_in_call_order()
+                if request.compaction_id is not None
+            )
+        )
```
mikasenghaas: what do these mean?

mikasenghaas: e.g. wouldn't i expect all msgs on a brnch to ahve the same compaction ids?

hallerite: That intuition exposed the problem with this helper. In the RLM rollout, calls on a physical branch normally share one context epoch, but request-level compaction_id can mean either the compaction that created the current context or the new summary attempt being performed. A branch between two compactions could therefore aggregate both IDs. I removed the property; compaction is now read where it belongs, as the semantic source-context to target-context transition in the manifest.

hallerite: Good catch. These were unused convenience views that mixed semantic runtime provenance into the physical training Branch API. I removed the private lineage index, all three aggregate properties, calls_by_session, and branches_by_session in a1a05a9dd. Branch is now purely the root-to-leaf message/token path; consumers can join ModelCall.lineage_request_id to Trace.lineage when they need semantic execution data.

verifiers/v1/trace.py
hallerite: lineage_request_id is intentional: it is a foreign key to one specific record in LineageManifest.requests. lineage_id would suggest an ID for the whole manifest or execution graph, but there is no such entity. Provider attempts/retries can share this logical request ID.

verifiers/v1/semantic.py
```diff
@@ -0,0 +1,129 @@
...
+            )
+            if identity in identities:
+                raise ValueError(f"duplicate semantic edge: {identity!r}")
+            identities.add(identity)
+            children.setdefault(edge.source_request_id, []).append(
+                edge.target_request_id
+            )
+            nodes.update((edge.source_request_id, edge.target_request_id))
+
+        visiting: set[str] = set()
```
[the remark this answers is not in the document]
hallerite: Fixed in [`ea11820`](https://github.com/PrimeIntellect-ai/verifiers/commit/ea1182081f82c34251f2021f41a56edd453ab78d). `SemanticEdgeManifest` now uses iterative Kahn traversal, so validation is independent of Python recursion depth while still rejecting cycles. Added a focused 2,000-edge acyclic-chain regression; the full suite passes (921 passed, 76 credential-gated skipped).

verifiers/v1/trace.py
```diff
@@ -493,6 +514,65 @@ def branches(self) -> list[Branch]:
...
+
+        resolved: list[SemanticEdge] = []
+        for edge in manifest.edges:
+            endpoints: list[int] = []
+            for request_id in (
+                edge.source_request_id,
+                edge.target_request_id,
+            ):
+                nodes = nodes_by_request.get(request_id, set())
+                if len(nodes) != 1:
```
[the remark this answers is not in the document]
hallerite: Fixed in [`ea11820`](https://github.com/PrimeIntellect-ai/verifiers/commit/ea1182081f82c34251f2021f41a56edd453ab78d). Reconciliation now walks calls in append order and resolves a logical request ID to its last committed sampled node. Failed attempts still have no node, and a focused regression covers two committed streaming attempts under one ID. Full suite: 921 passed, 76 credential-gated skipped.

verifiers/v1/trace.py
```diff
@@ -493,6 +514,65 @@ def branches(self) -> list[Branch]:
...
+            node = stack.pop()
+            visited += 1
+            for child in children[node]:
+                indegree[child] -= 1
+                if indegree[child] == 0:
+                    stack.append(child)
+        if visited != len(self.nodes):
+            raise ValueError("semantic edges create a cycle in the message graph")
+
+        self.semantic_edges = resolved
```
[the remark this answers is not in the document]
hallerite: Fixed in [`ea11820`](https://github.com/PrimeIntellect-ai/verifiers/commit/ea1182081f82c34251f2021f41a56edd453ab78d). A logical request now resolves deterministically to the last committed sampled node, matching sequential SDK retry semantics; failed attempts are ignored. Added the multiple-committed-retry regression and reran the full suite (921 passed, 76 credential-gated skipped).

verifiers/v1/harnesses/rlm/harness.py
```diff
@@ -36,7 +36,7 @@ class _SessionSnapshot(BaseModel):
 
 class RLMHarnessConfig(HarnessConfig):
     version: str = Field(
-        default="d4ce3e10e63b359f4f3d432d58a77471e9e21fe7", min_length=1
+        default="a1d7856df609d539389b6069544f373a68c7f727", min_length=1
```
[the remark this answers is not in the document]
hallerite: False positive: Git object IDs are not chronological. The previously pinned `a1d7856` is a signed direct child of `e010abf`, so it includes the continuation-edge implementation; the new signed pin `e4ad6c5` is in turn a direct child of `a1d7856`. The default therefore contains continuation edges, the failed-subagent return fix, and the final naming cleanup. The PR description now links the exact final commit.

verifiers/v1/trace.py
```diff
@@ -171,6 +176,8 @@ class ModelCall(BaseModel):
     """The failure that ended this call, coupled to the exchange that caused it."""
     policy: PolicyEvent | None = None
     """Policy mediation applied to the request before this call."""
+    model_request_id: str | None = Field(default=None, pattern=MODEL_REQUEST_ID_PATTERN)
```
mikasenghaas: don't need to say opaque here imo. and should make sure that the producer is ACP (not the actual http model request). myb even in the name say `acp_request_id`? or more general `ACPInfo.request_id`?

verifiers/v1/trace.py
```diff
@@ -188,7 +195,14 @@ def min_new_input_tokens(calls: Iterable[ModelCall]) -> Iterator[tuple[ModelCall
 
 
 class Branch(BaseModel):
-    """A root-to-leaf graph path; each branch becomes one training sample."""
+    """A root-to-leaf message-graph path; each branch becomes one training sample.
+
+    A branch often matches one harness context window, but the concepts are not
```
mikasenghaas: remove narration

verifiers/v1/trace.py
```diff
@@ -387,7 +401,11 @@ class Trace(BaseModel, Generic[DataT, StateT, AgentConfigT]):
     """The tools advertised to the agent, automatically recorded from last intercepted turn."""
 
     nodes: list[MessageNode] = Field(default_factory=list)
-    """The message graph; branches are derived views and storage stays linear in turns."""
+    """The message graph, including physical and semantic parent links.
+
+    Branches are derived solely from exact-prefix ``MessageNode.parent`` links; semantic
```
mikasenghaas: would also remove this comment

verifiers/v1/trace.py
```diff
@@ -493,6 +511,70 @@ def branches(self) -> list[Branch]:
             )
         return branches
 
+    def reconcile_semantic_edges(self, edge_set: SemanticEdgeSet) -> None:
```
mikasenghaas: when is this called? this looks like a one-time op, any way we cna build this on-the-fly like the physical links

p soon i want to build "streaming episodes/traces" in which case i would have to rebuild this graph at every turn to accurate semantic edges i would guess? would like to avoid this

verifiers/v1/semantic.py
```diff
@@ -0,0 +1,129 @@
...
+EDGE_TYPE_PATTERN = r"^[A-Za-z][A-Za-z0-9._:-]{0,127}$"
+
+ACP_SEMANTIC_EDGES_METADATA_KEY = "ai.prime.acp/semantic-edges-v1"
+ACP_MODEL_REQUEST_ID_HEADER = "X-ACP-Model-Request-ID"
+
+ACP_EXTENSION_HEADERS = frozenset({ACP_MODEL_REQUEST_ID_HEADER.lower()})
+"""Private ACP extension headers consumed at interception and never sent upstream."""
+
+
+class _StrictSemanticModel(BaseModel):
```
mikasenghaas: nit: imo can just duplicate the model config into the classes and remove this

verifiers/v1/semantic.py
```diff
@@ -0,0 +1,129 @@
...
+
+from __future__ import annotations
+
+import re
+from collections.abc import Mapping
+
+from pydantic import BaseModel, ConfigDict, Field, model_validator
+
+MODEL_REQUEST_ID_PATTERN = r"^[A-Za-z0-9._:-]{1,128}$"
+EDGE_TYPE_PATTERN = r"^[A-Za-z][A-Za-z0-9._:-]{0,127}$"
```
mikasenghaas: myb small comment/example what this looks like?

verifiers/v1/graph.py
```diff
@@ -68,6 +72,12 @@ class MessageNode(BaseModel):
 
     parent: int | None = None
     """Index into `Trace.nodes` of the predecessor message; None for a root."""
+    semantic_parents: list[ParentLink] = Field(default_factory=list)
```
mikasenghaas, line 75: why list?

mikasenghaas (commented):
nice this looks quite clean indeed. i think one thing we need to make super clear in the api itself is that the difference between
- vf-owned message node linking (rn we have a list of nodes and use indices to point ot parents and link model calls to nodes)
- harness-owned (via acp) linking, rn via `request_id`
it shoud be ultra clear which ids/indices come from vf and which are advertised by the harness

verifiers/v1/trace.py
hallerite: Done in ce548237. ModelCall now stores acp: ACPInfo | None, with the harness-owned identifier at call.acp.request_id. The docs no longer say opaque and explicitly distinguish ACP request IDs from Verifiers node indices.

verifiers/v1/semantic.py
hallerite: Done in ce548237. EDGE_TYPE_PATTERN now includes examples such as subagent_return and vendor:review.

verifiers/v1/graph.py
hallerite: Kept a list and documented why in ce548237: a node may have multiple parents of the same relationship type, application appends incrementally, and serialized traces retain stable advertised order. Duplicate (node, type) links are prevented during application; a set would require hashable models and lose wire order, while a dict would need an artificial key.

verifiers/v1/trace.py
hallerite: Done in ce548237. The Branch docstring is back to the one-line definition.

verifiers/v1/semantic.py
hallerite: Done in ce548237. Removed _StrictSemanticModel and placed the strict model_config directly on ACPInfo, SemanticEdge, ParentLink, and SemanticEdgeSet.

verifiers/v1/trace.py
hallerite: Done in ce548237. The Trace.nodes comment is reduced to the field definition.

verifiers/v1/trace.py
hallerite: Done in ce548237. This is called after each ACP turn and again on close. It is now Trace.add_semantic_edges: cumulative replays skip existing links, new links are cycle-checked against the current physical plus semantic parents, and the batch is appended only after validation. Existing parent lists are never rebuilt. Tests cover list identity across a cumulative update and atomic cycle rejection.

verifiers/v1/graph.py
```diff
@@ -575,6 +624,29 @@ def _commit_turn(turn: PendingTurn, response: Response) -> int:
...
+    # time. Exact token-span equality preserves intentional renderer-level forks; tokenless relay
+    # turns use message identity. Because commit is synchronous, once this loop stops the rest of
+    # the tail can be appended without another request interleaving.
+    prefix = list(prefix)
+    while len(prefix) < len(prompt):
+        i = len(prefix)
+        span = spans[i] if spans and i < len(spans) else None
+        end = span[1] if span else path_len
+        node_tokens = prompt_ids[path_len:end]
+        existing = _matching_node(
```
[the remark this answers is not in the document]
hallerite: Legit. Fixed in 8f77bfac8: commit-time reconciliation now matches an unspanned prompt message against the existing node’s exact token prefix and advances by that node’s stored physical token length. Added a regression where a concurrently committed sampled assistant has message_spans[i] = None; the graph stays linear and token attribution remains [[1], [2, 3], [4], [5]]. Focused graph + semantic trace suite passes.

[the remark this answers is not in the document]
hallerite: Good catch. I fixed the unspanned-assistant case, then moved the entire generic concurrent-commit fix out of this semantic-edge PR. The signed fix and regression now live in standalone PR #2477: https://github.com/PrimeIntellect-ai/verifiers/pull/2477

verifiers/v1/interception/server.py
```diff
@@ -508,19 +514,15 @@ async def handle_request(
...
                     status=400,
                 )
-            replay_key = f"explicit:{idempotency_key}"
-            upstream_headers = {
-                name: value
-                for name, value in upstream_headers.items()
-                if name.lower() != IDEMPOTENCY_KEY_HEADER.lower()
-            }
+            if not streaming:
+                replay_key = f"explicit:{idempotency_key}"
```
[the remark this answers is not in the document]
hallerite: Valid finding. Fixed in signed commit 6b0b33e5d: after ACP metadata is extracted, interception now removes the locally owned `Idempotency-Key` from provider headers for both non-streaming replay calls and ACP-correlated streams. The key still participates in local replay/coalescing, but it can no longer reach an upstream request whose body has been rewritten.

mikasenghaas (approved):
nice!

PR #2450, opened 2026-08-26
fix: parse last XML answer in message completions

### LINKED ISSUE #2447: XMLParser.parse_answer is inconsistent between string and message completions
The docstring of `parse_answer` states it extracts the *last* answer from a completion. The two branches disagree:

- **String branch:** calls `self.parse(completion, last=True)` → returns the **last** matching block.
- **Message branch:** calls `self.parse(content)` without `last=True` → returns the **first** matching block.

Iterating messages with `reversed()` selects which *message* to read, but not which *block* within that message.

**Reproducer:**

```python
import verifiers as vf

parser = vf.XMLParser(["reversed_text"], answer_field="reversed_text")
text = "<reversed_text>FIRST</reversed_text> filler <reversed_text>LAST</reversed_text>"

print("as string:  ", parser.parse_answer(text))
print("as messages:", parser.parse_answer([{"role": "assistant", "content": text}]))
```

**Output:**

```
as string:   LAST
as messages: FIRST
```

**Environment:** verifiers 0.3.1, Python 3.13.0, Windows.

**Impact:** the same completion text is parsed differently depending on completion type, so scores are not comparable between single-turn and multi-turn evaluation modes. `XMLParser` is the default parser, so this is not scoped to a single environment. The failure is silent — no exception is raised.

**Suggested fix:** pass `last=True` in the message branch, or document explicitly that behaviour differs by completion type.

zakazaka95:
## Summary

- make `XMLParser.parse_answer` consistently select the last matching answer block for string and message completions
- add regression coverage for multiple answer tags within one assistant message

Fixes #2447

## Testing

- `pytest tests/test_xml_parser.py` — 19 passed
- Ruff check and format check passed for both changed files

> [!NOTE]
> ### Fix `XMLParser.parse_answer` to use last `<answer>` tag in message completions
> In `parse_answer`, assistant message content is now parsed with `last=True`, so the last matching answer tag is returned instead of the first. Adds a test confirming `45` is extracted from a message with multiple `<answer>` tags.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized fcd1cdd.</sup>
> 
>

### DIFF
--- a/tests/test_xml_parser.py
+++ b/tests/test_xml_parser.py
@@ -101,6 +101,17 @@ class TestXMLParser:
             result == "45"
         )  # Should return just the last answer, not a full parse() namespace
 
+    def test_parse_answer_from_message_completion_uses_last_answer(self, xml_parser):
+        """Test extracting the last answer from a message completion."""
+        completion = [
+            {
+                "role": "assistant",
+                "content": "<answer>44</answer><reasoning>Actually, that's not right either.</reasoning><answer>45</answer>",
+            }
+        ]
+        result = xml_parser.parse_answer(completion)
+        assert result == "45"
+
     def test_parse_answer_no_answer_field(self, xml_parser):
         """Test parse_answer when no answer field is found."""
         completion = [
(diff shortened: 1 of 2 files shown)


xeophon:
this only affected v0/legacy, which we don't support anymore. closing here

PR #2512, opened 2026-09-02
feat: stream served episodes turn by turn

mikasenghaas:
## Summary

- Served episodes stream turn by turn over the env-serve wire (`verifiers/v1/serve/delta.py`). `Trace.notify` fires at every phase change and after every recorded turn; the worker's `DeltaStreamer` diffs the run's live traces against what it already sent and ships the new part (header once, appended nodes/calls/errors, semantic links, changed scalars), so every byte crosses the wire once. A cursor advances only after its delta is on the wire. `EnvClient.run` assembles the episode from the deltas (`EpisodeAssembly`, which copies the nodes it keeps so a later link never mutates a delta the caller holds), validates it once, and takes `on_delta` so a caller can relay or persist the stream; the client's receive loop survives a malformed frame.
- `pending` preview: the interception server hands each model request's uncommitted tail (tool results, user turns) to `Trace.preview`, so a watcher sees a tool result before the model has answered; the committed turn replaces it and the finished record never carries it.
- The eval CLI is `uv run vf-eval` and runs in-process only. Serving an env to many consumers is prime-rl's job (its `uv run eval` runs env servers and consumes this stream), so the CLI keeps the quick local path and never assembles deltas; its TUI footer and console log point at prime-rl's `eval`. The env-server e2e fixture drives a worker pool through `EnvClient` directly.

Based on `main`; the eval CLI stays for a transition period (PrimeIntellect-ai/verifiers#2610 removes it later). prime-rl PrimeIntellect-ai/prime-rl#3471 consumes the stream for its live traces.

## Breaking

- Env-serve wire: replies are 4 frames (`client_id, request_id, kind, data`) with `kind` = `delta` | `reply`; `RunResponse` carries the episode head and per-trace counts instead of the traces. A client and server must be on the same side of this change.
- Every console script is prefixed: `vf-eval`, `vf-validate`, `vf-debug`, `vf-replay`, `vf-init`, `vf-gepa`. Generic names collide with the scripts of a workspace that installs verifiers as a dependency (prime-rl's `eval`), and the winner is whichever package was installed last. `vf-eval` has no `[serve]` / `--serve.*` / `--no-serve`: rollouts always run in-process; use prime-rl's `uv run eval` for env-server runs. Existing venvs: `uv sync --reinstall-package verifiers --reinstall-package prime-rl` regenerates the scripts.
- `Env.run_slot` takes `on_trace`; `Trace` gains `watch`/`notify`/`preview`/`clear_preview`.

## Verification

- `tests/v1/test_serve_delta.py`: turns, a link onto an earlier node, a retried attempt and the final reward round-trip through `DeltaStreamer` → `EpisodeAssembly` → `WireEpisode`; a refused send is diffed again; the pending preview streams, is replaced by the committed turn and never reaches the record; assembling never mutates a delta; the field lists cover every serialized `Trace` field.
- `uv run pytest tests/v1 -m "not e2e"` green apart from the config-parse cases of tasksets not installed in the venv; ruff clean.
- End to end through prime-rl: gsm8k single turn, terminal-bench-2 fix-git with bash and with the rlm harness delegating to a sub-agent (semantic links `subagent_call`/`subagent_return`), kuhn-poker two-seat episodes, 5-step reverse-text RL runs with token ids and logprobs, including router replay plus sampling replay on a MoE; no assembly mismatch.

> [!NOTE]
> ### Stream served episodes turn by turn with delta frames
> - Adds `DeltaStreamer` in [delta.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2512/files#diff-106a565d15f6f5c86f7357854d512768b3f8a4775026a0eae32c69b828f5da31) that watches trace changes during `run_slot` and sends incremental deltas; the client `EpisodeAssembly` reconstructs a `WireEpisode` from those deltas plus a final episode head and `TraceSummary` counts.
> - Changes the wire protocol to three-frame messages (request ID, kind, data). Delta frames are relayed by the pool and client without closing the request; reply frames close it. Updated in [server.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2512/files#diff-64422f6a8749fcc024415b2756129bc98000e830e74b3b5245de50a555480a22), [pool.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2512/files#diff-0c9c59578214be2580d340bc524321e352f48038207a68945637d00fc399600c), and [client.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2512/files#diff-ef98a21052536967b8f8723e36022f69e539f4e8b13794791e8254434ac4696e).
> - Adds a trace watcher/preview API in [trace.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2512/files#diff-834820c3aa80d87ee1c1c57fb54c74f09c66ac727036d964e6005c8613c00b5e). `Env.run_slot` gains an `on_trace` callback; `Rollout.open`/`close` and `InterceptionServer` notify the trace on phase changes and pending turn previews.
> - `PendingTurn.commit`/`commit_prompt` now clear the trace's pending preview after committing.
> - Risk: `RunResponse` in [types.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2512/files#diff-8e290c0dfc583c92b20ee9f20fa2ecd0df82459ad63f8e819e87aa856de97cf4) removes the serialized `WireEpisode` field and replaces it with an episode head and `TraceSummary` list — any out-of-tree consumer expecting the full episode in the run response will break.
>
> 
> #### Changes since #2512 opened
>
> - Renamed console script entry points in `pyproject.toml` and updated all documentation and usage strings to use `vf-` prefixed command names [02a6eff]
> - Refactored trace preview system from unkeyed to keyed architecture [55500a3]
> - Updated all trace preview and clear operations to use keyed API with PendingTurn instances as keys [55500a3]
> - Added abandon mechanism for cleaning up previews when turns fail or are cancelled [55500a3]
> - Changed `verifiers.v1.trace.Trace` to store preview entries by `id()` of the key object rather than by the key object itself [73d602d]
> - Updated `PRIME_RL_HINT` constant in the `hint.py` module [fc7e45b]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized fe771d8.</sup>
> 
>

---

> [!NOTE]
> **High Risk**
> Breaking env-serve wire protocol and `RunResponse` shape require matching client/server versions; CLI renames and removal of eval’s serve path change how users run evaluations.
> 
> **Overview**
> This PR **streams env-server episodes incrementally** instead of returning one fat `RunResponse`. Workers use **`DeltaStreamer`** to diff live traces on `Trace.notify()` and send **`delta`** frames; clients **`EpisodeAssembly`** rebuilds traces and validate once against a final **head + `TraceSummary` counts**. The wire format is now **`[request_id, kind, data]`** with `kind` = `delta` | `reply`; the pool relays deltas without closing the request.
> 
> **`Trace`** gains **`watch` / `notify` / keyed `preview` / `clear_preview`**; rollouts and the interception server fire notifications (including **pending tool/user tails** before commit). **`Env.run_slot`** adds **`on_trace`** so the server can subscribe traces to the streamer.
> 
> **`vf-eval` is renamed and narrowed**: console scripts are **`vf-*`** to avoid clashing with prime-rl’s **`eval`**, docs/config comments follow suit, and **`vf-eval` always runs in-process**—**`[serve]` / `--no-serve` are removed** from eval config—with dashboard/log hints to use **prime-rl** for env-server scale and live dashboards.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fc7e45b4da088043c408685b784fd40471f2d020. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hallerite (commented):
The streaming approach looks useful. I found one mutation issue at the boundary between the client assembly and consumers that buffer deltas, described inline.

verifiers/v1/serve/delta.py
```diff
@@ -0,0 +1,274 @@
...
+            trace = self.traces[trace_id] = {
+                **delta["open"],
+                **{field: [] for field in LIST_FIELDS},
+                "pending": [],
+            }
+        for index, links in (delta.get("links") or {}).items():
+            trace["nodes"][int(index)]["semantic_parents"].extend(links)
+        for field in LIST_FIELDS:
+            if field in delta:
+                trace[field].extend(delta[field])
```
hallerite, line 246: The assembly retains references to the node dictionaries in the incoming delta. When a later `links` delta extends a node's `semantic_parents`, it also mutates that earlier delta, which `EnvClient.run()` has already handed to `on_delta`. This matters for prime-rl's buffered dashboard writer: if both deltas arrive before the next flush, the earlier node already contains the link when serialized, and replaying the subsequent link update adds it again. I reproduced one link in the client assembly becoming two in the delayed monitor replay. Could we ensure assembly updates cannot mutate deltas already handed to consumers, for example by copying the mutable node/link containers at this boundary?

verifiers/v1/cli/eval/hint.py
```diff
@@ -0,0 +1,5 @@
+PRIME_RL_HINT = (
+    "prime-rl's `uv run eval` streams traces live into a dashboard, evaluates several "
```
mikasenghaas: make this a recommendation to move over and notice that this entrypoint will be deprecated

mikasenghaas: Done in fc7e45b4: "vf-eval will be deprecated - move to prime-rl's `uv run eval`: it streams traces live into a dashboard, evaluates several envs in one run and scales through env servers." Shown in the TUI footer and the console log.

samsja (approved):

pyproject.toml
```diff
@@ -152,12 +152,12 @@ prime-pydantic-config = false
 renderers = false
 
 [project.scripts]
-eval = "verifiers.v1.cli.eval.main:main"
-validate = "verifiers.v1.cli.validate:main"
-debug = "verifiers.v1.cli.debug:main"
-replay = "verifiers.v1.cli.replay:main"
-init = "verifiers.v1.cli.init:main"
-gepa = "verifiers.v1.cli.gepa:main"
+vf-eval = "verifiers.v1.cli.eval.main:main"
```
hallerite, line 155: The hosted-eval breakage has moved into this PR with the executable rename. I checked the platform's current source: `backend/app/packages/evals/constants.py` still allows `verifiers>=0.3.2`, and `script_generator.py` installs that dependency and runs `env eval @ "$CONFIG_FILE"`. A fresh hosted environment picking up this change will therefore have `vf-eval` but no `eval` executable, even though #2610 is on hold. Could we update the hosted runner to invoke `vf-eval` with a compatible version requirement, or constrain its verifiers dependency before releasing this? Keeping the implementation for the transition period is useful, but its current caller needs to move with the command rename.

hallerite (commented):
Thanks for fixing the delta aliasing issue; the node/link copying addresses the earlier finding, and the 99 non-live verifiers tests passed locally. One remaining compatibility issue: holding the CLI-removal PR does not by itself preserve hosted evals, because this PR renames the executable they invoke.

pyproject.toml
mikasenghaas: i want hosted evals to use the prl entrypoint, will do the update pr after this lands

hallerite (approved):
lgtm

PR #2565, opened 2026-09-08
fix(lean-taskset): audit proof axioms to close the injected-axiom bypass

### LINKED ISSUE #2177: Lean taskset: injected `axiom` bypasses the reward (#1271 `#print axioms` follow-up)
The `lean_compiled` reward accepts a proof when the protected signature is intact,
`lake env lean` exits 0, and there's no `declaration uses 'sorry'`. It never checks
the proof's axioms, so a rollout can prove any goal by injecting one:

```lean
import Mathlib
axiom cheat : False
-- DO NOT MODIFY the theorem statement below.
theorem fake_irrational : Irrational (1 : ℝ) := by
  exact cheat.elim
```

`Irrational (1 : ℝ)` is false. The `axiom` sits above the protected signature, so
the guard passes and the reward returns 1.0.

#1271 flagged this exact gap as future work; #1480 closed a sibling bypass in the
same reward.

## Reproduction

Taskset image (Lean 4.27.0, Mathlib), exact `lake env lean` command:

| proof body | compile | reward |
| --- | --- | --- |
| `axiom cheat : False` + `exact cheat.elim` | exit 0, silent | 1.0 |
| `sorry` | `declaration uses 'sorry'` | 0.0 |
| axiom removed | unknown identifier | 0.0 |

`#print axioms fake_irrational` → `[cheat, propext, Classical.choice, Quot.sound]`.

## Fix

`#print axioms <name>` after compile, reject axioms outside the trusted set. Ready
on a branch with a regression test. Two calls for you first:

- **native_decide** rests on `Lean.ofReduceBool` / `Lean.trustCompiler` — allow it
  (my default) or reject it as #1271 lists separately?
- The probe output needs last-match parsing like #1480's `EXIT_CODE`, else a proof
  spoofs a clean line with `#eval`.

Can open the PR whenever it's useful.

CaptainAni187:
Closes #2177

`lean_compiled` awards 1.0 when the protected signature is intact and `lake env lean` exits 0 with no `sorry`. It never checks what the proof depends on, so a rollout can close any goal by supplying its own axiom above the signature:

```lean
import Mathlib
axiom cheat : False
-- DO NOT MODIFY the theorem statement below.
theorem fake_irrational : Irrational (1 : ℝ) := by
  exact cheat.elim
```

`Irrational (1 : ℝ)` is false. The `axiom` sits above the protected region, so the signature guard passes, the compile is silent, and the reward returns 1.0.

#1271 flagged this exact gap as future work ("or axiomatize the goal"); #1480 closed a sibling bypass in the same reward.

## Change

After a clean compile, write the rollout's file to a sibling path with `#print axioms <name>` appended, compile that, and reject any axiom outside Mathlib's trusted base — `propext`, `Classical.choice`, `Quot.sound`. The rollout's own file is left untouched.

The parsing and trust decision are pure functions in `scoring.py`, alongside the existing helpers.

### One detail worth flagging

`parse_axioms_output` matches the **last** `depends on axioms` line, not the first. Our query is appended at the end of the file, so anything earlier came from the rollout — without this, a model could inject

```lean
#eval IO.println "'fake_irrational' depends on axioms: [propext, Classical.choice, Quot.sound]"
```

above its proof and have the audit read the decoy. That is the same failure #1480 fixed for `EXIT_CODE` markers, so I followed the reasoning already documented in `parse_compile_output`. There is a test for it.

`declaration_name` returns `""` for an anonymous `example`, which skips the audit rather than failing the rollout — the pre-existing behaviour for statements the reward cannot pin.

## Tests

`tests/v1/test_lean_scoring.py`, 9 tests:

- the pure helpers: name extraction, axiom-list parsing, the trusted baseline, the injected `cheat` axiom, and the decoy-shadowing case
- the reward itself, driven through a stub runtime: a compiling proof resting on `cheat` scores **0.0** with `lean_untrusted_axioms == ["cheat"]`, and a genuine proof still scores **1.0**

Reverting only the `taskset.py` wiring (keeping the helpers) turns the first reward test into `assert 1.0 == 0.0`, so the test pins the bypass itself rather than the helpers.

`tests/v1` (excluding `test_e2e.py`): 76 passed, up from 67, with the same 11 pre-existing failures in `test_configs.py::test_eval_config_parses` — missing optional taskset packages in my environment, unrelated to this change. `test_color_codeword_tasks.py` does not collect locally for the same reason.

`ruff check` and `ruff format` clean.

I could not run the real Lean toolchain locally, so the audit is verified against recorded `#print axioms` transcripts rather than a live `lake env lean`. The transcripts match the ones in the issue, including `#print axioms fake_irrational → [cheat, propext, Classical.choice, Quot.sound]`. Happy to adjust if the real output differs in shape.

> [!NOTE]
> ### Reject Lean proofs resting on injected axioms in `lean_compiled`
> - Adds `LeanTask._print_axioms` which writes an audit copy of the proof with a declaration-specific `#axioms` query, runs Lean on it, and returns the combined transcript without touching the original proof file.
> - Adds parsers in [scoring.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2565/files#diff-6df81277181f7dcb921afcca61e1eb6a8630d07219d543c6a6784e0ae7fea305): `declaration_name` extracts the theorem/lemma name from a signature, `parse_axioms_output` selects the last reported axiom list, and `untrusted_axioms` filters names against the `TRUSTED_AXIOMS` Mathlib baseline.
> - `lean_compiled` in [taskset.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2565/files#diff-4e7d754229e5803fddc0fea71640b92984c26d8a2da8fb28aa8c2c86150fa25a) now runs the audit after a successful compile and sets score to zero with tampering info when any untrusted axiom is found.
> - Behavioral Change: a proof that compiles but depends on an axiom outside the three accepted Mathlib baseline names now scores zero and is marked tampered; if no declaration name is found or the audit yields no untrusted axioms, full credit is retained.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 148dfce.</sup>
> 
>

### DIFF
--- a/tests/v1/test_lean_scoring.py
+++ b/tests/v1/test_lean_scoring.py
@@ -1,0 +1,134 @@
+"""Tests for the Lean taskset's scoring helpers and axiom audit."""
+
+from types import SimpleNamespace
+
+from verifiers.v1.tasksets.lean.scoring import (
+    TRUSTED_AXIOMS,
+    declaration_name,
+    parse_axioms_output,
+    untrusted_axioms,
+)
+
+CHEAT_OUTPUT = (
+    "'fake_irrational' depends on axioms: "
+    "[cheat, propext, Classical.choice, Quot.sound]"
+)
+CLEAN_OUTPUT = (
+    "'real_theorem' depends on axioms: [propext, Classical.choice, Quot.sound]"
+)
+
+
+def test_declaration_name_reads_theorem_and_lemma() -> None:
+    assert (
+        declaration_name("theorem fake_irrational : Irrational (1 : ℝ) := by")
+        == "fake_irrational"
+    )
+    assert declaration_name("lemma my_lemma (x : ℕ) : x = x := by") == "my_lemma"
+    assert declaration_name("import Mathlib\ntheorem foo : True := by") == "foo"
+
+
+def test_declaration_name_returns_empty_when_unparsable() -> None:
+    assert declaration_name("example : True := by") == ""
+    assert declaration_name("") == ""
+
+
+def test_parse_axioms_output_reads_the_axiom_list() -> None:
+    assert parse_axioms_output(CHEAT_OUTPUT) == [
+        "cheat",
+        "propext",
+        "Classical.choice",
+        "Quot.sound",
+    ]
+
+
+def test_parse_axioms_output_handles_no_axioms() -> None:
+    assert parse_axioms_output("'t' does not depend on any axioms") == []
+    assert parse_axioms_output("") == []
+
+
+def test_untrusted_axioms_flags_an_injected_axiom() -> None:
+    """Regression test for the ``axiom cheat : False`` reward bypass."""
+    assert untrusted_axioms(CHEAT_OUTPUT) == ["cheat"]
+
+
+def test_untrusted_axioms_accepts_the_mathlib_baseline() -> None:
+    assert untrusted_axioms(CLEAN_OUTPUT) == []
+    assert set(TRUSTED_AXIOMS) == {"propext", "Classical.choice", "Quot.sound"}
+
+
+def test_untrusted_axioms_matches_the_last_report() -> None:
+    """A printed decoy must not shadow the real audit result.
+
+    Our ``#print axioms`` query is appended last, so anything earlier in the
+    transcript came from the rollout -- the same reasoning ``parse_compile_output``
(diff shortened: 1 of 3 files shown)


CaptainAni187:
Closing this for now to tidy up my open PRs. Happy to reopen if it's still useful.

Issue #2570, opened 2026-09-09
Served runs rebuild every task as the taskset's declared `Task` type, silently dropping per-case subclasses


willccbb:
## What happens

A taskset whose `load()` builds some cases as a `Task` **subclass** (to give those cases their own `@vf.reward` methods, `toolsets`, or hooks) scores differently on the two supported run paths:

- `uv run eval <taskset>` (served, the default): every case is rebuilt inside the env-server worker as the taskset's single declared `Task` type. Rewards, toolsets and hooks that live only on the subclass are gone. A base class with no rewards scores exactly `0.000` for every rollout.
- `uv run eval <taskset> --no-serve`: the subclass runs and scores as authored. (In-process keeps the class; it does not fix the tunnel-placement gap below, which reads the base class on both paths.)

Nothing reports the difference. `validate` never scores; the served run just returns the wrong numbers.

## Mechanism

`serve/server.py`: `EnvServer.__init__` fixes `self.task_cls = type(self.env.taskset).task_type()` once, and `_build_task` does `self.task_cls(self.data_cls.model_validate(task_data), config.taskset.task)`. `serve/types.py:RunRequest` carries `task_data`, `client`, `model`, `sampling`; nothing names the class the client built.

A second instance of the same assumption, in-process as well as served: `Env._requires_tunnel` (`env.py`) sizes tunnels off `task_type().toolsets(...)`, so a subclass-declared toolset is invisible to placement on remote runtimes.

## Repro (model-free, main `27bbd216`)

```python
import verifiers.v1 as vf

class CaseData(vf.TaskData):
    answer: str

class ExtraToolset(vf.Toolset[vf.ToolsetConfig]):
    TOOL_PREFIX = "extra"
    @vf.tool
    def ping(self) -> str:
        """Reply pong."""
        return "pong"

class BaseTask(vf.Task[CaseData]):
    @vf.reward
    async def base_match(self, trace: vf.Trace) -> float:
        return float(trace.last_reply == self.data.answer)

class SpecialTask(BaseTask):
    @classmethod
    def toolsets(cls, config):
        return [ExtraToolset(vf.ToolsetConfig())]
    @vf.reward
    async def special_match(self, trace: vf.Trace) -> float:
        return 1.0

class SubclassTaskset(vf.Taskset[BaseTask, vf.TasksetConfig]):
    def load(self):
        return [BaseTask(CaseData(idx=0, prompt="say a", answer="a"), self.config.task),
[...]

PR #2572, opened 2026-09-09
fix(v1): refuse a served run whose taskset loads per-case Task subclasses

willccbb:
Fixes the silent half of #2570.

A served run (the default) rebuilds each task inside the env-server worker as the taskset's declared `Task` type from its wire data. A taskset whose `load()` builds some cases as a `Task` subclass loses that subclass's rewards, toolsets and hooks on the served path and scores them silently wrong, while `--no-serve` scores them as authored.

This change makes the runner refuse before dispatch when any loaded task's type is not the taskset's `task_type()`, naming the classes and the two ways out: dispatch on a `TaskData` field inside the declared class, or run in-process with `--no-serve` (which keeps the class; remote tunnel placement still sizes off the base class, see #2570). Nothing is written or spawned before the check. Whether per-case subclasses should become a supported authoring pattern (carrying the class over the wire) is left to #2570.

- `verifiers/v1/cli/eval/runner.py`: the refusal, after task selection.
- `tests/v1/fixtures/subclass_task_v1.py`: a two-case taskset whose second case is a subclass with its own reward and toolset.
- `tests/v1/test_e2e.py`: a deterministic test (no model, no `e2e` mark) asserting the `ValueError` and that no run directory is created.

Validation, model-free: the new test passes; `uv run pytest tests/ -m "not e2e"` 81 passed; ruff check, ruff format and the pre-commit hooks pass; the in-process loader still yields both classes for the fixture. Reproduction of the underlying bug against `EnvServer._build_task` is in #2570.

---

> [!NOTE]
> **Low Risk**
> Adds a pre-dispatch validation on the served eval path only; in-process runs are unchanged and the failure mode is an explicit error instead of silent mis-scoring.
> 
> **Overview**
> **Served eval runs** now fail fast if the taskset’s `load()` returns tasks whose runtime type is not the taskset’s declared `Task` class. On the default served path, workers rebuild tasks from wire data using only that declared type, so per-case subclasses would lose extra rewards, toolsets, and hooks and score incorrectly without an obvious error.
> 
> The check runs in `run_eval` after task selection and **before** any output directory or worker pool is created. The raised `ValueError` names the unexpected class(es) and points authors to branch on `TaskData` inside the declared task type or use **`--no-serve`** for in-process runs that preserve the subclass instances.
> 
> A **`subclass-task-v1`** fixture (base task + `SpecialTask` with its own reward and toolset) and a model-free test assert the refusal message and that no run directory is written.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a1514f66eb5f4f29aa1e43562715d454962c55b5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Refuse served runs when taskset loads per-case `Task` subclasses in `run_eval`
> - Adds validation in `run_eval` ([runner.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2572/files#diff-65be4a576330571f2729a3bbcdbe92ac81bbcba450c313597c1bdca3a1c3f2e4)) that checks whether any selected task's exact type differs from the taskset's declared `Task` type, raising `ValueError` before output setup or dispatch for served runs.
> - In-process runs skip this check and retain their existing path; the error message recommends using a `TaskData` field or `--no-serve`.
> - Adds a `subclass-task-v1` fixture ([subclass_task_v1.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2572/files#diff-a216994fa899026c8fdcab07790294f03e3be66358751d8d2a9fc4dd131c6f41)) and an end-to-end test ([test_e2e.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2572/files#diff-ebbbd919487889487f891b359a756e6c4630e106b16ddaef8ccddc6b900a96e1)) that confirms the `ValueError` mentions `SpecialTask` and `--no-serve` and that no served output directory is created.
> - Risk: served evaluations that previously relied on per-case task subclasses will now fail with `ValueError` instead of proceeding; verify tasksets in `verifiers/v1` for `SubclassTaskset`-style loaders.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a1514f6.</sup>
> 
>

### DIFF
--- a/tests/v1/fixtures/subclass_task_v1.py
+++ b/tests/v1/fixtures/subclass_task_v1.py
@@ -1,0 +1,44 @@
+"""A taskset whose second case is a per-case `Task` subclass carrying its own reward and
+toolset — the shape a served run cannot carry over the wire."""
+
+import verifiers.v1 as vf
+
+
+class CaseData(vf.TaskData):
+    answer: str
+
+
+class ExtraToolset(vf.Toolset[vf.ToolsetConfig]):
+    TOOL_PREFIX = "extra"
+
+    @vf.tool
+    def ping(self) -> str:
+        """Reply pong."""
+        return "pong"
+
+
+class BaseTask(vf.Task[CaseData]):
+    @vf.reward
+    async def base_match(self, trace: vf.Trace) -> float:
+        return float(trace.last_reply == self.data.answer)
+
+
+class SpecialTask(BaseTask):
+    @classmethod
+    def toolsets(cls, config: vf.TaskConfig) -> list[vf.Toolset]:
+        return [ExtraToolset(vf.ToolsetConfig())]
+
+    @vf.reward
+    async def special_match(self, trace: vf.Trace) -> float:
+        return 1.0
+
+
+class SubclassTaskset(vf.Taskset[BaseTask, vf.TasksetConfig]):
+    def load(self) -> list[BaseTask]:
+        return [
+            BaseTask(CaseData(idx=0, prompt="say a", answer="a"), self.config.task),
+            SpecialTask(CaseData(idx=1, prompt="say b", answer="b"), self.config.task),
+        ]
+
+
+__all__ = ["SubclassTaskset"]
--- a/tests/v1/test_e2e.py
+++ b/tests/v1/test_e2e.py
@@ -10,6 +10,9 @@ import sys
 
 import pytest
 
+from verifiers.v1.cli.eval.runner import run_eval
+from verifiers.v1.configs.cli.eval import EvalConfig
+
 mark = pytest.mark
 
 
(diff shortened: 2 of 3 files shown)


willccbb:
On the risk note in the summary: no bundled taskset trips the refusal. Every module under `environments/` and `verifiers/v1/tasksets/` was imported at `a1514f66` and its `Task` classes checked for subclass chains; 60 of 65 modules import in isolation and none defines a `Task` that subclasses another `Task` in the same module. The five that do not import alone are two harness/verify scripts (`compact/program.py`, `gsm8k/verify.py`) and the nemo_gym taskset and weather example (package-relative imports), and by inspection they define one `Task` class each (`NeMoGymTask`). The only taskset that builds a subclass per case is the new test fixture.

PR #2575, opened 2026-09-10
fix(runtimes): harness setup reuses an installed uv

faresobeid:
## What

`_ENSURE_UV` (runtimes/base.py; also used by `mcp/launch.py`) now runs `command -v uv` first and skips both `pip install -q -U --user uv` and the astral installer when `uv` is already on PATH. The PATH export still comes first, so a `~/.local/bin/uv` from an earlier rollout on the same box counts. Images without `uv` take the same install path as before.

## Why

In a pipeline opening ~90 concurrent rollouts on prime sandboxes, about 3% of opens failed in harness setup because `_ENSURE_UV` downloads `uv` on every `prepare_uv_script` (pip on PyPI, else astral.sh) over flaky egress — even on images that ship `uv`. A `uv` that is already present needs no network at all.

Behavior change: `uv` is no longer upgraded to the latest release when one is present. `prepare_uv_script` needs a `uv` that knows `uv sync --script` / `uv python find --script`; an image with a very old pre-installed `uv` would now surface uv's own argument error instead of being silently upgraded.

## Validation

- `tests/v1/test_runtimes.py` runs the real snippet under `sh` with a stub `pip` (leaves a marker, drops a `uv` into `~/.local/bin`) and an optional stub `uv`: with `uv` present, `pip` does not run; without it, `pip` runs and `uv` is then found. Also asserts `command -v uv` precedes `pip install` in the composed command.
- e2e `test_single_turn[null-harness-in-subprocess]` and `test_agentic[bash-harness-in-subprocess]` pass (they run `prepare_uv_script` on a host with `uv` pre-installed).
- `uv run pytest tests/v1 -m "not e2e"`, `ruff`, `pre-commit` pass.

> [!NOTE]
> ### Fix `_ENSURE_UV` to reuse an installed `uv` on PATH
> - Changes the shell sequence in [base.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2575/files#diff-0811c040a163eb5a3212208f1e3a6c6388d7690fcc4780c081cc7754036232b9) to check whether `uv` is already discoverable before the pip-install branch; the curl/wget fallback still follows.
> - Adds test helpers in [test_runtimes.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2575/files#diff-c8621bddd8207b6fd39881c4208d82a11882b2a5126487fbdbe6274aabc94422) that stub `pip` and `uv` on a temporary PATH, plus a test asserting existing `uv` is preserved and missing `uv` triggers installation.
> - Behavioral Change: runtime setup now skips installation when `uv` is already on PATH, instead of always running the pip branch.
>
> 
> #### Changes since #2575 opened
>
> - Removed the `tests.v1.test_runtimes` test module [cdcb7ee]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b94df19.</sup>
> 
>

---

> [!NOTE]
> **Low Risk**
> Localized change to sandbox bootstrap shell logic; reduces egress-dependent failures but drops automatic uv upgrades when one is already on PATH.
> 
> **Overview**
> **Harness setup** no longer re-downloads or reinstalls `uv` on every `prepare_uv_script` run when a usable binary is already discoverable.
> 
> The `_ENSURE_UV` shell snippet in `base.py` (also wired into MCP launch setup) now runs `command -v uv` **before** the `pip install --user uv` and Astral curl/wget fallback. PATH is still prefixed with `~/.local/bin` first, so an `uv` left from a prior rollout counts. Images without `uv` follow the same install chain as before.
> 
> **Behavior change:** an existing `uv` is **not** upgraded to the latest release; only missing `uv` triggers install. Very old pre-installed `uv` may fail with uv’s own errors instead of being silently refreshed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cdcb7eed0c22887cad5f5a5a6d53a9587de60616. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/v1/test_runtimes.py
```diff
@@ -0,0 +1,41 @@
+"""Runtime shell snippets, run under a real `sh` with stubbed tools: the harness setup keeps a
```
xeophon: can you remove the test? its a pointless one

hallerite: agreed

faresobeid:
Removed the shell-snippet test as requested; the change is now the `_ENSURE_UV` snippet only (`command -v uv` before any install).

hallerite (approved):

PR #2579, opened 2026-09-10
Support framework aware network policies in the Modal runtime

hrdkbhatnagar:
Tasks declaring `network_allow=[]` currently fail runtime resolution on Modal as a backend for sandboxes. 

Implement `NetworkPolicyConfig` and execution time enforcement so trusted setup can install dependencies, then the agent can reach its framework endpoints while other outbound traffic is restricted. Support TLS domain allowlists on port 443, restore setup access when reusing a sandbox, and reject unsupported rules or failed policy updates before launching the agent.

This uses Modal's experimental runtime policy update API and native SNI filtering. It inherits the provider's domain-fronting limitations. Existing unrestricted and `network_access=false` configurations retain their behavior. The existing agent e2e matrix includes a restricted Modal case

Validation: 80 deterministic tests passed; Ruff and pre-commit passed. Live Modal checks verified allowed/blocked traffic and setup reuse. A scripted model endpoint completed the restricted agent/tool/scoring test with proper reward. DeepSWE task loading and one Prime oracle/grading check also passed against the local checkout. 

> [!NOTE]
> ### Add framework-aware outbound network policies to `ModalRuntime`
> - Restricted Modal sandboxes are created with broad domain and CIDR egress grants for setup, then narrowed at execution time to a domain-only allowlist derived from framework routes and configured user rules
> - Adds a domain validation helper that normalizes egress hostnames, recognizes IP addresses, rejects credentials and unsupported schemes, and enforces HTTPS-only framework rules without wildcards
> - `ModalConfig` now extends `NetworkPolicyConfig` and validates restricted configurations: requires `network_access`, rejects the wildcard allow rule, and validates each allow rule through the new helper
> - `ModalRuntime.start` checks for Modal's `_experimental_set_outbound_network_policy` SDK method and raises `SandboxError` when it is missing; test harness widened to accept dictionary runtime params and a Modal test case with an empty allowlist is added
> - Risk: restricted runtimes now fail validation when `network_access` is disabled, when the wildcard allow rule is used, or when an allow rule is not a supported Modal DNS rule; `prepare_execution` in [modal.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2579/files#diff-a6217d92f043d6050e8609d4341d62b5d097d5db796f6197cd604e09d81683c0) applies the narrowed policy via a 60-second RPC and wraps failures in `SandboxError`
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c4f6c01.</sup>
> 
>

### DIFF
--- a/verifiers/v1/runtimes/modal.py
+++ b/verifiers/v1/runtimes/modal.py
@@ -10,14 +10,18 @@ program in the sandbox reaching a host service) is the shared host-side `Tunnel`
 import asyncio
 import contextlib
 import logging
+import re
 import shlex
 import uuid
 from collections.abc import AsyncIterator
+from ipaddress import ip_address
 from pathlib import PurePosixPath
 from typing import ClassVar, Literal
+from urllib.parse import urlsplit
 
-from pydantic_config import BaseConfig
+from pydantic import model_validator
 
+from verifiers.v1.configs.runtime import NetworkPolicyConfig
 from verifiers.v1.errors import SandboxError
 from verifiers.v1.runtimes.base import (
     SERVICE_PORT,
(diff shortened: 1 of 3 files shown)


verifiers/v1/runtimes/modal.py
```diff
@@ -124,6 +181,14 @@ async def start(self) -> None:
                 "ModalRuntime requires the Modal SDK; install `verifiers[modal]`."
             ) from e
 
+        if self.network_restricted and not hasattr(
```
xeophon, line 184: we pin a modal sdk version that has the capability

xeophon (approved):

PR #2592, opened 2026-09-14
feat(v1): Flow with typed units, explicit call reuse, and boundary steering

eligotts:
Flow runs async stages over durable typed units. Pipelines explicitly seed units, choose admission policy, and interpret the factual result of `run()`. Execution reservations retain the running stage even when steering changes the next cursor.

The author boundary is `Ctx[Data, Config] -> Transition[Data]`: stages edit a typed working copy and publish it explicitly. The single `Transition(...)` constructor infers the data model, call results distinguish `Success[T]` from `Failure`, and failures reuse v1's `Error`. Different units can use different data models. Shared event and inspection records give consumers the same field contracts as the writer. Call and rollout events share one typed `invocation` record.

Successful keyed calls save a typed result or native trace reference. Repeating a partially completed stage reuses successes and executes missing calls. Pipelines declare the complete reuse identity through `inputs`; core fingerprints it without inferring dependencies from tasks, configurations, function arguments, or code. `AgentWork` supplies native agent calls, with a rollout hook for caller-driven interactions; `fn` adapts host functions with an explicit output type. Native `Agent.run` owns rollout retries; failed caller-driven interactions return to the pipeline with their error and trace. Native runtime primitives own teardown.

Optional `GitArtifacts` stores retained immutable `ArtifactRevision` values independently of workflow checkpoints. Pipelines own JSON parsing and content exclusions; the helper provides file reads, writes, listing, and archive export/import. Cached values do not restore sandbox side effects. Stage, call, and native rollout IDs connect execution evidence; attachments retain the original producer's provenance.

Reports are pipeline-owned files under the run's `reports/` directory. A transition names the file it publishes; readers list reports from recorded transition references and read ordinary Markdown files. Cross-unit links identify affected units, without implying a particular target execution.

The control surface is Python plus `inspect`, `steer`, and `drain`. Per-field steering revisions preserve live routes and holds, including changes away and back. Notes are strings: completed stages consume their original prefix, while held transitions retain all notes. Pipeline `commit()` publishes data/files; controls go through `steer()`. Data updates require a settled unit, schema validation, and the inspected workflow revision. Named semaphore pools bound concurrency. Stage work failures hold with evidence for the monitoring agent to act on. Dirty units admitted for execution stop the launch through the existing unit guard; active stages are cancelled and settled.

Flow scopes `VF_RUN_LABEL` to its lock lifetime and records the stable host/root-derived label in `run_started`. Prime sandboxes, Docker/Podman containers (including helpers), and subprocess children carry that identity so operators can identify run resources in shared accounts. Previous process settings are restored on exit or failed entry; labels add no automatic cleanup policy.

There is no Flow retry layer, automatic sweeping, recovery supervisor, command wrapper, inferred cache policy, or compatibility path. Agent-config discovery is shared with native v1. Flow reuses existing v1 trace persistence and leaves CLI episode reading unchanged; its exclusive launch trims an unfinished final append.

Validation: six focused offline Flow cases cover interrupted-stage recovery, partial-call reuse and input invalidation, call and rollout attribution, live steering, retained artifact revisions, admission, and quiescence versus drain. All six and pre-commit pass. Additional temporary offline checks verify dirty-unit launch failure, cancellation and lock release, cross-unit UI links, file-based draining, unit publication guards, note retention, and whitespace-preserving historical Git reads. Expensive rollouts use deterministic substitutes; no live model or sandbox runs. Label propagation was also checked with captured provider creation requests and real local subprocesses. Flow type checks pass; explicitly checking the touched runtime modules reports the same 31 diagnostics as the unchanged baseline.

Core development happens directly on this PR's `feat/flow` branch. E2B runtime implementation remains in #2615.


examples/flow/flywheel.py
```diff
@@ -0,0 +1,98 @@
+"""Offline partial-evaluation recovery, using deterministic substitutes for expensive work.
+
+With Prime-RL installed:
+    uv run flow run examples.flow.flywheel.entrypoint ./out
```
mikasenghaas: should use same discovery mechanism as for envs/tasksets etc. -- load `Flow` subclass from locally installed Python package

also out dir should not be a positional arg. use same pydantic config surface as we have for train/eval

examples/flow/flywheel.py
```diff
@@ -0,0 +1,98 @@
...
+    uv run flow inspect ./out
+    uv run flow steer ./out task --status ready
```
mikasenghaas: what does inspect and steer do? i thought we don't ahve live steerability and monitoring is via external coding agent?

examples/flow/flywheel.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+async def solve(task: TaskData, slot: int, available: bool) -> Evaluation:
+    print(f"executing solver {slot}")
+    if slot >= 6 and not available:
+        raise RuntimeError("provider unavailable")
+    return Evaluation(answer="42", score=float(task.expected == "42"))
+
+
+async def run(root: Path, config: Config) -> int:
```
mikasenghaas: should we not author via `MyFlow[Flow]` subclass to be consistent with other primitives? imo worth doing even if it just has a single public `run` method

mikasenghaas: also `root` should not be here imo. those should use same `output_dir`/ `run.name` scaffolding as we have for other endpoints but should be completely blind to authoring surface (handled in bg)

examples/flow/flywheel.py
```diff
@@ -0,0 +1,98 @@
...
+            [
+                fn(
+                    solve,
+                    ctx.data,
+                    i,
+                    config.available,
+                    output=Evaluation,
+                    inputs=ctx.data,
+                )
+                for i in range(8)
```
mikasenghaas: not a huge fan of meta function like this, e.g. `fn(Callable)` always find those a bit hard to read. is there a particular reason to have this here?

examples/flow/flywheel.py
```diff
@@ -0,0 +1,98 @@
...
+async def solve(task: TaskData, slot: int, available: bool) -> Evaluation:
+    print(f"executing solver {slot}")
+    if slot >= 6 and not available:
+        raise RuntimeError("provider unavailable")
+    return Evaluation(answer="42", score=float(task.expected == "42"))
+
+
+async def run(root: Path, config: Config) -> int:
+    async def evaluate(ctx: Ctx[TaskData]) -> Transition[TaskData]:
+        results = await ctx.spread(
```
mikasenghaas: also what is ctx.spread doing more than just a loop/asyncio.gather?

verifiers/v1/flow/__init__.py
```diff
@@ -0,0 +1,81 @@
+"""Typed units, async stages, and explicit successful-result reuse.
```
mikasenghaas: cut docstring on init file

verifiers/v1/configs/flow.py
mikasenghaas, line 1: should be in `verifiers.v1.configs.flow`

verifiers/v1/flow/events.py
```diff
@@ -0,0 +1,110 @@
...
+from datetime import UTC, datetime
+from pathlib import Path
+from typing import Annotated, Literal
+
+from pydantic import BaseModel, ConfigDict, Field, JsonValue, TypeAdapter
+
+from verifiers.v1.trace import Error
+
+Status = Literal["ready", "held", "waiting", "terminal"]
+RunReason = Literal["quiescent", "draining"]
```
mikasenghaas: `quiescent` what does this mean lol

verifiers/v1/flow/events.py
```diff
@@ -0,0 +1,110 @@
...
+CallStatus = Literal[
+    "started", "succeeded", "failed", "attached", "stopped", "cancelled"
+]
+
+
+def now() -> str:
```
mikasenghaas, line 21: can put this in more general util file

verifiers/v1/flow/events.py
```diff
@@ -0,0 +1,110 @@
...
+    return datetime.now(UTC).isoformat(timespec="milliseconds")
+
+
+class Event(BaseModel):
```
mikasenghaas: would call this base event

verifiers/v1/flow/events.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class SteerEvent(Event):
+    type: Literal["steer"] = "steer"
+    unit: str
+    sha: str
+    action: Steering
+
+
+EventRecord = RunEvent | StageEvent | CallEvent | SteerEvent | LinkEvent
```
mikasenghaas: so we can call this event

verifiers/v1/flow/events.py
```diff
@@ -0,0 +1,110 @@
...
+
+class CallEvent(Event):
+    type: Literal["call", "rollout"] = "call"
+    invocation: Invocation
+    status: CallStatus
+    trace_id: str | None = None
+    error: Error | None = None
+    rollout: int | None = None
+    source_call: str | None = None
+    source_execution: str | None = None
```
mikasenghaas: does invocation and call event belong to each other? if so would try to make this clear in the name that one is the action and on the event record being stored

verifiers/v1/flow/calls.py
```diff
@@ -0,0 +1,267 @@
...
+    def dump(self, value: T) -> StoredValue:
+        """The record fields that carry the value: a `payload`, or a `trace_id`."""
+
+    @abstractmethod
+    def load(self, ctx: Ctx[Any, Any], record: Record) -> T:
+        """The value a record carries; `LookupError` when it cannot be rebuilt."""
+
+
+@dataclass(frozen=True)
+class AgentWork(Work[Trace[Any, Any, Any]]):
```
mikasenghaas: so `AgentWork` and `FnWork` wrap a `Agent.run` and random python fn and add some syntatic sugar like adding events and coalescing into a standard output type? in the way it's currently coded this would live in user author space which i don't rly like bc it's more (redundatn) concepts and syntax to learn. i think we can hide these things in the `Flow` class, e.g. a `Agent.run` invocation would in the bg create a trace + events etc. myb there is smth im missing rn, but im not loving the user api atm

verifiers/v1/flow/traces.py
mikasenghaas, line 1: why do we need new logic here do we not alr have helpers for this?

mikasenghaas: / if we need new stuff would prefer to unify it with existing helpers

mikasenghaas (commented):
some high level comments:
- need to bring the code style a bit closer to existing entrypoints (e.g. pydantic config for flow cli in prl, class based not script based flow authoring)
- i feel like there rn there is a lot of new concepts to learn for authoring a pipeline, e.g unit, stage, transition, event, agent/ fn work etc etc. i was hoping one can do more without learning many new abstractions. this might be just me being new to the api tho but gut feeling is we can simplify a bit and do more behind the scenes if possible . would be good to go through each component i think and see why it is there currently and if it can be removed or hidden

verifiers/v1/flows/draft_review.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @stage
+    async def draft(self, unit: Unit[Data]) -> Transition[Data]:
+        trace = await self.agents.writer.run(Task(TaskData(prompt=unit.data.prompt)))
+        unit.data.revision = GitArtifacts(unit.path).write(
+            base=None, files={"draft.md": trace.last_reply}
+        )
+        return Transition("written", stage="review", data=unit.data)
+
+    @stage
```
mikasenghaas, line 40: how do i know that draft runs before review? by order of the definition? if so, how would i do parallel stages?

verifiers/v1/flows/draft_review.py
```diff
@@ -0,0 +1,50 @@
...
+    async def review(self, unit: Unit[Data]) -> Transition[Data]:
+        assert unit.data.revision is not None
+        draft = GitArtifacts(unit.path).read(unit.data.revision, "draft.md")
+        trace = await self.agents.reviewer.run(
+            Task(TaskData(prompt=f"Review this draft:\n{draft}"))
+        )
+        return Transition("reviewed", trace.last_reply, status="terminal")
```
mikasenghaas, line 47: is "terminal" a reserved status basically?

Issue #2189, opened 2026-07-30
v1: model Harbor artifacts as durable outputs, not only grading transport


hallerite:
## Context

#2144 adds artifact collection/restoration for isolated grading and imports Harbor's `artifacts` and `[[verifier.collect]]` task fields.

The current implementation uses artifacts only as an in-process transport channel:

- `HarborTask.finalize()` runs collect hooks and archives `/logs/artifacts/` plus declared paths for every Harbor rollout.
- The tar bytes are stored in `trace.state.artifacts`, which is excluded from trace serialization.
- Isolated `agentic-judge` restores those bytes into the fresh judge sandbox.
- Single-agent and shared-judge runs still archive the files, but nothing consumes or persists the bytes.

Current code: [Harbor collection](https://github.com/PrimeIntellect-ai/verifiers/blob/4228c44c72887a33f8edfcd22bdd4d316694944c/verifiers/v1/tasksets/harbor/taskset.py#L115-L142), [transient state](https://github.com/PrimeIntellect-ai/verifiers/blob/4228c44c72887a33f8edfcd22bdd4d316694944c/verifiers/v1/state.py#L14-L16), [isolated restore](https://github.com/PrimeIntellect-ai/verifiers/blob/4228c44c72887a33f8edfcd22bdd4d316694944c/verifiers/v1/envs/agentic_judge/env.py#L179-L193).

## Harbor semantics

Harbor describes artifacts as durable trial outputs: collected files are written under `<trial_dir>/artifacts/`, accompanied by a `manifest.json`, and exposed in the results viewer. With a separate verifier, those same artifacts are additionally restored into the verifier at their original source paths.

Harbor collection is best-effort and records failures in the manifest rather than failing the trial. Its `destination` field controls host-side placement without changing verifier-side restoration.

References:

- https://www.harborframework.com/docs/run-jobs/results-and-artifacts
- https://www.harborframework.com/docs/tasks#verifier-environment-shared-vs-separate

## Problem

We currently conflate two responsibilities:

1. **Result archival** — preserving rollout outputs for inspection and downstream use.
2. **Grading transport** — moving selected files into an isolated verifier sandbox.

Consequences include:

- shared and single-agent Harbor rollouts pay archive/network/memory costs for bytes that are then unused;
- artifacts are not durably available in eval output or a viewer;
- the grading-transport 32 MB cap and strict failure behavior also define the imported Harbor artifact surface;
[...]

PR #2620, opened 2026-09-17
feat(v1): separate artifact grading transport from archive

jurij-jukic:
So far, the grading transport and output archival functionalities were conflated as described in issue #2189, which this PR addresses.

This PR takes collection from sandbox and splits it into two features:
- `grading_collect` function moves artifacts from sandbox to RAM for scoring by another sandbox.
- `archive` function writes the desired artifacts from a sandbox into the output folder, for later inspection.

## Usage

The archive policy can be specified in the eval config file:

```toml
[archive]
extra = ["hello.txt"]   # optional paths beyond task defaults
max_mb = 256            # per-rollout host dump cap (MiB)
```

Note that grading and archive caps are now separated (32 MiB for transport vs 256 MiB default for archive).

Archives have the following layout:

```
outputs/<run>/
  traces.jsonl
  artifacts/<episode.id>/<trace.id>/
    manifest.json
    *.tar
```

Harbor tasks can now use `task.toml` with artifacts `destination`, and `[[verifier.collect]]`:

```toml
artifacts = [{ source = "hello.txt", destination = "greeting.txt" }]

[[verifier.collect]]
command = "cp hello.txt /tmp/snap.txt"
```

## Design

`collect` and `restore` functions from `artifacts.py` retain their functionality (moving files between RAM and sandbox), however, `collect` is now abstracted away from grading, and serves solely as a function that moves files back and forth.

`collect` has two callers:
- `grading_collect` function from `grading.py` which specifies the error behaviors for `collect` depending on collection policy needed for grading (explained in next section)
- `archive` function from `archive.py` which utilizes collect to archive the desired files.

### Grading Transport Policy

`GradingCollect` enum is now used to specify the policy for collection. Previously there were flags such as `collect_artifacts`, and each `Artifact` object had its own `required` flag. Now, those are removed, and GradingCollect can be specified per each env.

- `OFF` — default, no collection happens, used for single sandbox environments.
- `STRICT` — declared paths must exist or the rollout fails, with the exception of `/logs/artifacts` which may be empty. `STRICT` is used in isolated verifier and isolated agentic judge.
- `BEST_EFFORT` — declared paths can be missing and the rollout won't fail, this is used with Harbor.

### Archive Policy

Every rollout attempts to archive the following:
1. `/logs/artifacts` - default
2. `task.data.artifacts` - paths the task declared for grading
3. `[archive].extra` - additions defined in the eval config.

Same archive policy is applied to all agents in the episode. So e.g. solver and judge will use the same `[archive].extra` list. 

Each archive has a `manifest.json` file. Each manifest entry refers to one file being collected. The values in a `ManifestEntry` are the following:
- `source` - path of the file in the sandbox
- `file` - the output path it was stored on (or null if the file was missing)
- `harbor_destination` - destination if specified by a Harbor task, otherwise null
- `status`- `ok` or `skipped`.

Archive error policy:
- If a source is missing, it shows up as `skipped` in the manifest (`file: null`).
- If a source is over the size limit (`[archive].max_mb`), a warning is logged and it shows up as `skipped` in the manifest.

## Env Behavior

Note the following design decision: archival happens _only_ at the end of a rollout. So when there is a verifier sandbox without a rollout, its grading logs are _not_ archived. This behavior applies to isolated verifier and Harbor evals with separated grading. Their scores can still be found in `traces.jsonl`.

In the case of agentic judge, the judge is also a rollout, so both the solver and the judge get their own dirs inside an episode:

```
outputs/<run>/
  traces.jsonl
  artifacts/<episode.id>
    /<solver.trace.id>/
      manifest.json
      *.tar
    /<judge.trace.id>/
      manifest.json
      *.tar
```

Then, there is also behavior specific to Harbor tasks which this PR implements:

- Harbor task config can specify a `destination` onto which a file can be archived. General non-Harbor behavior simply mirrors the sandbox paths.
- Harbor configs can specify `environment_mode = "separate"` and `[[verifier.collect]]` which now work with verifiers repo. Note that if the `[[verifier.collect]]` hook fails then an error is raised, which is stricter than the Harbor implementation.
- Harbor sidecars are still not supported.

## Tests

Because clearing of RAM was implemented at the end of rollout, the isolated `agentic-judge` test now expects `solver.state.artifacts == {}`, instead of `"/app/answer.txt" in solver.state.artifacts`.

The following checks pass:

```bash
uv run ruff check verifiers/
uv run pytest tests/v1 -n auto -m "not e2e"
```

> [!NOTE]
> ### Replace boolean `collect_artifacts` with `GradingCollect` policy and add `ArchiveConfig` for durable sandbox archives
> - Splits grading artifact transport (in-memory collection for verifier handoff) from durable host-side archiving (tar files on disk) into two independent mechanisms.
> - Adds `GradingCollect` enum (`OFF`, `STRICT`, `BEST_EFFORT`) in [grading.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2620/files#diff-9cd028eae348cb372e0dc47c84db76f5543f4c4036adf1c087b1082606fc80cf); strict raises on missing/over-budget roots, best-effort omits missing but still raises on overflow.
> - Adds `ArchiveConfig` in [archive.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2620/files#diff-8c9ba808efa2bec39d611b59efa6a08992ec5b8d9a95de51f92e024803573c53) with a default 256 MiB per-rollout size limit; `Rollout.close` writes best-effort tar archives and a JSON manifest when an archive directory is configured.
> - Removes the `required` field from `Artifact` in [artifacts.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2620/files#diff-0d082bad54931be61d7ef55ddce0757034be3013f69ae71cb803cd8416df9138); collection limits and missing-root behavior are now driven by policy, not per-artifact flags.
> - Episode lifecycle in [env.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2620/files#diff-69be10bd5960e88d88f219d4f261d0bdd16586857f9126451b04d05bac4c2cbb) clears artifact mappings from all retained traces after each episode attempt; discarded retry traces have their archive directories removed.
> - Risk: callers passing the removed `collect_artifacts` boolean to `Agent.run` or constructing `Artifact` with `required` will break; subprocess-backed rollouts now fail before execution if a non-`OFF` `GradingCollect` policy is set.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ecf860a.</sup>
> 
>


jurij-jukic:
Formatting fixed, should pass now.

PR #2566, opened 2026-09-08
Stream Bash/edit eval requests through long reasoning turns

samsja:
Long Bash/edit reasoning turns can exceed the bundled client's read timeout while a healthy model is still generating. The shared Bash/null chat helper now always requests streaming, so model deltas and interception-server SSE keepalives reach the sandbox throughout the turn. There is no stream CLI flag, harness setting, or helper/compactor parameter; the shared launcher and compaction plumbing remain unchanged.

The SDK assembles complete responses before tools execute, preserving fragmented arguments, reasoning, usage, and `finish_reason="length"`. Assistant role metadata is normalized because some providers repeat it in each delta. Empty and unfinished streams fail before tool execution.

Training clients continue generating complete responses through the normal graph-commit path, retaining token IDs and log probabilities. The interception server frames the completed training response as SSE for the harness; this does not add token-by-token generation to training. The chat dialect includes tool-call indexes when framing complete responses. Browser-use retains its existing launch arguments.

Validation:
- Repository checks: 80 tests passed, 76 credential-dependent tests skipped; all-file Ruff and pre-commit passed.
- Three live CI-model tests passed: colocated tools, repeated tool use, and Bash execution.
- Temporary HTTP regressions cover repeated role metadata, keepalives, interleaved Bash/edit arguments, reasoning/usage, length termination, and empty/truncated streams.
- A real interception-server test with a stubbed training generator verifies two tool calls and a follow-up, committed token IDs/log probabilities, usage, and length termination. It passes against both the PR checkout and the installed runtime backport.
- The previous revision's full GitHub CI passed. CI is rerunning for this revision; Docker/browser live coverage runs there because Docker is unavailable on this host.

This addresses idle request deadlines; it does not establish that all observed gateway 504s have the same cause.

---

> [!NOTE]
> **Medium Risk**
> Eval completion transport changes for bash/edit runs; truncated streams now raise `RuntimeError` instead of proceeding with partial output, though training paths remain non-streaming.
> 
> **Overview**
> Bash/edit eval harnesses can now use **streaming chat completions** so long model turns stay alive past idle read timeouts (e.g. 600s) via SSE keepalives, while the loop still waits for a **full** completion before running tools.
> 
[...]

PR #2623, opened 2026-09-17
fix: restore buffered bash completions

mikasenghaas:
## Summary

- Restore complete-response requests in the bundled Bash and null chat harnesses.
- Remove the stream-only response assembly added in #2566.
- Keep the 600-second Bash client timeout and the OpenAI SDK retry policy unchanged.

## Verification

- Before: the harness set `stream=true`, so each response chunk reset the
  600-second read timeout. One active stream had no client-side wall-clock limit.
- After: the harness waits for one complete response. The regression test checks
  that the request does not enable streaming.
- OpenAI 2.54.0 retries a read timeout twice. A silent model call can therefore
  take at most 1,801.5 seconds: three 600-second attempts plus up to 1.5 seconds
  of retry backoff. The outer rollout timeout remains the final backstop.
- `uv run pytest tests/` — 101 passed, 81 skipped.
- `uv run ruff check --fix .` — passed.
- `uv run pre-commit run --all-files` — passed.

> [!NOTE]
> ### Restore buffered completions in `chat` async helper
> - Switches `chat` in [core.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2623/files#diff-4641f7de541bf21c7f3e30a2d8d49f14c1179c70cf21f3dff5fa19deb9a1e206) from a streaming `AsyncChatCompletionStream` path to a direct non-streaming `client.chat.completions.create` call, returning the completion produced by the OpenAI client directly.
> - Removes the streamed-message accumulation utility and its associated imports that merged role, reasoning, and reasoning-detail deltas.
> - Updates [test_e2e.py](https://github.com/PrimeIntellect-ai/verifiers/pull/2623/files#diff-ebbbd919487889487f891b359a756e6c4630e106b16ddaef8ccddc6b900a96e1) to assert the request does not set `stream` and to supply a complete JSON chat completion response.
> - Behavioral Change: `chat` no longer requests stream output, includes stream usage options, or reconstructs reasoning fields from streamed deltas.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 31c93a3.</sup>
> 
>

---

> [!NOTE]
> **Medium Risk**
> Changes the default completion path for bundled bash/null chat loops (via `Compactor`), affecting timeouts and how reasoning fields are obtained; behavior now depends on non-streaming API responses instead of custom stream reassembly.
> 
> **Overview**
> Reverts the bundled chat harness **`chat`** helper to **buffered** OpenAI completions instead of **`stream=true`** with manual delta assembly.
> 
> **`verifiers/v1/harnesses/utils/core.py`:** `chat` now calls `client.chat.completions.create(**kwargs)` directly. The streaming path (`AsyncChatCompletionStream`, `stream_options`, and `_accumulate_streamed_message` for role/reasoning/reasoning_details) is removed.
> 
> **`tests/v1/test_e2e.py`:** The harness unit test is renamed to **`test_chat_harness_requests_complete_response`** and asserts the request does **not** set `stream`, using a single JSON **`chat.completion`** mock instead of SSE chunks.
> 
> **Why it matters:** Each model turn waits for one full response, so the Bash harness’s **600s read timeout** applies per completion (with SDK retries) rather than resetting on every stream chunk—addressing unbounded wall time on long streams while keeping reasoning fields when the API returns them on the complete object.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 31c93a3800487ea3e65f93279c43620770afd22c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hallerite (approved):

samsja (approved):

mikasenghaas:
We could not reproduce a higher error rate from the streaming Bash harness.

We ran Terminal-Bench 2 through the prime-rl eval entrypoint against
`deepseek/deepseek-v4.1-flash` on Prime Inference:

- Main with streaming, avg@1, concurrency 128: 68 landed rollouts. The only
  error was task 27 (`financial-document-processor`). Excluding that
  deterministic task failure, the error rate was 0/67 (0%).
- This PR with buffered completions, avg@1, concurrency 128: 66 landed
  rollouts. The same task 27 failed. Excluding it, the error rate was 0/65
  (0%).
- Main with streaming, avg@4, concurrency 512: 231 landed rollouts. The only
  four errors were the four replicas of task 27. Excluding them, the error
  rate was 0/227 (0%).

The task-27 failures were first-call `openai.APIConnectionError` failures in
both streamed and buffered runs. The evidence does not show a streaming error
regression, so we are closing this PR without merging it.

